PR #47, opened 2020-04-15
Add Posts Inserter block

### LINKED ISSUE #18: Inserting Posts Into Newsletters
Users will need a way to insert recent posts into Newsletters—a list of recently published content is an essential feature of an editorial newsletter. A complicating factor: newsletter authors often tweak the titles, excerpts, and read more links to be more appropriate to the email context. As a result, this challenge is a bit more complex than homepage layout.

## Homepage Posts Block

One way to achieve this would be to support the Homepage Posts and/or Latest Posts blocks. This would be the most straightforward and achievable approach at the moment. The content of these blocks is dynamic and not editable, so context tweaks are not possible. Also, these blocks reflect the latest published content, so the content could change in the interval between creating and sending a newsletter. This may be undesirable if other parts of the newsletter refers to specific posts. 

## Dynamic Patterns

A second approach would be an extension of block patterns that could be called "dynamic patterns." When one of these patterns is inserted, a modal appears with post selection UI (similar to the Homepage Posts block query options). The user selects the post(s) to insert and the blocks are inserted into content. The pattern would consist of simple, editable blocks: Heading, Paragraph, Image, Buttons. 

## Convert To Editable Blocks

A third approach would be a utility that converts an instance of the Homepage Posts block into a set of editable blocks: Heading, Paragraph, Image, Buttons. User would insert Homepage Posts, adjust the attributes so that they are seeing just the posts they want in the newsletter, then click a "convert" button. The operation replaces the Homepage Posts block with an equivalent group of editable blocks so the author can "own" the content.

adekbadek:
### All Submissions:

Adds a "Posts Inserter" block. It's in large part mimicking the core "Latest Posts" block, but adds the _preview, then edit_ flow. 

Closes #18 

1. Add new Newsletter
2. Add a "Posts Inserter" block
3. Observe a block with a latest posts list preview and an "Insert posts" button
1. Tweak the block settings in the right hand panel, observe the preview updating
1. Click the "Insert posts" button, observe that the previewed content is now in newsletter content
1. Edit the post titles and other elements
1. Send a test email/campaign. Observe that the email matches the editor content

### Other information:


thomasguillot:
Could we allow users to choose how they want to display the featured image?

Screenshot 2020-04-16 at 15 38 05

Could we have the ability to select specific post(s), just like the Homepage Posts block does. This would allow users to create "complex" layouts

Screenshot 2020-04-16 at 15 43 00

adekbadek:
> Could we allow users to choose how they want to display the featured image?

Sure, I'd like to add this and other features – but this PR is already big, so I'd prefer to do that as a separate PR. I've added #53 to keep track of proposed improvements.

> Could we have the ability to select specific post(s), just like the Homepage Posts block does. 

How would the UI for that look like? Feel free to add it to #53

jeffersonrabb:
This is clever and works well. Here are a few issues and questions:

1) What if, instead of using an `InnerBlocks` implementation, this block simply inserts the blocks directly into the content, replacing itself? In this approach, the block would only exist in the post as a selection tool. After the Insert button is clicked, its usefulness is over and it is removed. Crazy? Several of my other concerns become non-issues if we were to go this way but I'll list them anyway...

2) After inserting blocks it's very easy for the next block created to end up within the `InnerBlocks` of Latest Posts. Add Latest Posts, click "Insert," then click the + to add another block. The new block ends up a child of Latest Posts which isn't right. `templateLock` and/or `renderAppender` might hold a solution to this.

3) The `InspectorControls` empties out after the Insert button is clicked. It makes perfect sense why this is so, but could be confusing for a user trying to understand why the block sometimes has controls and sometimes does not.

4) It makes me uneasy to have the name match the core block. The rationale makes lots of sense - we want users to use it for the same reason - but I worry this will create confusion.

5) Just to confirm, `isReady` essentially means "insert has been clicked?"

adekbadek:
@jeffersonrabb I've did as you suggested in – ce5e863a54efc22386994731492adbc2ade4a722 – this was a great hint, thanks! 
So with 1, 2, 3 out of the way, that leaves us with:

4. That's true, and as @thomasguillot pointed out, with better querying these might not even be "latest" posts. How about `Posts List`?

5. Yes, I've changed it to `areBlocksInserted` so it's more understandable.

adekbadek:
@thomasguillot Updated to match placeholder style in 42a16c3 :
image

I think the block description might not be needed here.

adekbadek:
Renamed to "Posts Inserter" in ae62f57

PR #121, opened 2020-04-22
feat: error handling in save and publish operations

jeffersonrabb:
### All Submissions:

Adds error notifications in save and publish operations. These are trickier to catch since they are handled by standard Gutenberg XHR requests as opposed to custom endpoint defined by the plugin. To work around this, errors caught in save/publish are stored in a transient for 45 second, and are then displayed the next time the Mailchimp data retrieval endpoint is hit. This is tricky to test since many things you could do to break the flow will _also_ break the retrieval endpoint, so it's not clear where the error is coming from. The best way to test is to break the MML operation because this only occurs during save. 

1. Create a Newsletter, select a template.
2. In Newsletters->Settings, add a character to one of the MJML fields.
3. In the editor, save or publish the post. Observe the `MJML Error.` error notification.

### Other information:

### DIFF
--- a/includes/class-newspack-newsletters-renderer.php
+++ b/includes/class-newspack-newsletters-renderer.php
@@ -533,6 +533,7 @@ final class Newspack_Newsletters_Renderer {
 	 *
 	 * @param WP_Post $post The post.
 	 * @return string email-compliant HTML.
+	 * @throws Exception Error message.
 	 */
 	public static function render_html_email( $post ) {
 		$mjml_creds = self::mjml_api_credentials();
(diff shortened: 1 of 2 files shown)


jeffersonrabb:
> Since the errors are persisted for 45 seconds only, what's the plan for handling errors of sending scheduled campaigns?

Publish errors arguably require a more substantial response whether scheduled or not. First of all, if the newsletter failed to send (which may not always be the case), the post should be unpublished so the user can try again. Second the user should be notified in a more obvious way than an editor notification. This could take the form of an email notification and/or an admin notification shown throughout wp-admin. In any case this flow exceeds the scope of this PR, and can probably be addressed after the first beta release. I'll open an issue to track.

PR #127, opened 2020-04-23
Posts Inserter block - handle before insertion state

### LINKED ISSUE #114: Posts Inserter block – no-insert flow
The Posts Inserter block should allow the following flow:

1. user inserts the block into post
2. user sends the newsletter, without clicking "Insert posts"
1. content is inserted at the time of the publication, but in the post itself, the block remains in the "before insertion" state

This way 1) there is no accidental publication of an empty block, 2) a kind of automated posts insertion possible

adekbadek:
### All Submissions:

If the posts are not inserted, they will be rendered in email as they are, without changes. 
The posts' blocks, as they _would_ be inserted, are stored in `innerBlocksToInsert` attribute.

Closes #114

1. Create a newsletter, add "Posts Inserter" block. Do **not** click "Insert posts"
2. Preview or send email - observe the posts are displayed, as if they were inserted
3. Change the settings of the block; repeat step 2 – changes should be reflected
4. Click "Insert Posts" button – observe the "Posts Inserter" blocks being replaced with the core blocks
5. Edit the blocks, repeat step 2 – changes should be reflected

### Other information:


jeffersonrabb:
Why is `innerBlocksToInsert` needed, as opposed to simply using normal `InnerBlocks`?

adekbadek:
Handling the posts' blocks as `innerBlocks` (akin to how column handles its contents) would make them editable, so we'd have to disable that. It seems more hacky than storing the blocks in an attribute.

jeffersonrabb:
Second question: what's the reason to use `memoize-one` vs. [lodash memoize](https://lodash.com/docs/4.17.15#memoize), or [memize](https://www.npmjs.com/package/memize) which is used in some Gutenberg libraries e.g. https://github.com/WordPress/gutenberg/blob/master/packages/i18n/package.json#L31?

adekbadek:
Just my fond memories of using it and a good track record. Lodash's `memoize` would need a custom resolver for the cache key, because the default is not suitable in this case. And I have no experiences with `memize`, but the npm numbers are on `memoize-one`'s side ;)

PR #149, opened 2020-04-27
Update editor and MJML style

thomasguillot:
### All Submissions:

This brings the editor closer to what the email will actually look like. There are still a few differences (e.g. padding on blocks with background colour) but these are pretty minor differences.

1. Create a new newsletter. Send a test email. Spot the differences.
2. Switch to this branch.
3. Refresh editor and send another test email. There should be less differences.

### Other information:


includes/class-newspack-newsletters-renderer.php
```diff
@@ -145,7 +145,7 @@ function ( $key ) use ( &$attrs ) {
 		);
 
 		if ( isset( $attrs['background-color'] ) ) {
-			$attrs['padding'] = '16px';
+			$attrs['padding'] = '0';
```
adekbadek, line 148: This makes background-colored paragraphs in columns render incorecctly:

image

thomasguillot: Should be fixed in https://github.com/Automattic/newspack-newsletters/pull/149/commits/3b3b1cda698db21bcb7106f4beaf4bac8ff8db7f

adekbadek: One more thing I forgot about earlier - same issue is with Heading block in a column.

thomasguillot: Oh yeah. I've been less specific in https://github.com/Automattic/newspack-newsletters/pull/149/commits/d2a5bdd1e3b767c102e000333e595d951159e95f, hopefully it should do the trick.

PR #165, opened 2020-04-28
Add layouts management

### LINKED ISSUE #31: Templates CRUD
Allow user to modify the templates collection.

### LINKED ISSUE #216: Layout Picker Issues In Safari
This will only be relevant after https://github.com/Automattic/newspack-newsletters/pull/165 lands, but there are a two layout problems in Safari in the Layout Picker modal. 

1) Preview Pane doesn't fill full space when empty:

Screen Shot 2020-06-04 at 9 57 41 AM

2) After making a selection the left column empties out. It actually appears that the height of the overall container becomes the full height of the layout, which effectively pushes the thumbnails way down:

Screen Shot 2020-06-04 at 9 59 40 AM

adekbadek:
### All Submissions:

Enables managing layouts.

Closes #31 
Closes #216

1. Create a new newsletter – observe two tabs in the layout selection pane: "Default layouts", "My layouts"
2. Observe that the latter tab is empty and displaying a message about new layout creation
3. Choose one of the default layouts to create a newsletter 
3. Observe a new "Create a new layout" button in the "Layout" section of the sidebar Document settings, click it
3. Observe a modal displayed with new layout name input. It should be set to the newsletter title 
3. Click "Save" 
3. Create a new newsletter. Observe that the new layout is listen in the "My layouts" tab
3. Select the new layout to create a new newsletter
3. Edit the newsletter content
3. Observe a "Update layout" button besides the "Create a new layout" button in the "Layout" section of the sidebar Document settings, click it
3. Change the name of the layout and save it
3. Add new newsletter. Observe the layout content & title have been updated
3. Click "Delete" button below the layout tile in the selection tab, confirm
3. Observe the layout has been deleted

### Other information:


thomasguillot:
> Layout settings in Newsletter editor – "save as layout" flow should enable setting a title (now it's hardcoded to "Newsletter Layout"). In a modal, maybe?

Good idea.

Screenshot 2020-05-04 at 15 38 21

jeffersonrabb:
Should Posts Inserter suppress the "Insert Posts" button and functionality in the context of an editable layout? A common mistake may be: 

- create a template
- add Posts Inserter block
- insert posts
- create a newsletter from the layout

At this point there would be an expectation of dynamic post selection but they will actually be hardcoded. 

A similar question will come up when creating a layout from a newsletter then reusing that layout:

- New newsletter from "Daily/Weekly" layout.
- Insert the posts.
- Make other edits.
- Save Newsletter as a Layout
- Start new newsletter from the layout.

adekbadek:
> Should Posts Inserter suppress the "Insert Posts" button and functionality in the context of an editable layout? 

Good idea, done in 25a924752508da200a0180e413ed36ca83b678cc

> A similar question will come up when creating a layout from a newsletter then reusing that layout

That's a hard-to-handle edge case IMO – if there is "static" content (post-insertion), and the newsletter is saved as layout, I think it's understandable that the content stays as it is. There's no UI indicating that the inserted posts are dynamic content.

Retrofitterdk:
I know I am entering a discussion, which I am not really a part of, so please let me apologize beforehand.

But I am curious as to why newsletters need a proprietary layout system, when the Gutenberg Full Site Editing project is trying to do exactly that and adding layout templates as a CPT.

Also I am wondering if it wouldn’t be easier to let users add predefined layouts and placeholder content as a Block Patterns in stead choosing a complete layout for at newsletter to begin with.

With something like [Block Pattern Builder](https://github.com/justintadlock/block-pattern-builder) –  – from @justintadlock it may even be possible to create their own pattern.

And by the way I am really excited about this project - been for years for something like this.

adekbadek:
Hi @Retrofitterdk! The layouts are here for two primary reasons:
1. A "newsletter" post is a special kind of post that whitelists a set of blocks and provides some helpful notices to the user – this is why proprietary handling of this kind of post is needed
2. We assume layouts are more heavily used in for newsletters, unlike in regular posts

And block patterns can still be used ;)

jeffersonrabb:
There's a lot of great work on this branch bit I think we should rethink the approach a bit—there are a few aspects of this implementation that don't feel quite right:

- Instantiating the user layouts on plugin installation makes it challenging for us to add, remove or alter layouts in future releases.
- It also makes it difficult for the user to start over if they have broken their layout, or if they want to experiment with different approaches based on the same factory layout.
- It's confusing to have two surfaces for editing Newsletter content: Layouts and Newsletters. The "Add New" button illustrates this confusion, but if we separated the two out  it could become even more complex and confusing. In other words, there should be only one place to interact with Newsletter content, and UI for saving the current work for reuse.

Based on these principles, here's what I propose:

- In the layout picker there would be two tabs: "Default" and "Custom". (We can find better names for the categories before rolling it out, maybe Factory|User)
- The Default tab looks exactly like the current Layout Picker on `master`, and User lists custom defined layouts. Initially the User tab is empty.
- There is no admin menu for Layouts.
- For Newsletters created from default Layouts the Layout Sidebar Panel has a "Create Custom Layout" button that works just like the one in this branch. The title modal is great too.
- If a Newsletter is based on a Custom Layout (either because the button was clicked, or it was created from one), the Layout Sidebar Panel has an "Update Layout" button that overwrites the layout with the current content. 
- The Custom tab in the picker has UI for deleting custom layouts.
- Perhaps there could also be UI for duplicating layouts, but this is less important.

In summary, the CPT functions very much as it does in this branch, but the user can only interact with their custom layouts through the template picker and the Layout sidebar panel. Management of custom layouts occurs in the picker — that's the only time the user sees them so it's the only place where they would feel compelled to delete unused ones. The only way to edit a layout is to work on a Newsletter then update the Layout it came from, which feels like a more natural way to work since all of the previewing and testing tools are available. 

WDYT?

adekbadek:
I agree, a simpler UI will be better 👍

adekbadek:
@jeffersonrabb I've updated the branch and PR description with new testing instructions

adekbadek:
Update: I still have to remove the code for handling editing a layout in the editor (separate JS and CSS bundles)

jeffersonrabb:
Really like the changes! A few more notes:

1) After clicking `Create Layout` I expected the sidebar panel to immediately reflect the new name of the template. In other words, if I create a Newsletter using the "Breaking News" template, then create layout and name it 'My Breaking News," I expected the sidebar to change to say "My Breaking News" and show the thumbnail of the newly created layout.

2) After creating a Newsletter from a User Layout when clicking Update Layout I didn't expect the modal to name it. Ideally this modal would only appear when the User Layout is created, and there would be a way to edit the Layout name in place in the picker. 

3) If there are any User Layouts defined, it might be nice to make My Layouts the default tab. It seems likely that a user who has made custom layouts will probably use them much more than the default layouts.

jeffersonrabb:
@laurelfulford this PR introduces a few new pieces of UI — could you give it a look? Here are the two areas where UI has changed:

Screen Shot 2020-05-26 at 1 47 05 PM

Screen Shot 2020-05-26 at 2 00 09 PM

Some specific UI questions:

- Style of the Default/User tabs in the picker.
- Deletion UI in the picker.
- How to handle multiple buttons in the sidebar panel.
- Button text.

adekbadek:
@jeffersonrabb I've made the requested changes:
1. Updates in Sidebar panel: 08c16edcbd95e2c24a4e1ae3c5ee66218b82aec3 , 8b85307e2287b5a3028ebadb93ab3ab4c7e8463a. The newsletter's `template_id` will be set to the newly created layout. 
2. Renaming in layout picker: 64ffe38f16e18e234f20736316cf373ae0fa9c3b
3. Tab switching e04537fb3c89d297b7c1bc351d313129253aaa66

laurelfulford:
**Style of the Default/User tabs in the picker.**

Digging into this a bit, there is a prescient in Gutenberg for tabs that you can see in the right sidebar, as well as the block picker now that patterns have been added:

image

image

There's some information in the [Tab Panel component documentation](https://github.com/WordPress/gutenberg/blob/6b8eb7ae105b8078f606c2ad368d1c2f49cb7702/packages/components/src/tab-panel/README.md).

The one issue with that and where the page is now, is that the left side isn't really a separate panel like these examples, with a hard left and right edge -- on the right, it's open at the top and bottom to create a frame around the newsletter preview. So adding it as is would look like:

newsletter-mockups-tabs-A-floating

To give it the right container, we could remove the frame around the newsletter and make it more editor-like:

newsletter-mockups-tabs-B-panel-change

Or add a vertical line to make the left area into a separate contained 'panel':

newsletter-mockups-tabs-C-add-panel

(Note: The tabs are only the width of their labels in the right sidebar, and split evenly 50-50 in the block picker).

---

**Deletion UI in the picker.**

I like this approach

The text button style seems to be rendered with an underline more often than not, though -- what do you think about adding one here? 

In testing it, it also seemed a bit confusing that the whole preview, name of preview, and delete button share the same hover area -- the cursor stays as a pointer, and the Delete button doesn't change when you're in its hit area:

delete-hover

What do you think about adding a hover state specifically for the button? The Remove Featured Image is probably the closest example -- it changes to the lighter blue used by other text-style buttons: 

featured-image-delete

---

**How to handle multiple buttons in the sidebar panel.**

I'm not sure how to get the 'Update Layout' button to appear -- this is what I get, after editing a layout I created:

image

The buttons should be ordered and styled based on priority, with only one using the 'primary' style (if any). What's the most important thing someone would do? It seems like Create Layout, then Update Layout, then Reset Newsletter Layout. So something like:

image

Or, forgoing the primary style and sticking with default and text:

image

I also wonder if 'Update Layout' should also be set as `isDestructive`, since it can override existing layouts? (please disregard if this is already set to flash a warning before saving!). 

---

**Button text**

The one that confused me the most of these was 'Create a Layout' -- I wasn't sure if it would bring me to a entirely blank editor, or save the changes I made as a new layout. 

They're much clunkier, but I think something along the lines of 'Save new Layout' or 'Create Layout from changes' would be clearer. The button should also be disabled until changes are made in the editor, since you'd be otherwise creating a duplicate of an existing layout. 

image

image

---

@jeffersonrabb let me know what you think of the above, if you have any questions or if you'd like me to take a swing at any of these changes!

includes/templates/breaking-news-no-image.json
```diff
@@ -0,0 +1,5 @@
+{
```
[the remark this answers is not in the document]
adekbadek: Yep, fixed in 93ee31bcb96f2485c78a2f9991117b9afb02afdd

adekbadek:
@jeffersonrabb Thanks for catching those! Fixed now.

- When editing titles in the picker, I can't use spaces: 20a418d56793c414c86621729f2e39a7361cbb79
- When titling a layout (either in the modal or in the picker) typing is slow and "sticky.": bce62f49587382d505ca4b4c86dde44cb6bc7d7c
- The picker shows a flash of the default layout tab before switching to the user layout tab: 21260b4f6b9d1ec3ce7ef6342ed48d62a46d167a
- The "Reset newsletter layout" button no longer works: 2501e02fc09549ca28b30bc1fadcc4ad140c3e02

---

@laurelfulford I like those suggestions!

1. For layout picker tabbing, I'd go with the second option. 
2. "a bit confusing that the whole preview, name of preview, and delete button share the same hover area " - fixed in 20a418d56793c414c86621729f2e39a7361cbb79
3. "adding a hover state specifically for the button" 👍
4.  "I'm not sure how to get the 'Update Layout' button to appear" - you need to edit a newsletter with a set user-defined layout 
5. "I also wonder if 'Update Layout' should also be set as isDestructive, since it can override existing layouts?" - there's no warning here, so isDestructive is a good idea
6. "They're much clunkier, but I think something along the lines of 'Save new Layout' or 'Create Layout from changes' would be clearer. " 👍 
7. "The button should also be disabled until changes are made " 👍 

I'll wait with applying these changes for Jeff's opinion.

jeffersonrabb:
> I'll wait with applying these changes for Jeff's opinion.

Agree with all of it, and I also like the second option for the tabs best.

adekbadek:
1. Tabs in layout picker: d3c4e513ed82b7b3646d4891de0e9c973cb3288f
2. -
3. Hover state for layout delete button: 62ef68b4888ba3908680a9ad2f6362207bedc5c2
4. -
5. 'Update Layout' styling:  `isDestructive` seems to be handled only for `isLink` buttons. I've added a confirm dialog, as there is for deleting layouts: 7b588bafa87f8067c117627e28ab0cfe65ebde93
6. Layout settings tab buttons… layout: 7b588bafa87f8067c117627e28ab0cfe65ebde93
7. Disable 'Update Layout' button unless changes are made: c4fd6f8abc6302c7bff9d8d13df80e8cc251b273

includes/class-newspack-newsletters-layouts.php
```diff
@@ -0,0 +1,130 @@
...
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Main Newspack Newsletters Class.
+ */
+final class Newspack_Newsletters_Layouts {
+	/**
+	 * CPT for Newsletter layouts.
+	 * Name if funky because of 20 character restriction.
```
jeffersonrabb, line 16:
```suggestion
	 * Name is funky because of 20 character restriction.
```

includes/class-newspack-newsletters.php
```diff
@@ -284,6 +294,24 @@ public static function rest_api_init() {
...
+			array(
+				'post_type'      => Newspack_Newsletters_Layouts::NEWSPACK_NEWSLETTERS_LAYOUT_CPT,
+				'posts_per_page' => -1,
+			)
+		);
+		$layouts       = array_merge(
+			$layouts_query->get_posts(),
+			Newspack_Newsletters_Layouts::get_default_layouts()
+		);
+
```
jeffersonrabb, line 311: Could we throw `apply_filters( 'newspack_newsletters_templates', [] )` back in so plugins and themes can add their own layouts?

includes/class-newspack-newsletters-layouts.php
adekbadek: Fixed in c35826a7f6e6264189d2d1801e01c78ec2e699f9

includes/class-newspack-newsletters.php
adekbadek: Done in c35826a7f6e6264189d2d1801e01c78ec2e699f9

adekbadek:
@jeffersonrabb Please confirm if ce91f53d11b99f6f7f0cef7de368dd9df9725ce3 solves #216

jeffersonrabb:
> @jeffersonrabb Please confirm if [ce91f53](https://github.com/Automattic/newspack-newsletters/commit/ce91f53d11b99f6f7f0cef7de368dd9df9725ce3) solves #216

Confirmed!

PR #176, opened 2020-04-30
Handle color palette as it appears in the editor

### LINKED ISSUE #173: Gutenberg color palette handling
New version of Gutenberg has at least one changed color in the color palette. This plugin uses a hardcoded color palette, which means that it ought to be updated when Gutenberg changes the colors. 
There must be a better way! If the editor color palette lives on the frontend of the editor only, we might have to send it over when saving the post, so that the string colors can be translated to hex values.

adekbadek:
### All Submissions:

* [X] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [X] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

Closes #173

1. On `master` branch, create a newsletter and set color of something to "Vivid Purple"*
2. Send or preview email, observe that the color is not handled
3. Switch to this branch
3. Send or preview email, observe that the color is there

*
image

### Other information:

* [X] Have you added an explanation of what your changes do and why you'd like us to include them?
* [X] Have you written new tests for your changes, as applicable?

### DIFF
--- a/includes/class-newspack-newsletters-renderer.php
+++ b/includes/class-newspack-newsletters-renderer.php
@@ -14,6 +14,13 @@ require_once NEWSPACK_NEWSLETTERS_PLUGIN_FILE . '/vendor/autoload.php';
  */
 final class Newspack_Newsletters_Renderer {
 	/**
+	 * The color palette to be used.
+	 *
+	 * @var Object
+	 */
+	protected static $color_palette = null;
+
+	/**
 	 * Convert a list to HTML attributes.
 	 *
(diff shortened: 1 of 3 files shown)


adekbadek:
I know, but Gutenberg is built with the assumption that the editor and site frontend can share a stylesheet – which is not the case in this plugin.

Issue #203, opened 2020-05-08
No Campaign ID Error After Initial Setup


jeffersonrabb:
After creating a first newsletter and inputting API keys, the following message is seen upon closing the layout selection modal: `No Mailchimp campaign ID found for this Newsletter`

### To reproduce

- Navigate to Newsletters->Settings.
- Clear Mailchimp API key and Save
- Add a new Newsletter.
- Input Mailchimp API Key.
- Choose a Layout.
- Observe error message in editor.

jeffersonrabb:
From @iamstevebeatty:

> I don’t get any errors if I change my workflow a bit. After installing and activating the plugin, i go straight to Newsletter --> Settings and enter my info. Things work as expected after that. That settings page is barebones, not like the helpful one Jeff set up with links to get keys and such. Seems like if we require the settings to be entered before you can hit “add new,” that’d do the trick. And maybe change the menu to Setup, so it has that one-time feel to it. This was successful on the newspack.blog and on the model home.

Issue #202, opened 2020-05-08
WP.com Business Site: New Mailchimp Campaign Created With Every Save


jeffersonrabb:
@iamstevebeatty reports that on a WP.com Business site (https://newspack.blog/) saving a Newsletter creates a new MC campaign every time. This suggests the Mailchimp campaign ID is not being persisted properly in options.

PR #208, opened 2020-05-10
fix: situation that overwrote mc campaign id

jeffersonrabb:
### All Submissions:

This PR addresses a regression introduced in https://github.com/Automattic/newspack-newsletters/pull/176. When creating the very first Newsletter, a sequence occurs which clears the Mailchimp Campaign ID field which leads to a confusing error message (`No Mailchimp campaign ID found for this Newsletter`) and creation of duplicate Mailchimp campaigns. The `useEffect` that sets the `color_palette` meta field executes before there is a campaign ID set. This leads to a sequence where the campaign ID is set, then unset, and multiple campaigns are created.

One side effect of this fix is that the colors may be incorrect the very first time a campaign is synced, but the post will always be saved again before testing or sending so this is not a serious issue. 

Closes https://github.com/Automattic/newspack-newsletters/issues/203
Closes https://github.com/Automattic/newspack-newsletters/issues/202

1. On `master` navigate to Newsletter->Settings, clear the Mailchimp key, and save.
2. Create a new Newsletter, paste in the MC key, choose a template. 
3. Observe `No Mailchimp campaign ID found for this Newsletter` error message.
4. Make changes and save the Newsletter.
5. In Mailchimp, observe that multiple campaigns have been created. 
6. Apply the patch and repeat all steps. 
7. Observe there is no error message and only one campaign is created.
8. Verify color palette aspects continue to work as described in https://github.com/Automattic/newspack-newsletters/pull/176. 

### Other information:

### DIFF
--- a/src/newsletter-editor/editor/index.js
+++ b/src/newsletter-editor/editor/index.js
@@ -31,7 +31,6 @@ const Editor = compose( [
 
 		return {
 			isCleanNewPost: isCleanNewPost(),
-			meta,
 			postId: getCurrentPostId(),
 			isReady: meta.newsletterValidationErrors
 				? meta.newsletterValidationErrors.length === 0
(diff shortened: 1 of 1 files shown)


jeffersonrabb:
> This solution works fine, but I'd like to propose an alternative: [931d1f5](https://github.com/Automattic/newspack-newsletters/commit/931d1f50a7d7bf5c15d093d6c7e707d6042920f7)
> I think it's more readable.

Yup, I like this better. Merging...

Issue #198, opened 2020-05-07
Add a fonts setting


thomasguillot:
We could add a panel in the sidebar to allow users to pick the fonts they want -- carefully selected web-safe fonts -- for headings and the body.

At the moment we only support a non-web-safe font: Ubuntu (MJML default).

![1 init](https://user-images.githubusercontent.com/177929/81283642-e1502500-9054-11ea-910e-ae1269ca6fcb.png)
![2 select](https://user-images.githubusercontent.com/177929/81283646-e2815200-9054-11ea-8c96-c64c06ff305f.png)
![3 change-body](https://user-images.githubusercontent.com/177929/81283648-e3b27f00-9054-11ea-8a4a-34b74d1970da.png)
![4 change-heading](https://user-images.githubusercontent.com/177929/81283652-e44b1580-9054-11ea-9103-2ad5f3390244.png)

kinseywilson:
A couple of quick observations about visual display of the newsletter. The header should be constrained to the width of the newsletter itself, and not span to the with of the browser display:

![image](https://user-images.githubusercontent.com/37083693/83355619-9614f380-a32e-11ea-9450-70ab529c04f0.png)

kinseywilson:
This is the standard used by most news websites, large and small:

![image](https://user-images.githubusercontent.com/37083693/83355651-cbb9dc80-a32e-11ea-91c8-0a569d71db6a.png)

![image](https://user-images.githubusercontent.com/37083693/83355664-e55b2400-a32e-11ea-8530-d698217bf385.png)

jeffersonrabb:
@kinseywilson, notes on the header display probably make more sense in a separate issue, or possibly in https://github.com/Automattic/newspack-newsletters/issues/211 (which would then need to be reopened).

kinseywilson:
> 

The shaded section directly below this opening part was creating using the Post Inserter, devised by our own Jefferson Rabb. We asked it to pull the four most recent items published on newspack.blog, and it dropped in images, headlines and text with a couple of clicks.
--

This is from the first version of the newsletter Steve sent out using the plugin. The body copy of the items pulled from the inserter needs to be styled in a way that is much more consistent with the rest of the news letter. Here's what Post Inserter copy looks like (I shrunk it down considerably to capture the full contents):

![image](https://user-images.githubusercontent.com/37083693/83355819-ca3ce400-a32f-11ea-9ddd-438a18270880.png)

Versus other elements:
[...]

PR #210, opened 2020-05-18
feat: font picker

jeffersonrabb:
### All Submissions:

Add font selection to the Newsletters plugin. Adds a Typography sidebar panel with `SelectControl`s for choosing body and header fonts. The options are all web-safe fonts, described in screenshots in https://github.com/Automattic/newspack-newsletters/issues/198. Selected fonts will immediately be used in the editor and in rendered emails. 

Note: the editor font switching uses CSS variables, which are not supported in IE11: https://caniuse.com/#feat=css-variables.

Closes https://github.com/Automattic/newspack-newsletters/issues/198

Screen Shot 2020-05-26 at 8 45 54 AM

1. Create a Newsletter.
2. Experiment with different changes to the Header and Body fonts in the Typography sidebar panel.
3. Observe the typography in the editor changes to reflect the selections.
4. Send yourself a test email, observe the correct fonts are used.
5. View test emails in a variety of email clients. Verify the correct fonts are seen everywhere.

### Other information:

### DIFF
--- a/includes/class-newspack-newsletters.php
+++ b/includes/class-newspack-newsletters.php
@@ -28,7 +28,6 @@ final class Newspack_Newsletters {
 		'Tahoma',
 		'TrebuchetMS',
 		'Verdana',
-		'Serif',
 		'Georgia',
 		'Palatino',
 		'TimesNewRoman',
--- a/includes/class-newspack-newsletters.php
+++ b/includes/class-newspack-newsletters.php
@@ -19,6 +19,23 @@ final class Newspack_Newsletters {
 	const NEWSPACK_NEWSLETTERS_CPT = 'newspack_nl_cpt';
 
 	/**
+	 * Supported fonts.
+	 *
+	 * @var array
+	 */
+	public static $supported_fonts = [
+		'Arial',
+		'Tahoma',
+		'TrebuchetMS',
+		'Verdana',
+		'Serif',
+		'Georgia',
+		'Palatino',
+		'TimesNewRoman',
+		'Courier',
+	];
+
+	/**
 	 * The single instance of the class.
 	 *
 	 * @var Newspack_Newsletters
@@ -383,17 +400,7 @@ final class Newspack_Newsletters {
 	public static function validate_newsletter_typography_value( $key ) {
 		return in_array(
 			$key,
-			[
-				'Arial',
-				'Tahoma',
-				'TrebuchetMS',
-				'Verdana',
-				'Serif',
-				'Georgia',
-				'Palatino',
-				'TimesNewRoman',
-				'Courier',
-			]
+			self::$supported_fonts
 		);
 	}
 
--- a/includes/class-newspack-newsletters.php
+++ b/includes/class-newspack-newsletters.php
@@ -358,6 +358,10 @@ final class Newspack_Newsletters {
 
 	/**
 	 * Set typography meta.
+	 * The save_post action fires before post meta is updated.
+	 * This causes newsletters to be synced to the ESP before recent changes to custom fields have been recorded,
+	 * which leads to incorrect rendering. This is addressed through custom endpoints to update the typography fields
+	 * as soon as they are changed in the editor, so that the changes are available the next time sync to ESP occurs.
 	 *
 	 * @param WP_REST_Request $request API request object.
 	 */
--- a/includes/class-newspack-newsletters.php
+++ b/includes/class-newspack-newsletters.php
@@ -103,6 +103,28 @@ final class Newspack_Newsletters {
 				'auth_callback'  => '__return_true',
 			]
 		);
+		\register_meta(
+			'post',
+			'font_header',
+			[
+				'object_subtype' => self::NEWSPACK_NEWSLETTERS_CPT,
+				'show_in_rest'   => true,
+				'type'           => 'string',
+				'single'         => true,
+				'auth_callback'  => '__return_true',
+			]
+		);
+		\register_meta(
+			'post',
+			'font_body',
+			[
+				'object_subtype' => self::NEWSPACK_NEWSLETTERS_CPT,
+				'show_in_rest'   => true,
+				'type'           => 'string',
+				'single'         => true,
+				'auth_callback'  => '__return_true',
+			]
+		);
 		/**
 		 * The default color palette lives in the editor frontend and is not
 		 * retrievable on the backend. The workaround is to set it as post meta
--- a/includes/class-newspack-newsletters.php
+++ b/includes/class-newspack-newsletters.php
@@ -24,14 +24,14 @@ final class Newspack_Newsletters {
 	 * @var array
 	 */
 	public static $supported_fonts = [
-		'Arial',
-		'Tahoma',
-		'TrebuchetMS',
-		'Verdana',
-		'Georgia',
-		'Palatino',
-		'TimesNewRoman',
-		'Courier',
+		'Arial, Helvetica, sans-serif',
+		'Tahoma, sans-serif',
+		'Trebuchet MS, sans-serif',
+		'Verdana, sans-serif',
+		'Georgia, serif',
+		'Palatino, serif',
+		'Times New Roman, serif',
+		'Courier, monospace',
 	];
 
 	/**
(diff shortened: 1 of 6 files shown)


jeffersonrabb:
@laurelfulford could you take a look at this PR with two particular questions in mind?

1) Is our list of fonts correct? It's a very safe list, although reading https://templates.mailchimp.com/design/typography/ seems to suggest that there might be opportunities to stretch out a bit more.
2) What is the best strategy for applying the selected fonts to the Gutenberg editor? In the current state of the PR the fonts will be applied to the actual email, but there is no change in the editor.

jeffersonrabb:
@adekbadek before I get too far along on this PR I wanted to check if the approach runs afoul of https://github.com/Automattic/newspack-newsletters/pull/132? This PR creates a new sidebar panel for typography—should this be override-able in other ESP implementations, or should it be universal?

laurelfulford:
> Is our list of fonts correct? It's a very safe list, although reading https://templates.mailchimp.com/design/typography/ seems to suggest that there might be opportunities to stretch out a bit more.

The only potential issue with stretching the font list has more to do with the experience editing the newsletter than those receiving it. If we're relying strictly on system fonts, it means someone will have to have a font installed to see it -- and though this is fine for the emails (we'll have font stacks to make sure there are a few similar options to fall back on if they don't have the first font), if someone creating a newsletter doesn't have the font installed, it won't preview the right one for them.

We could get around this with some education/expectation settings, though, maybe even wrangle some links to download the fonts if we do hear from any publishers. The benefits of stretching the list a bit -- and providing some more variety! -- could be worth the possible bit of friction.

We could also just start with the short "guaranteed" list for now, and build on it after the feature is launched :) 

> What is the best strategy for applying the selected fonts to the Gutenberg editor? In the current state of the PR the fonts will be applied to the actual email, but there is no change in the editor.

Since we're just looking at applying these styles to the editor (vs the front-end of the site), I think it could be a good opportunity to experiment with CSS variables, to follow what's being used in Global Styles in Gutenberg. 

We'd just need a change whatever the variable values are, both on editor load (if they're editing an existing newsletter), and on the fly via JavaScript if they mess with the font picker. 

We're currently not using CSS Variables in the theme because they aren't supported by IE11. It seems unlikely that any publishers would be using IE11 to edit their site, but if it happens it would just mean that they would not see a change in the preview when they changed the fonts.

adekbadek:
> This PR creates a new sidebar panel for typography—should this be override-able in other ESP implementations, or should it be universal?

In the end the font rendering depends on the email client, so I think this feature is universal. 

We should add a note though that it's not guaranteed that the font will be rendered in some email clients.

laurelfulford:
I pushed an update to switch the editor to use CSS variables for the header and body fonts. 

With those updates, you can switch the fonts by adding the following JavaScript to the console:

`document.documentElement.style.setProperty('--body-font', 'chalkboard');`

or 

`document.documentElement.style.setProperty('--header-font', 'georgia');`

The defaults for the variables are both set to an Helvetica/Arial font stack, but when the fonts aren't set to the default, something like the above would need to fire on page load (for editing existing newsletters), and when a new font is selected.

I've also updated the styles so the font changes shouldn't affect the placeholder text (like the "Posts Inserter" header, or blocks like the Image block before you've uploaded anything).

laurelfulford:
> Also, a couple of the blocks -- buttons, image captions -- are picking up the body font in the editor, but not picking up any font in the actual email

Jeff and I chatted about this and decided the email should be considered correct here, so I'll be following up shortly with an update to the editor to fix it there.

jeffersonrabb:
> 'System' as is isn't a font -- when I pick it, the preview switches the font to a serif font in the editor.

Opted to simply remove `System` in https://github.com/Automattic/newspack-newsletters/pull/210/commits/6564dadca32a81f178fbe4b1cb5dd620ad433581.

jeffersonrabb:
> Also, a couple of the blocks -- buttons, image captions -- are picking up the body font in the editor, but not picking up any font in the actual email (there, they're using the original `Ubuntu, Helvetica, Arial, sans-serif` default). I'm not sure which is actually correct -- it might just be that I have it wrong in the editor -- but wanted to confirm before I make any changes:

Fixed in https://github.com/Automattic/newspack-newsletters/pull/210/commits/f212f982bcf3b724038d5c2a6b6d6c9695f88f9c

jeffersonrabb:
> We should add a note though that it's not guaranteed that the font will be rendered as intended in some email clients.

For now we're offering only email-safe fonts as described by Mailchimp, so I would think there's no need for a warning. @adekbadek have you seen cases where these don't render in email clients?

jeffersonrabb:
Resolved an issue where emails sent immediately after a font change do not reflect the change, in https://github.com/Automattic/newspack-newsletters/pull/210/commits/fe0e4b16b331ad755fc9dad795a911587f6f9926. 

Set the default fonts to Arial/Georgia in https://github.com/Automattic/newspack-newsletters/pull/210/commits/3dc85bd508891c553524acd2d2fef37aa6f02664

The PR should be ready for review again cc: @adekbadek @laurelfulford

src/newsletter-editor/typography/index.js
```diff
@@ -0,0 +1,106 @@
...
+		label: __( 'Palatino', 'newspack-newsletters' ),
+	},
+	{
+		value: 'TimesNewRoman',
+		label: __( 'Times New Roman', 'newspack-newsletters' ),
+	},
+	{
+		value: null,
+		label: __( 'Monospace', 'newspack-newsletters' ),
+		disabled: true,
```
adekbadek, line 53: Why are these three disabled option in the select? Also, two of them are not in the backend list of fonts.

includes/class-newspack-newsletters.php
```diff
@@ -282,6 +321,77 @@ public static function rest_api_init() {
...
+						'sanitize_callback' => 'sanitize_text_field',
+					],
+				],
+			]
+		);
+	}
+
+	/**
+	 * Set typography meta.
+	 *
```
adekbadek, line 351: I think a comment should be added here describing the need for a custom endpoint to set meta, because it might not be obvious.

src/newsletter-editor/typography/index.js
jeffersonrabb: `Serif`, `Sans Serif` and `Monotype` are category groupings, not actual fonts. It's meant to match [this screenshot](https://user-images.githubusercontent.com/177929/81283646-e2815200-9054-11ea-8c96-c64c06ff305f.png), although without the indentation. Actually all three of these should be absent from the list of fonts—which was the one you did find?

includes/class-newspack-newsletters.php
jeffersonrabb: Good idea. https://github.com/Automattic/newspack-newsletters/pull/210/commits/121ca5c15dd34231c57abe892fdd52a72a6b09bb

src/newsletter-editor/typography/index.js
adekbadek: [`'Serif'` is on the PHP list](https://github.com/Automattic/newspack-newsletters/pull/210/files#diff-76aefac9a481217d439a23e638846effR31)

adekbadek: > category groupings

`SelectControl` does not support [`optgroup`](https://www.w3schools.com/TAGs/tag_optgroup.asp) :/

jeffersonrabb: > `'Serif'` is on the PHP list

Whoops: https://github.com/Automattic/newspack-newsletters/pull/210/commits/7bb8e8b866aba8c631ff609729fb71859d49b350

jeffersonrabb:
@adekbadek pointed out that the mockups call for `optgroups` in the font `SelectControl`s. The `SelectControl` component has no support for optgroups at the moment (feature request [here](https://github.com/Automattic/newspack-newsletters/compare/add/font-picker...add/font-picker-with-optgroups?expand=1)), so I've added this in https://github.com/Automattic/newspack-newsletters/pull/210/commits/f487d7f5a0ed51dfab53e217808e47a174cb071a

PR #223, opened 2020-06-12
Add background color setting

### LINKED ISSUE #219: Background color setting
Currently, there's no background color explicitly set on the newsletter's HTML. Some email clients might thus render a newsletter with dark background, which may result in a bad experience if a newsletter has transparent images. 

There should be a setting for background color, with white as default (as it appears in the editor). Ideally, it should also allow for transparency, if a user intends to keep the dark-mode-supporting UX.

adekbadek:
### All Submissions:

Add setting for background color. During testing I've found that GMail and Outlook on iOS in dark mode disregard a white background color even if set, there's nothing we can do about it I think. 

Closes #219.

1. Create a newsletter. Observe the "Typography" panel in sidebar is now named "Styling"
3. Set a background color, observe the editor background updating
3. Send a test email, observe the background color is as set in the settings

### Other information:


src/newsletter-editor/styling/index.js
```diff
@@ -72,37 +73,54 @@ export default compose( [
...
 			/>
 			<SelectControlWithOptGroup
-				label={ __( 'Body', 'newspack-newsletters' ) }
+				label={ __( 'Body font', 'newspack-newsletters' ) }
 				value={ fontBody || 'Georgia' }
 				optgroups={ fontOptgroups }
-				onChange={ value => updateFontValue( 'font_body', value ) }
+				onChange={ value => updateStyleValue( 'font_body', value ) }
 			/>
+			<BaseControl label={ __( 'Background Color', 'newspack-newsletters' ) } id={ id }>
```
jeffersonrabb, line 116:
```suggestion
			<BaseControl label={ __( 'Background color', 'newspack-newsletters' ) } id={ id }>
```

adekbadek: Fixed: 
8fb1840

adekbadek:
> The typography and background color should be saved with user defined layouts.

Created #225 

> Some of the group blocks in the default templates have solid white backgrounds. We should remove these so the page background is always visible.

Done in 0fd4468931b7c7dfe94db30afa102973e126d2f2

PR #232, opened 2020-06-22
Service Provider Refactor

jeffersonrabb:
### All Submissions:

This is a refactor of the Mailchimp-related code to prepare for adding support for other ESPs. All ESP code is moved to a new class `Newspack_Newsletters_Service_Provider` and each service provider's implementation lives in a descendant class. 

There are eight abstract methods that need to be implemented in each ESP class (`list`|`retrieve`|`save`|`send`|`sender`|`sync`|`test`|`trash`). Some of these correspond to API endpoints (`list`|`retrieve`|`sender`|`test`|), and others are triggered by WordPress hooks (`save`|`send`|`trash`). The ESP class can also implement its own methods and endpoints if needed, e.g. for list segmentation and other ESP-specific activity.

This is a broad code refactor but it should not impact any functionality. Functional testing should ideally cover everything the plugin does: setting/changing keys, creating newsletters, saving templates, sending test emails, sending campaigns.

### Other information:


includes/service-providers/class-newspack-newsletters-service-provider.php
```diff
@@ -0,0 +1,278 @@
...
+	 * @param string $reply_to Reply to email address.
+	 * @return object|WP_Error API Response or error.
+	 */
+	abstract public function sender( $post_id, $from_name, $reply_to );
+
+	/**
+	 * Synchronize post with corresponding ESP campaign.
+	 *
+	 * @param WP_POST $post Post to synchronize.
+	 * @return object|null API Response or error.
```
adekbadek, line 90: Shouldn't it be
```suggestion
	 * @return object|WP_Error API Response or error.
```
?

includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php
```diff
@@ -0,0 +1,614 @@
...
+		if ( ! $mc_campaign_id ) {
+			return new WP_Error(
+				'newspack_newsletters_no_campaign_id',
+				__( 'Mailchimp campaign ID not found.', 'newspack-newsletters' )
+			);
+		}
+		try {
+			$mc      = new Mailchimp( $this->api_key() );
+			$payload = [
+				'test_emails' => $test_emails,
```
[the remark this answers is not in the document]
jeffersonrabb: Fixed in https://github.com/Automattic/newspack-newsletters/pull/232/commits/4e9a31ce9a3baf74e3f07fdc773483207267d175

includes/service-providers/class-newspack-newsletters-service-provider.php
jeffersonrabb: I used `null` because of this possible outcome: https://github.com/Automattic/newspack-newsletters/blob/4e9a31ce9a3baf74e3f07fdc773483207267d175/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php#L340. Because of this approach to storing errors for future notification perhaps https://github.com/Automattic/newspack-newsletters/blob/4e9a31ce9a3baf74e3f07fdc773483207267d175/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php#L287-L292 should be changed to persist the error and return `null`?

jeffersonrabb:
> 1. [Settings page](https://github.com/Automattic/newspack-newsletters/blob/4729abe110045cb152f7bbf750915ce67895ceee/includes/class-newspack-newsletters-settings.php#L29-L32) and plenty of `mailchimp_api_key` (variable or method) usage – the `api_key` method from this PR should be used there.

https://github.com/Automattic/newspack-newsletters/pull/232/commits/7998bd1b2992245cbc1c53cbae8519ba98a5f809 is a further refactor that moves all of the API key operations into the service provider class. Settings page is a different animal since it uses `register_setting`/`add_settings_field`. Settings and the modal screen where keys are set will need a rethink as soon as we add a second service provider.

jeffersonrabb:
> 2\. The new `Newspack_Newsletters::service_provider()` should be used [here](https://github.com/Automattic/newspack-newsletters/blob/4729abe110045cb152f7bbf750915ce67895ceee/includes/class-newspack-newsletters-editor.php#L125), so we have a single source of truth.

Done in https://github.com/Automattic/newspack-newsletters/pull/232/commits/a00acbbc25f53d3e06d481f9d2a087483819cf65

iuravic:
@jeffersonrabb ,

## Disclaimer & TL;DR

I created a new branch [`refactor/provider-oop-ivan-demo`](https://github.com/Automattic/newspack-newsletters/commits/refactor/provider-oop-ivan-demo) based on your branch, to show commit-by-commit what and how I personally would add. It's up to you to see how far you'd like to use all this, and also how far we'd like to take the backend in the Newsletters Plugin.

I focused almost entirely on the ServiceProvider class. My guiding thought is this one -- every class "should" just only one thing -- I've quoted _should_ because that's a golden rule in full-fledged OOP, but not necessarily for our needs.

## First thoughts about your refactor

### Newspack_Newsletters_Service_Provider

- I love the kind of approach you took with this class. You've employed the the Composite Pattern -- [here](https://designpatternsphp.readthedocs.io/en/latest/Structural/Composite/README.html) or [here](https://www.sitepoint.com/community/t/php-composite-design-pattern/331675). Folks usually warn this pattern is so good, that one tends to overuse it. I don't think you overused it at all. In earlier days it was done with abstract functions, and now it's usually done with interfaces; but the mechanism is identical.
- you moved all the ESP-specific code into this new class, which resulted in decluttering and clearly separating all ESP functionality from any other plugin functionality. This is already great improvement, and IMO the most important one. Even with this, the whole app/plugin is now much, much easier to get involved in, to maintain and extend further. Both thumbs up for this move!
- when looking for improvements/next steps, I first notice that the Service Provider class actually does several things, which is as I said something that's avoided in OOP. Here's all the different things which it does:
	- it defines what all children classes must provide - the abstract functions
	- it is a Controller - `rest_api_init()` defines endpoints, and it has all the `api_*` functions used to handles request, filter and validate params, call and check stuff, give response. And we actually have two different `rest_api_init()` functions in this refactor
 	- it has a few shared functions which all the children classes get to inherit and use - namely just the singleton handling `instance()`, and some validation too. Not much in that layer, but can still be seen as a separate thing

###  Newspack_Newsletters_Editor
- my focus here is on the `Newspack_Newsletters_Service_Provider`, so I'm just gonna notice how the changes you made reflect positively on this class too, since it now holds just the editor functionality

## The steps I continued with

Basically I followed the observations above to bring the following structure, in very standard OOP ways. What I basically did was:
1. replaced abstract functions with interfaces
2. singled out APIs into controllers -- a general Service Provider one (shared by all children), and a specific Mailchimp one
3. a bit tidying up -- where files get included, etc.
4. since that's the main work, I end with some more notes, which might be better handled in a separate PR or review

Here goes. The actual commit is in the title so you can see how it looks like.
And I actually tested this code, and it works well -- sends the newsletters.

## 1. Interfaces -- added in commit 98301ae

Created two Interfaces which replace the abstract functions:
1. `Newspack_Newsletters_ESP_API_Interface` == mandatory API functionality
2. `Newspack_Newsletters_WP_Hookable_Interface` == defines Mandatory integration with WP Hooks

The abstract functions did a great job, mind you, no real problems with those. But the Interfaces are a bit more convenient, main benefits to us at Newspack:
- they define the requirements a more clearly by decoupling functions' contexts -- grouping required functions by their purpose
- they code-document exactly what you explained in your PR description -- _**"There are eight abstract methods that need to be implemented in each ESP class (list|retrieve|save|send|sender|sync|test|trash). Some of these correspond to API endpoints (list|retrieve|sender|test|), and others are triggered by WordPress hooks (save|send|trash)."**_ -- all this can now be read directly from the code. So they improve code readability
- they make some code more reusable
- future extensibility -- e.g. we may want to add a third interface later on, perhaps which only certain ESP Service need to adopt, etc.

An extremely exciting and interesting digression -- but Interfaces go much beyond this in full-fledged PHP OOP. Their most important role is when Type Hinting is used, when they enable real Dependency Injection, which is the core of PHP flavor OOP.

## 2. Singleton use -- added in commit 5fa0acf

This is a one liner. No need to call this one at the bottom of the file. Every time we call the `instance()` method, we'll always get its singleton.

## 3. Including dependencies -- added in commit da85c1d

Here I moved all main plugin file includes to the main plugin file. At this point needed to add two more classes, and it became a bit unclear where to include them, since files were included in 4 different places. There were also some include repeats.

Including files is basic plugin bootstrapping, and it could maybe be easiest to manage when it's all done just once in one place -- in the main plugin file. Which is exactly how Composer works -- the single `vendor/autoload.php` file handles _all_ the files and libs in the `vendor/`.

## 4. Controllers -- added in commit 6b44541

This is the largest and most complex commit, but it finalizes the refactor:

image

What I did:
- all the Mailchimp-specific logic got moved out of the `ServiceProvider` and into the `Mailchimp class`,
- and all the API stuff got moved out to the two controllers -- the
[Newspack_Newsletters_Mailchimp_Controller](https://github.com/Automattic/newspack-newsletters/blob/refactor/provider-oop-ivan-demo/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp-controller.php) got the MC API code, while the general API stuff should be pulled out into the [Newspack_Newsletters_ESP_Service_Provider_Controller](https://github.com/Automattic/newspack-newsletters/blob/refactor/provider-oop-ivan-demo/includes/service-providers/class-newspack-newsletters-service-provider-controller.php). I had trouble differentiating this code well enough, so for now I put all the endpoints into the `Mailchimp Controller`, but some stuff should go into the general `Service Provider Controller`

The controllers should contain only the API endpoints and, and only the light API-specific stuff -- get request params, filter and validate params, call stuff, check, route and return response. All the "business logic" should go to the ServiceProvider (general logic) i.e.
Mailchimp (Mailchimp-specific logic):
- and that's why the controller gets a service in its constructor -- example [here](https://github.com/Automattic/newspack-newsletters/blob/refactor/provider-oop-ivan-demo/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp-controller.php#L27) -- the `service` has all the "business logic" injected in Controllers instructor. In Controller code, that class simply makes calls to the "intelligent" service, and keeps focus on API light and breezy
- thi previous step is the most basic form of [Dependency Injection pattern](https://designpatternsphp.readthedocs.io/en/latest/Structural/DependencyInjection/README.html).

## 5. Remaining possible improvements

Here's a few more smaller possible improvements:
- the [`Newspack_Newsletters`](https://github.com/Automattic/newspack-newsletters/blob/refactor/provider-oop-ivan-demo/includes/class-newspack-newsletters.php) class could also get its API extracted into a controller.
- this if-else block [right here](https://github.com/Automattic/newspack-newsletters/blob/refactor/provider-oop-ivan-demo/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php#L316) would be ideal to single out in a separate function called `get_campaign_result()`. If I saw `$this->get_campaign_result()` in the code, I would'nt have to spend a second thinking about what it does -- and like this I need to read and study it. Similarly, breaking down a long function into _readable_ calls is very OOP-like.
- naming -- I'd propose just a few naming changes, to make the function 100% understandable even without reading any of its content. Examples: `api_key()` -> `get_api_key()`, `retrieve( $post_id )` -> `retrieve_newsletter( $post_id )`, `test( $post_id, $emails )` -> `send_test_newsletter( $post_id, $emails )`, ...
- I haven't tried this one -- now that the number of classes is growing, we might use a better directory structure, just like you created the `includes/service-providers` folder, we could perhaps use a `controllers` folder too

includes/service-providers/class-newspack-newsletters-service-provider.php
adekbadek: Ah, that's true that there's a persisted error there, not returned one. I think it's ok. My comment was prompted more by the `or error` wording :)

jeffersonrabb:
Closing in favor of https://github.com/Automattic/newspack-newsletters/pull/247

PR #233, opened 2020-06-23
Add self-serve ads

### LINKED ISSUE #166: Email ads
The News Project has a newsletter ads feature that seems very similar to the popups feature, but for newsletters. The user builds a newsletter ad using the block editor, and configures where it gets inserted (X paragraphs) and until when it will run (see bottom left of picture). 

screen-shot-2020-04-06-at-2 11 41-pm

Newsletter ads tend to have pretty good rates, and a feature like this could really help drive publisher revenue.

There's quite a bit of flexibility in how this feature could be implemented. A block pattern could possibly work. In v1, I don't think we need to support GAM email ads, and I doubt any of our publishers are using those. An image with a link is probably sufficient.

adekbadek:
### All Submissions:

Adds self-serve ads. Note that the ads will be placed at the bottom of the email – setting placement will be handled in another PR (issue: #237 ), as this one is pretty big already.

Closes #166 

1. Observe a new item under "Newsletters" menu in admin - "Ads". Visit that page:

image

2. Click "Create new ad". Create the content and try to publish – observe that it's not possible until an expiry date is set in the sidebar. If an expiry date is set in the past, a warning notice should be shown in the pre-publish panel.
3. Go back to the "Ads" view – observe the new ad being listed there. 
4. Create a new newsletter and send a test email. Observe the ad being appended to the bottom of the email (but above the ESP footer)
5. Send the newsletter – observe a notice in the pre-send modal about the ads which will be inserted. 
6. Observe the ad being placed in the email, just like in the test email.
7. Create a couple of ads which are drafts, published, expired – to test the UI in "Ads" view

image

### Other information:

### DIFF
--- a/includes/class-newspack-newsletters-renderer.php
+++ b/includes/class-newspack-newsletters-renderer.php
@@ -529,12 +529,13 @@ final class Newspack_Newsletters_Renderer {
 	}
 
 	/**
-	 * Convert a WP post to MJML markup.
+	 * Convert a WP post to MJML components.
 	 *
 	 * @param WP_Post $post The post.
-	 * @return string MJML markup.
+	 * @param Boolean $include_ads Whether to include ads.
+	 * @return string MJML markup to be injected into the template.
 	 */
-	private static function render_mjml( $post ) {
+	private static function post_to_mjml_components( $post, $include_ads ) {
 		self::$color_palette = get_post_meta( $post->ID, 'color_palette', true );
 		self::$font_header   = get_post_meta( $post->ID, 'font_header', true );
 		self::$font_body     = get_post_meta( $post->ID, 'font_body', true );
@@ -543,11 +544,6 @@ final class Newspack_Newsletters_Renderer {
 		}
 		if ( ! in_array( self::$font_body, Newspack_Newsletters::$supported_fonts ) ) {
 			self::$font_body = 'Georgia';
-		}
-		$title            = $post->post_title;
-		$background_color = get_post_meta( $post->ID, 'background_color', true );
-		if ( ! $background_color ) {
-			$background_color = '#ffffff';
 		}
 		$blocks = parse_blocks( $post->post_content );
 		$body   = '';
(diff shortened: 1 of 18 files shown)


includes/class-newspack-newsletters-renderer.php
```diff
@@ -558,6 +554,42 @@ private static function render_mjml( $post ) {
...
+			);
+			$ads        = $ads_query->get_posts();
+			$ads_markup = '';
+			foreach ( $ads as $ad ) {
+				$expiry_date = new DateTime( get_post_meta( $ad->ID, 'expiry_date', true ) );
+				$now_date    = new DateTime();
+				if ( $expiry_date > $now_date ) {
+					$ads_markup .= self::post_to_mjml_components( $ad, false );
+				}
+			}
```
claudiulodro, line 573: You could use [the meta query options for `WP_Query`](https://developer.wordpress.org/reference/classes/wp_query/#custom-field-post-meta-parameters) to make it so that the WP_Query only fetches valid ads, and then we don't have to loop through them all checking if they're valid. The query would only fetch newsletter ad posts with `expiry_date` meta greater-than-or-equal-to the current date. I'm not sure whether there's a good way to do that currently since the meta values look like `2020-06-29T11:37:00`, but it would be trivial if the meta were stored as unix timestamps  because it could just compare the integer values of the expiry timestamp to the timestamp from the current day at `00:00:00`.

Something like this is what I'm trying to explain here. :) 
```
array(
    'post_type'      => Newspack_Newsletters_Ads::NEWSPACK_NEWSLETTERS_ADS_CPT,
    'posts_per_page' => -1,
    'meta_key' => 'expiry_date',
    'meta_value_num' => $start_of_today,
    'meta_compare' => '>=',
)

```

I don't think the current implementation will cause any issues since there won't be thousands of newsletter ads to iterate through, but just a potential improvement.

adekbadek: Fixed in 3bbc88a – and also ensured that in both editor and on the backend the expiry date is handled in the same way – comparing full days.

claudiulodro: Good solution! I didn't realize we could do meta queries by date. :)

jeffersonrabb:
@claudiulodro deletion ui added in https://github.com/Automattic/newspack-newsletters/pull/233/commits/ed518ad9211ce282441b5b7c476e0922f8d7a920 — could you give this one more look?

PR #241, opened 2020-06-25
Mailchimp: tags segmentation

### LINKED ISSUE #21: Support For Mailchimp Tags
Add support for Mailchimp Tags: https://us7.admin.mailchimp.com/lists/members?id=453669&previewName=Contacts#p:1-s:10-sa:last_update_time-so:false

adekbadek:
### All Submissions:

Adds Mailchimp segmentation via tags. Tag creation will be handled in a separate PR. 

Closes #21

1. Create a newsletter, after selecting a list observe a new input below Groups selection – Tags
2. (Create some tags in Mailchimp if there are none)
3. Start typing a tag name, observe autocomplete in action, choose a tag
4. Observe the tag is added to the campaign on Mailchimp

### Other information:

### DIFF
--- a/includes/class-newspack-newsletters.php
+++ b/includes/class-newspack-newsletters.php
@@ -884,7 +884,7 @@ final class Newspack_Newsletters {
 			$tags = [];
 			if ( $list_id ) {
 				$tags_response = self::validate_mailchimp_operation(
-					$mc->get( "lists/$list_id/segments" ),
+					$mc->get( "lists/$list_id/segments?count=1000" ),
 					__( 'Error retrieving Mailchimp tags.', 'newspack_newsletters' )
 				);
 				$tags          = $tags_response['segments'];
--- a/includes/class-newspack-newsletters.php
+++ b/includes/class-newspack-newsletters.php
@@ -283,8 +283,14 @@ final class Newspack_Newsletters {
 				'callback'            => [ __CLASS__, 'api_set_mailchimp_segments' ],
 				'permission_callback' => [ __CLASS__, 'api_authoring_permissions_check' ],
 				'args'                => [
-					'id' => [
+					'id'          => [
 						'sanitize_callback' => 'absint',
+					],
+					'interest_id' => [
+						'sanitize_callback' => 'esc_attr',
+					],
+					'tag_ids'     => [
+						'sanitize_callback' => 'wp_parse_list',
 					],
 				],
 			]
--- a/includes/class-newspack-newsletters.php
+++ b/includes/class-newspack-newsletters.php
@@ -720,7 +720,7 @@ final class Newspack_Newsletters {
 
 			if ( $has_interest || $tag_ids_param ) {
 				$segment_opts = [
-					'match'      => 'any',
+					'match'      => 'all',
 					'conditions' => [],
 				];
 
--- a/includes/class-newspack-newsletters.php
+++ b/includes/class-newspack-newsletters.php
@@ -685,8 +685,8 @@ final class Newspack_Newsletters {
 			$is_unsetting_interest = 'no_interests' === $interest_id_param;
 			if ( ! $is_unsetting_interest && ( ! $field || ! $interest_id ) ) {
 				return new WP_Error(
-					'newspack_newsletters_incorrect_post_type',
-					__( 'Invalid Mailchimp Interest .', 'newspack-newsletters' )
+					'newspack_newsletters_invalid_mailchimp_interest',
+					__( 'Invalid Mailchimp Interest.', 'newspack-newsletters' )
 				);
 			}
 		}
--- a/includes/class-newspack-newsletters.php
+++ b/includes/class-newspack-newsletters.php
@@ -277,17 +277,14 @@ final class Newspack_Newsletters {
 		);
 		\register_rest_route(
 			'newspack-newsletters/v1',
-			'mailchimp/(?P<id>[\a-z]+)/interest/(?P<interest_id>[\a-z]+)',
+			'mailchimp/(?P<id>[\a-z]+)/segments',
 			[
 				'methods'             => \WP_REST_Server::EDITABLE,
-				'callback'            => [ __CLASS__, 'api_set_mailchimp_interest' ],
+				'callback'            => [ __CLASS__, 'api_set_mailchimp_segments' ],
 				'permission_callback' => [ __CLASS__, 'api_authoring_permissions_check' ],
 				'args'                => [
-					'id'          => [
+					'id' => [
 						'sanitize_callback' => 'absint',
-					],
-					'interest_id' => [
-						'sanitize_callback' => 'esc_attr',
 					],
 				],
 			]
@@ -664,11 +661,8 @@ final class Newspack_Newsletters {
 	 *
 	 * @param WP_REST_Request $request API request object.
 	 */
-	public static function api_set_mailchimp_interest( $request ) {
-		$id          = $request['id'];
-		$exploded    = explode( ':', $request['interest_id'] );
-		$field       = count( $exploded ) ? $exploded[0] : null;
-		$interest_id = count( $exploded ) > 1 ? $exploded[1] : null;
+	public static function api_set_mailchimp_segments( $request ) {
+		$id = $request['id'];
 
 		if ( self::NEWSPACK_NEWSLETTERS_CPT !== get_post_type( $id ) ) {
 			return new WP_Error(
@@ -677,11 +671,18 @@ final class Newspack_Newsletters {
 			);
 		}
 
-		if ( 'no_interests' !== $request['interest_id'] && ( ! $field || ! $interest_id ) ) {
-			return new WP_Error(
-				'newspack_newsletters_incorrect_post_type',
-				__( 'Invalid Mailchimp Interest .', 'newspack-newsletters' )
-			);
+		$interest_id_param = $request['interest_id'];
+		if ( $interest_id_param ) {
+			$exploded              = explode( ':', $interest_id_param );
+			$field                 = count( $exploded ) ? $exploded[0] : null;
+			$interest_id           = count( $exploded ) > 1 ? $exploded[1] : null;
+			$is_unsetting_interest = 'no_interests' === $interest_id_param;
+			if ( ! $is_unsetting_interest && ( ! $field || ! $interest_id ) ) {
+				return new WP_Error(
+					'newspack_newsletters_incorrect_post_type',
+					__( 'Invalid Mailchimp Interest .', 'newspack-newsletters' )
+				);
+			}
 		}
 
 		$mc_campaign_id = get_post_meta( $id, 'mc_campaign_id', true );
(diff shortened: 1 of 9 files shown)


includes/class-newspack-newsletters.php
```diff
@@ -277,18 +277,15 @@ public static function rest_api_init() {
...
 				'methods'             => \WP_REST_Server::EDITABLE,
-				'callback'            => [ __CLASS__, 'api_set_mailchimp_interest' ],
+				'callback'            => [ __CLASS__, 'api_set_mailchimp_segments' ],
 				'permission_callback' => [ __CLASS__, 'api_authoring_permissions_check' ],
 				'args'                => [
-					'id'          => [
+					'id' => [
 						'sanitize_callback' => 'absint',
 					],
-					'interest_id' => [
```
jeffersonrabb, line 289: Why is this arg removed? It's still being used in https://github.com/Automattic/newspack-newsletters/pull/241/commits/e15b9e734c8a1409a52dfc66ef1282bab89bdbc4#diff-76aefac9a481217d439a23e638846effR674

includes/class-newspack-newsletters.php
```diff
@@ -277,18 +277,15 @@ public static function rest_api_init() {
...
 		\register_rest_route(
 			'newspack-newsletters/v1',
-			'mailchimp/(?P<id>[\a-z]+)/interest/(?P<interest_id>[\a-z]+)',
+			'mailchimp/(?P<id>[\a-z]+)/segments',
 			[
 				'methods'             => \WP_REST_Server::EDITABLE,
-				'callback'            => [ __CLASS__, 'api_set_mailchimp_interest' ],
+				'callback'            => [ __CLASS__, 'api_set_mailchimp_segments' ],
 				'permission_callback' => [ __CLASS__, 'api_authoring_permissions_check' ],
 				'args'                => [
```
jeffersonrabb, line 285: `tag_ids` should be listed as well.

includes/class-newspack-newsletters.php
```diff
@@ -677,11 +671,18 @@ public static function api_set_mailchimp_interest( $request ) {
...
-			);
+		$interest_id_param = $request['interest_id'];
+		if ( $interest_id_param ) {
+			$exploded              = explode( ':', $interest_id_param );
+			$field                 = count( $exploded ) ? $exploded[0] : null;
+			$interest_id           = count( $exploded ) > 1 ? $exploded[1] : null;
+			$is_unsetting_interest = 'no_interests' === $interest_id_param;
+			if ( ! $is_unsetting_interest && ( ! $field || ! $interest_id ) ) {
+				return new WP_Error(
+					'newspack_newsletters_incorrect_post_type',
```
jeffersonrabb, line 682:
```suggestion
					'newspack_newsletters_invalid_mailchimp_interest',
```

includes/class-newspack-newsletters.php
```diff
@@ -677,11 +671,18 @@ public static function api_set_mailchimp_interest( $request ) {
...
+					__( 'Invalid Mailchimp Interest .', 'newspack-newsletters' )
```
jeffersonrabb, line 683:
```suggestion
					__( 'Invalid Mailchimp Interest.', 'newspack-newsletters' )
```

includes/class-newspack-newsletters.php
```diff
@@ -706,22 +707,40 @@ public static function api_set_mailchimp_interest( $request ) {
 				);
 			}
 
-			$segment_opts = ( 'no_interests' === $request['interest_id'] ) ?
-				(object) [] :
-				[
+			$has_interest  = $interest_id_param && ! $is_unsetting_interest;
+			$tag_ids_param = $request['tag_ids'];
+
+			$segment_opts = (object) [];
```
jeffersonrabb, line 713: Why the cast?

includes/class-newspack-newsletters.php
```diff
@@ -856,6 +875,15 @@ public static function retrieve_data( $id ) {
 				}
 			}
 
+			$tags = [];
+			if ( $list_id ) {
+				$tags_response = self::validate_mailchimp_operation(
+					$mc->get( "lists/$list_id/segments" ),
```
jeffersonrabb, line 881: According to the [Mailchimp docs](https://mailchimp.com/developer/reference/lists/list-segments/) this will only return up to 10 tags. I don't really have an understanding of how tags are typically used so its hard to guess how many there might be, but it might make sense to make the count pretty high just to be safe.

src/service-providers/mailchimp/ProviderSidebar.js
```diff
@@ -2,13 +2,81 @@
...
+							value: 'no_interests',
+						},
+						...availableInterests,
+					] }
+					onChange={ interest_id => updateSegmentsData( { interest_id } ) }
+					disabled={ inFlight }
+				/>
+			) }
+			{ availableTags.length && (
+				<FormTokenField
```
jeffersonrabb, line 67: The help text beneath the `FormTokenField` reads "Separate with commas or the Enter key," which doesn't seem to apply to this use case? It looks like it can't be changed or removed idiomatically (https://github.com/WordPress/gutenberg/blob/master/packages/components/src/form-token-field/index.js#L660-L669) so we might need to hide with CSS for now.

src/service-providers/mailchimp/ProviderSidebar.js
```diff
@@ -2,13 +2,81 @@
...
+							label: __( '-- Select a group --', 'newspack-newsletters' ),
+							value: 'no_interests',
+						},
+						...availableInterests,
+					] }
+					onChange={ interest_id => updateSegmentsData( { interest_id } ) }
+					disabled={ inFlight }
+				/>
+			) }
+			{ availableTags.length && (
```
jeffersonrabb, line 66: If there are no tags at all in the Mailchimp account I get the numeral 0 in the UI:

Screen Shot 2020-07-05 at 12 58 39 PM

src/service-providers/mailchimp/ProviderSidebar.js
jeffersonrabb, line 67: Not sure if there's anything that can be done about this, but when I first start typing in the field I get autocomplete UI that seems to be based on browser cache or something, not the list of tags from Mailchimp. In this screenshot, I have no tag named "sports" in Mailchimp:

Screen Shot 2020-07-05 at 2 35 08 PM

includes/class-newspack-newsletters.php
```diff
@@ -706,22 +707,40 @@ public static function api_set_mailchimp_interest( $request ) {
...
+						$segment_opts['conditions'][] = [
+							'condition_type' => 'StaticSegment',
+							'field'          => 'static_segment',
+							'op'             => 'static_is',
+							'value'          => $tag_id,
+						];
+					}
+				}
+			}
+
```
jeffersonrabb, line 743: If you specify an Interest and Tags, Mailchimp will treat the segmentation as OR logic, which would mean if no subscribers match the tags the campaign will still be sent to every subscriber in the Interest. Here's what the segmentation logic in MC looks like for a campaign where I specified both:

Screen Shot 2020-07-05 at 2 44 33 PM

I'm not 100% sure, but I _think_ we would want this segment to be all subscribers in that Interest Group bearing the specified Tags?

includes/class-newspack-newsletters.php
```diff
@@ -706,22 +707,40 @@ public static function api_set_mailchimp_interest( $request ) {
...
+					foreach ( $tag_ids_param as $tag_id ) {
+						$segment_opts['conditions'][] = [
+							'condition_type' => 'StaticSegment',
+							'field'          => 'static_segment',
+							'op'             => 'static_is',
+							'value'          => $tag_id,
+						];
+					}
+				}
+			}
```
jeffersonrabb, line 742: If a Tag segment is specified, the plugin will activate the Send button even if there are 0 recipients. Ideally the campaign can only be sent if the segmentation logic includes at least one subscriber, and the pre-send modal should indicate the recipient count. Here are screenshots showing the disconnect between the segmentation in Mailchimp vs. the plugin:

Screen Shot 2020-07-05 at 2 52 17 PM
Screen Shot 2020-07-05 at 2 52 12 PM

includes/class-newspack-newsletters.php
adekbadek: To be explicit that `$segment_opts` is an object, not an array – it was the default before (see [removed line 710 above](https://github.com/Automattic/newspack-newsletters/pull/241/files#diff-76aefac9a481217d439a23e638846effL710))

includes/class-newspack-newsletters.php
adekbadek: Fixed in 9067d0f9fd4ff3932b31eca029aedd8d97fb4f0f

includes/class-newspack-newsletters.php
adekbadek: Fixed in 9067d0f9fd4ff3932b31eca029aedd8d97fb4f0f

includes/class-newspack-newsletters.php
adekbadek: Fixed in 90d99c86765f6afd243ede7f5c4d9be397aec354, for some reason I thought that if the payload is in the request body (as opposed to URL param), listing args is not needed.

includes/class-newspack-newsletters.php
adekbadek: Fixed in 90d99c86765f6afd243ede7f5c4d9be397aec354

includes/class-newspack-newsletters.php
adekbadek: Good catch, fixed in df1cc478af262f1dfd3cace2d2297f3479a6a2ab

src/service-providers/mailchimp/ProviderSidebar.js
adekbadek: Fixed in a4c86988655e74e99319feb723fa23acfe5e0a81

adekbadek: Unfortunately the `id` of the rendered `input` is [not overridable](https://github.com/WordPress/gutenberg/blob/059e8bebbc19e67f2685d56dc000b4439eadbebb/packages/components/src/form-token-field/index.js#L694) :/

jeffersonrabb: Oh well, not big deal!

includes/class-newspack-newsletters.php
adekbadek: Treating this as AND logic makes sense, fixed in adb5a8031b587ed3db273aec70da5eb5ee336909

includes/class-newspack-newsletters.php
adekbadek: Fixed to display the actual number of recipients in caa29470a09f267b67896e74a7945f86997dadd9. I've also updated it so only tags with recipients are suggested in the tags field (0fc84692fd0fd23c77cc663f5c19293d33071c81).

adekbadek:
@dkoo 
1. "the dependence on there being Groups before the Tags field is displayed" – that was exactly the issue, fixed in b47d1e525f34dce59f7762bbe48545ab57103816
1. "A minor annoyance: after adding a tag by clicking on an autocomplete result, focus got removed from the input" – I know, I've experimented with debouncing here, but it results in the annoyance being experienced a bit later. We could update MC settings on post saving, would there be any downside to that, @jeffersonrabb ?
1. "the input was in a disabled state, but I was able to click on the tags and focus them, resulting in this slightly funky display" – I've submitted a PR to fix it here: https://github.com/WordPress/gutenberg/pull/23713

jeffersonrabb:
> I'm seeing incorrect subscriber counts in the pre-send modal. In this screenshot note the Groups UI which correctly shows 3 subscribers. There are no Tags specified. The modal incorrectly shows 1 subscriber.

@adekbadek and gave this a closer look and it seems to be a MC cacheing or propagation issue. The mismatch is the same in the Mailchimp dashboard, and after a period of time it seems to correct itself.

jeffersonrabb:
> 2\. We could update MC settings on post saving, would there be any downside to that, @jeffersonrabb ?

Seems fine to me.

PR #247, opened 2020-06-29
Refactor/provider oop ivan demo

iuravic:
### All Submissions:

* [ x ] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [ x ] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

Builds completely on top of the `refactor/provider-oop` and [PR 232](https://github.com/Automattic/newspack-newsletters/pull/232).

Adds next steps for refactoring.

The plugin should still send newsletters.

### Other information:

* [ x ] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ x ] Have you successfully ran tests with your changes locally?

The following explanation is taken from [PR 232's comment](https://github.com/Automattic/newspack-newsletters/pull/232#issuecomment-649775483).

### Newspack_Newsletters_Service_Provider

- I love the kind of approach you took with this class. You've employed the the Composite Pattern -- [here](https://designpatternsphp.readthedocs.io/en/latest/Structural/Composite/README.html) or [here](https://www.sitepoint.com/community/t/php-composite-design-pattern/331675). In earlier days it was done with abstract functions, and now it's usually done with interfaces; but the mechanism is identical.
- you moved all the ESP-specific code into this new class, which resulted in decluttering and clearly separating all ESP functionality from any other plugin functionality. This is already great improvement, and IMO the most important one. Even with this, the whole app/plugin is now much, much easier to get involved in, to maintain and extend further. Both thumbs up for this move!
- when looking for improvements/next steps, I first notice that the Service Provider class actually does several things:
	- it defines what all children classes must provide - the abstract functions
	- it is a Controller - `rest_api_init()` with the `api_*` functions
 	- it has a few shared functions which all the children classes get - just the singleton handling `instance()`, and some validation here. Not much in that layer, but can still be seen as a separate thing

###  Newspack_Newsletters_Editor
- my focus here is on the `Newspack_Newsletters_Service_Provider`, so I'll just notice how the changes made reflect positively on this class too, and it's now decluttered and holds just the editor functionality

## The steps I continued with

The actual commit is in the title so you can see how it looks like.
I actually tested this code, and it works well -- this branch still sends the newsletters.

## 1. Interfaces -- added in commit 98301ae

Created two Interfaces which replace the abstract functions:
1. `Newspack_Newsletters_ESP_API_Interface` == mandatory API functionality
2. `Newspack_Newsletters_WP_Hookable_Interface` == defines Mandatory integration with WP Hooks

The abstract functions did a great job, mind you, no real problems with those. But the Interfaces are a bit more convenient, main benefits in our case:
- they code-document exactly what you explained in your PR description -- _**"There are eight abstract methods that need to be implemented in each ESP class (list|retrieve|save|send|sender|sync|test|trash). Some of these correspond to API endpoints (list|retrieve|sender|test|), and others are triggered by WordPress hooks (save|send|trash)."**_ -- this can now be read directly from the code. So they improve code readability.
- they define the requirements a more clearly by decoupling functions' contexts -- grouping required functions by their purpose
- they make some code more reusable
- future extensibility -- e.g. we may later on want to add a third interface, which only certain ESP Service need to adopt, etc.

An extremely exciting and interesting digression -- in PHP's flavor of full-fledged OOP Interfaces serve for much beyond this. When Type Hinting is used, they enable "real" Dependency Injection. And finally, a Service Container pattern can get used, which can completely re-shape and simplify the whole app structure. End digression, moving on...

## 2. Singleton use -- added in commit 5fa0acf

This is a one liner. No need to call this one at the bottom of the file. Every time we call the `instance()` method, we'll always get its singleton.

I haven't been using [this particular "inherited-singleton" code](https://github.com/Automattic/newspack-newsletters/pull/232/files#diff-d40f89304ebfb31e46dbc533b9cd69c7R141). It's pretty nice that the singleton behavior can be passed down like this. Sharing and noticing that it has one flaw when compared to the standard Singleton pattern. In order to make [this line here](https://github.com/Automattic/newspack-newsletters/pull/232/files#diff-d40f89304ebfb31e46dbc533b9cd69c7R143) work, we must keep the `__construct()` public, while it's always made private in standard Singleton. Why? To disable calling `new $class()` on our own, and force us to use `$class::instance()`. Creating a new class with `new` here, one could break the singleton, and instantiate the class apart from it. If, for example, it were some important state object, sharing some dynamic value across the app, we couldn't allow this. But I don't think it matters very much in our case, it does a nice job, and I vote that we keep it.

## 3. Including dependencies -- added in commit da85c1d

At this point I needed to add two more classes, and it became a bit unclear where I should include them, since files were included in 4 different places. And also there was some duplication in includes/requires.

So I first moved all the includes/requires to the main plugin file. Including files is just basic plugin bootstrapping, and it would maybe be easiest to manage when it's all done in one place -- in the main plugin file. Which is exactly how Composer works -- the single `vendor/autoload.php` file handles all the libs and all the files from `vendor/`.

## 4. Controllers -- added in commit 6b44541

Created two Controllers:
1. `Newspack_Newsletters_ESP_API_Interface` == mandatory API functionality
2. `Newspack_Newsletters_WP_Hookable_Interface` == defines Mandatory integration with WP Hooks

A controller is just an API _interface_. Which is why it's recommended to keep this interface simple and thin, not to contain any business logic. Ideally it should just register endpoints, request params, do simple checks, and return responses. And for any actual heavy lifting, it should call the business logic class, which knows how to do specific tasks.

And that's why the controller got a service in its constructor, example [here](https://github.com/Automattic/newspack-newsletters/blob/refactor/provider-oop-ivan-demo/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp-controller.php#L27). It gets to call it for any logic. And that here is just the most basic form of the [Dependency Injection pattern](https://designpatternsphp.readthedocs.io/en/latest/Structural/DependencyInjection/README.html).

Next all the logic went to the Mailchimp class, but some of it should go to the ServiceProvider actually -- the general logic which is shared by all ESPs. I had trouble differentiating between the "general endpoints" and the "Mailchimp-only endpoints", so for now I put them all into the `Mailchimp Controller` forn, but some them should go into the SP one.

image

## 5. Other possible next steps

- the [`Newspack_Newsletters`](https://github.com/Automattic/newspack-newsletters/blob/refactor/provider-oop-ivan-demo/includes/class-newspack-newsletters.php) class could also get its API extracted into a controller.
- this if-else block [right here](https://github.com/Automattic/newspack-newsletters/blob/refactor/provider-oop-ivan-demo/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php#L316) would be ideal to single out to a separate function called `get_campaign_result()`. Why? Because if one saw `$this->get_campaign_result()` at that point in the code, one wouldn't have to spend any time thinking about what it does. Like this one needs to read to understand it first.
- naming -- I'd propose just a few naming changes, to make the functions immediately understandable, without being familiar with the Plugin. Examples: `api_key()` -> `get_api_key()`, `retrieve( $post_id )` -> `retrieve_newsletter( $post_id )`, `test( $post_id, $emails )` -> `send_test_newsletter( $post_id, $emails )`, ...
- I haven't tried this one -- but now that the number of classes is growing, we might use a better directory structure, just like you created the `includes/service-providers` folder, we could perhaps use a `controllers` folder too


iuravic:
@claudiulodro, commenting this part

> If someone wants to add a new service provider, do they need to open a new PR and get the service provider merged here? Or is the intention that they can make a new plugin that adds the service provider (similar to how WooCommerce payment gateways work)? Newspack Newsletters could have much wider reach if it supported the ability to make plugins that add providers. That would require some new filters or a registration function or something but should otherwise be very straightforward.

That's a great point! Agreed, if we'd like to, it should be quite doable to add an action/filter which registers additional Service Providers.

PR #248, opened 2020-06-30
Show spinner until layout previews are fully loaded

### LINKED ISSUE #200: Add placeholer/spinner state to Posts Inserter
Currently when we load the block we can sometimes see an image placeholder before being replaced with the actual featured image.

Instead I suggest we see a `Spinner` until the preview of the Posts Inserter is fully loaded -- kind of what we currently do with the Homepage Posts block.

Kapture 2020-05-07 at 13 11 09

dkoo:
### All Submissions:

When layout block previews are rendered, images may take time to load, resulting in a flash of unstyled content (FOUC) before they're loaded. Since the `BlockPreview` element doesn't currently support props for a `onReady` callback, we can use a `MutationObserver` to look for changes in the `BlockPreview` DOM tree, watch for featured images, and attach event listeners to the `load` event to tell when images in the preview are finished loading.

Unfortunately, this solution doesn't detect whether the featured image is actually visible in the preview, but that would add a lot more complexity.

If `BlockPreview` does end up supporting an `onReady` callback prop (one is apparently [in the works](https://github.com/WordPress/gutenberg/tree/master/packages/block-editor/src/components/block-preview#__experimentalonready)), we should be able to deprecate this home-grown solution in favor of the prop.

Closes #200.

1. On the master branch, create a new Newsletter.
2. In the Layout selector UI that appears, observe the FOUC as the featured image loads in the upper-left default layout.
3. Check out this branch. Refresh the new Newsletter page (or create a new Newsletter again).
4. This time, observe that the layout previews are covered with a spinner animation until all block preview elements are fully loaded.

Note: I tested in Chrome, Firefox, and Safari. [`MutationObserver`](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) has full support back to IE11. 

### Other information:

### DIFF
--- a/src/components/newsletter-preview/index.js
+++ b/src/components/newsletter-preview/index.js
@@ -2,7 +2,8 @@
  * WordPress dependencies
  */
 import { BlockPreview } from '@wordpress/block-editor';
-import { Fragment, useMemo } from '@wordpress/element';
+import { createRef, Fragment, useMemo, useState, useEffect } from '@wordpress/element';
+import { Spinner } from '@wordpress/components';
 
 /**
  * Internal dependencies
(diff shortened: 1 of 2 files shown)


dkoo:
Makes sense—I'll close this PR and add a spinner to the block instead.

PR #264, opened 2020-07-07
feat: show spinner while Post Inserter is loading image blocks

### LINKED ISSUE #200: Add placeholer/spinner state to Posts Inserter

dkoo:
### All Submissions:

In the Post Inserter block, images sometime take a while to get inserted into the `BlockPreview`. This adds a "loading" state with the WP spinner component while image blocks are still being inserted. Note that there may be a very brief delay after the blocks are shown while the image actually loads—this is unavoidable unless we attach event listeners to the actual DOM for image `load` events, which gets messy fast since the `BlockPreview` elements get rendered inside iframes and don't currently support any on ready callbacks. I tried testing this in #248, but after doing this refactor I believe we should avoid mixing virtual DOM and true DOM logic as a best practice.

Since the preview blocks are built from post data using array mapping functions instead of true asynchronous behavior, the logic goes like this:

* On mount or when the array of posts to be displayed changes, we show a loading spinner.
* Once we have an array of posts, we count all of the featured images for those posts.
* While the preview blocks are being built, we keep a running count of image blocks being inserted into the preview.
* Once the number of inserted preview image blocks matches the number of featured images, we consider the preview "ready" and remove the spinner.
* If the block is set to not show featured images, we can skip the loading logic, since text-based blocks render almost instantly.

Closes #200.

Replaces #248, which handled the loading state in the Layout preview dialogue but not in the actual Post Inserter block.

1. Create a new Newsletter.
2. In the Layout dialogue, confirm that any layouts that contain a Post Inserter will show a loading spinner until the featured images in the preview are finished loading.
3. Select a layout that contains a Post Inserter block. In the actual block editor, the Post Inserter's default settings show the most recent post with featured images enabled. If this post has a featured image, observe a loading spinner until the preview blocks are finished loading.
4. In the Post Inserter sidebar settings, enable **Display Specific Posts**. Confirm that the block displays as empty without a spinner.
5. Enter a post title in the auto-complete input and select at least one post. Confirm that the spinner reappears until the preview is finished loading.
6. In the Post Inserter sidebar settings, increase the number of items shown. Confirm that the spinner appears and remains until all post preview blocks are finished loading.

### Other information:

### DIFF
--- a/src/editor/blocks/posts-inserter/index.js
+++ b/src/editor/blocks/posts-inserter/index.js
@@ -45,7 +45,7 @@ const PostsInserterBlock = ( {
 	setInsertedPostsIds,
 	removeBlock,
 } ) => {
-	const [ ready, setReady ] = useState( ! attributes.displayFeaturedImage );
+	const [ isReady, setIsReady ] = useState( ! attributes.displayFeaturedImage );
 
 	// Stringify added to minimize flicker.
 	const templateBlocks = useMemo( () => getTemplateBlocks( postList, attributes ), [
@@ -58,16 +58,16 @@ const PostsInserterBlock = ( {
 
 		// No spinner if we're not dealing with images.
 		if ( ! attributes.displayFeaturedImage ) {
-			return setReady( true );
+			return setIsReady( true );
 		}
 
 		// No spinner if we're in the middle of selecting a specific post.
 		if ( isDisplayingSpecificPosts && 0 === specificPosts.length ) {
-			return setReady( true );
+			return setIsReady( true );
 		}
 
 		// Reset ready state.
-		setReady( false );
+		setIsReady( false );
 
 		// If we have a post to show, check for featured image blocks.
 		if ( 0 < postList.length ) {
@@ -77,7 +77,7 @@ const PostsInserterBlock = ( {
 
 			// If no posts have featured media, skip loading state.
 			if ( 0 === images.length ) {
-				return setReady( true );
+				return setIsReady( true );
 			}
 
 			// Wait for image blocks to be added to the BlockPreview.
@@ -85,7 +85,7 @@ const PostsInserterBlock = ( {
 
 			// Preview is ready once all image blocks are accounted for.
 			if ( imageBlocks.length === images.length ) {
-				setReady( true );
+				setIsReady( true );
 			}
 		}
 	}, [ JSON.stringify( postList ), JSON.stringify( templateBlocks ) ]);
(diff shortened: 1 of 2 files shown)


dkoo:
More testing needed on this, but wanted to get a draft started for my to-do list.

src/editor/blocks/posts-inserter/index.js
```diff
@@ -38,12 +45,51 @@ const PostsInserterBlock = ( {
...
+			const images = [];
+			postList.map( post => post.featured_media && images.push( post.featured_media ) );
+
+			// If no posts have featured media, skip loading state.
+			if ( 0 === images.length ) {
+				return setReady( true );
+			}
+
+			// Wait for image blocks to be added to the BlockPreview.
+			const imageBlocks = templateBlocks.filter( block => 'core/image' === block.name );
```
adekbadek, line 84: This will handle only image blocks which are top-level blocks. When testing with the "Daily/Weekly" template the feature did not work, because the image blocks there are nested in column blocks.

src/editor/blocks/posts-inserter/index.js
```diff
@@ -38,12 +45,51 @@ const PostsInserterBlock = ( {
 	setInsertedPostsIds,
 	removeBlock,
 } ) => {
+	const [ ready, setReady ] = useState( ! attributes.displayFeaturedImage );
```
adekbadek, line 48: Suggestion: There's a useful convention to prefix boolean variables with `is` (`isReady`, in this case)

src/editor/blocks/posts-inserter/index.js
dkoo: @adekbadek Thanks for pointing out—didn't realize this could contain images, too. Will revisit to see if I can capture those as well.

src/editor/blocks/posts-inserter/index.js
dkoo: Thanks for the tip—updated in 0bbe397.

src/editor/blocks/posts-inserter/index.js
dkoo: This should handle image blocks no matter how deeply nested—what do you think? d846beb

PR #265, opened 2020-07-09
Append utm_medium param to all links in newsletter

adekbadek:
### All Submissions:

For targeting visitors coming in from a Newsletter, all links in the newsletter will have a ~`utm_source`~ `utm_medium=email` parameter appended. ~The value of this parameter is `Newsletter` by default, and can be changed in the plugin settings.~

1. Create a newsletter with some links in the content. 
2. Send the campaign, observe that the links (after MC redirection) have a ~`utm_source=Newsletter`~ `utm_medium=email` parameter
3. ~Change the value of the parameter in plugin settings~
4. ~Send another newsletter with links – observe the parameter is updated~

### Other information:

### DIFF
--- a/includes/class-newspack-newsletters-renderer.php
+++ b/includes/class-newspack-newsletters-renderer.php
@@ -172,8 +172,8 @@ final class Newspack_Newsletters_Renderer {
 	 */
 	private static function process_links( $html ) {
 		preg_match_all( '/href="([^"]*)"/', $html, $matches );
-		$href_params       = $matches[0];
-		$urls              = $matches[1];
+		$href_params = $matches[0];
+		$urls        = $matches[1];
 		foreach ( $urls as $index => $url ) {
 			$url_with_params = add_query_arg(
 				[
--- a/includes/class-newspack-newsletters-renderer.php
+++ b/includes/class-newspack-newsletters-renderer.php
@@ -165,6 +165,30 @@ final class Newspack_Newsletters_Renderer {
 	}
 
 	/**
+	 * Append UTM param to links.
+	 *
+	 * @param string $html input HTML.
+	 * @return string HTML with processed links.
+	 */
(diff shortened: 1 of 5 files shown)


philipjohn:
Note that Mailchimp's own Google Analytics integration already does this, and adds more extensive (`utm_campaign`, `utm_medium`, `utm_content`) parameters including an ID for each campaign.

It's probably worth checking if this would override MC's own tracking first of all, but I would also recommend including at least `utm_campaign` and `utm_medium` parameters on top of `utm_source` if we're going to add this in the plugin.

jeffersonrabb:
> Note that Mailchimp's own Google Analytics integration already does this, and adds more extensive (`utm_campaign`, `utm_medium`, `utm_content`) parameters including an ID for each campaign.

I just verified this. I turned on the Google Analytics integration on in my Mailchimp account, sent a campaign, and clicked through the links. They look like this ("Test List 1" is the name of the list I sent to). 

`https://newspack1.ngrok.io/2020/03/27/dapibus-vehicula-malesuada-magnis-posuere-tincidunt-ante/?utm_source=Test+List+1&utm_campaign=da35ca99a2-EMAIL_CAMPAIGN_2020_07_09_06_23&utm_medium=email&utm_term=0_33012c60a9-da35ca99a2-364781390`

Perhaps we should suppress all newsletter sign-up campaigns for any user who has come to the site from a URL with a `utm_source` query param?

And @adekbadek, sorry you had to do this work! I should have researched first.

adekbadek:
I think we should use the `utm_medium` param (with `email` value) instead of `utm_source`. This way the `utm_medium` param will be appended both for users with GA in MC configured and the rest. This will allow us to treat traffic incoming from email accordingly, regardless of whether GA was integrated with MC or not.

adekbadek:
Update: change to parameter to `utm_medium=email`

PR #272, opened 2020-07-20
feat: enable public newsletter posts

### LINKED ISSUE #171: Convert Newsletter To Post
A Newspack customer recently asked whether it was possible to convert a Newsletter to a normal Post, so that Newsletter content could be fully integrated into the site. This could be useful for a site section such as a Newsletter Archive. This seems like it should be relatively easy since the Newsletter is already a post, but how would we approach the UI for this?

cc: @thomasguillot

### LINKED ISSUE #214: Archive page for newsletters
A potential subscriber might want to see what they're getting into. Since all Newsletters are WP Posts, it's trivial to add an archive page listing all newsletters.

image

dkoo:
### All Submissions:

Allows editors to mark sent/published newsletters as publicly viewable and creates an archive page for newsletter posts. If a published newsletter is publicly viewable, its single page is visible to all users. If not, it's visible as a preview only to logged-in WP admin users (drafts are also visible to logged-in admin users). Also hijacks the main query on the Newsletter archive pages and filters out any posts that are NOT marked as publicly viewable.

#### Further work required?

* ~[Posts Inserter block isn't rendering anything on the front-end](https://github.com/Automattic/newspack-newsletters/pull/272#issuecomment-673016984)—need to fix that before we merge~ fixed in 91b596c

* Currently, newsletter ads are not rendered in the page view. I'm not sure this is necessary, but feel free to change my mind.

* There might be some further design/front-end work needed on the theme side if we want the archive and single templates for newsletter posts to differ from regular articles (there is probably no need for a byline, or the sidebar widgets, for example). (cc @thomasguillot and/or @laurelfulford)

* Maybe disable some interactions with other plugins—e.g. Jetpack Related Posts (this example done in this PR)

Closes #171.
Closes #214.

1. Check out/build this branch, and flush permalinks. Go to **Newsletters > All Newsletters** in WP admin.

2. For any newsletters that have been sent, observe a new "Preview" link when hovering over the item in the post list:

Screen Shot 2020-08-12 at 10 05 55 AM

3. Click the Preview link. Observe that the newsletter is viewable as a single post page while logged into WP admin.

Screen Shot 2020-08-12 at 10 54 14 AM

4. Open a new Incognito window and attempt to visit the same newsletter post. Observe that the page is a 404 while not logged in.

Screen Shot 2020-08-12 at 10 57 58 AM

5. Go to https://[YOUR SITE URL]/newsletter/ to view the newsletter archive page. Right now it should be empty while both logged in and not logged in (only posts explicitly marked as publicly viewable should appear on the archive page).

6. Edit the newsletter post. Observe a new toggle option in the Document settings sidebar, disabled by default:

Screen Shot 2020-08-12 at 10 07 43 AM

7. Enable the setting. Observe that the previously disabled "Sent" button is now enabled and says "Update." Click "Update" to save the new setting.

8. Go back to the "All Newsletters" post list. Hover over the post and observe that there's now a "View" link instead of "Preview." Also observe that the "Sent" notice should now have an additional "Published as a page" note attached.

Screen Shot 2020-08-12 at 11 13 08 AM

9. Refresh the archive page—now observe that your public post is listed while both logged in and not logged in.

10. Visit the newsletter post's single page. Observe that the public post is now visible while both logged in and not logged in.

11. Set up at least one campaign in the Newspack Campaigns plugin (can be either overlay or inline). Observe that the campaign will never render in the newsletter post's single page under any circumstances.

12. Edit the newsletter post again and disable the "Make newsletter page public?" setting, then save. Observe that the page is no longer visible in the archive or as a single page to non-logged-in users.

13. As a sanity check, create a new newsletter but DON'T send it. DO enable the "Make newsletter page public?" setting.
  - Observe that the "Send" button says "Send and Publish" when the setting is enabled.
  - Click "Save Draft," then confirm that the draft is visible to logged-in users only.

14. As another sanity check, ensure that sending newsletters with both Mailchimp and Constant Contact still works as before.

### Other information:


dkoo:
@adekbadek I just realized that the Post Inserter block isn't rendering anything on the front-end. Before I go down the rabbit hole of trying to make it work, is there something off the top of your head we can change to fix this?

adekbadek:
> I just realized that the Post Inserter block isn't rendering anything on the front-end. Before I go down the rabbit hole of trying to make it work, is there something off the top of your head we can change to fix this?

@dkoo, you can render it [as the email renderer does](https://github.com/Automattic/newspack-newsletters/blob/4c1e33acf21fd8bd345137995894f0cc12e0468c/includes/class-newspack-newsletters-renderer.php#L526). The block saves resulting markup in a `innerBlocksToInsert` attribute. Other solution would be to duplicate the block rendering logic in PHP, as most blocks out there do.

dkoo:
> @dkoo, it can be rendered [as in the email renderer](https://github.com/Automattic/newspack-newsletters/blob/4c1e33acf21fd8bd345137995894f0cc12e0468c/includes/class-newspack-newsletters-renderer.php#L526). The block saves resulting markup in a `innerBlocksToInsert` attribute. Other solution would be to duplicate the block rendering logic in PHP, as most blocks out there do.

Thanks for the hint! I ended up using that attribute to just render the inner blocks via server-side callback. Added in 91b596c.

includes/class-newspack-newsletters.php
```diff
@@ -281,11 +333,132 @@ public static function display_post_states( $post_states, $post ) {
...
+	public static function disable_campaigns_for_newsletters( $disabled ) {
+		if ( self::NEWSPACK_NEWSLETTERS_CPT === get_post_type() ) {
+			return true;
+		}
+
+		return $disabled;
+	}
+
+	/**
+	 * Disable Jetpack Related Posts on Newsletter posts.
```
adekbadek, line 449: Maybe I'm missing something here, but why would recirculation here be unwanted? Or is that disabled because Jetpack will not handle content of Newsletters posts?

includes/service-providers/constant_contact/class-newspack-newsletters-constant-contact.php
```diff
@@ -374,45 +374,50 @@ public function save( $post_id, $post, $update ) {
...
 		if ( ! Newspack_Newsletters::validate_newsletter_id( $post_id ) ) {
 			return new WP_Error(
 				'newspack_newsletters_incorrect_post_type',
 				__( 'Post is not a Newsletter.', 'newspack-newsletters' )
 			);
 		}
 
-		try {
-			$sync_result = $this->sync( $post );
+		if ( 'publish' === $new_status && 'publish' !== $old_status ) {
```
adekbadek, line 391: Isn't that essentially replicating the `publish_` action?

includes/class-newspack-newsletters.php
```diff
@@ -281,11 +333,132 @@ public static function display_post_states( $post_states, $post ) {
 				/* translators:  Absolute time stamp of sent/published date */
 				$post_states[ $post_status->name ] = sprintf( __( 'Sent %1$s', 'newspack-newsletters' ), get_the_time( get_option( 'date_format' ), $post ) );
 			}
+
+			if ( $is_public ) {
+				$post_states[ $post_status->name ] .= __( ' | Published as a page', 'newspack-newsletters' );
```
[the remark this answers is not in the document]
dkoo: Good suggestion—changed to "Published as a post" in 6815be4

includes/class-newspack-newsletters.php
```diff
@@ -281,11 +333,132 @@ public static function display_post_states( $post_states, $post ) {
...
+
+	/**
+	 * Disable auto-saves after newsletter has been sent.
+	 */
+	public static function maybe_disable_autosave() {
+		if ( 'publish' !== get_post_status() || self::NEWSPACK_NEWSLETTERS_CPT !== get_post_type() ) {
+			return false;
+		}
+
+		wp_dequeue_script( 'autosave' );
```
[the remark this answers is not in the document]
dkoo: Didn't know we could do this! Fixed in 6815be4.

includes/class-newspack-newsletters.php
dkoo: This was an assumption on my part. When it's enabled, it only shows other newsletters, so I felt it was less useful than recirc on regular posts. Do you think publishers would want Related Posts to appear on newsletters if they only show other newsletters?

includes/service-providers/constant_contact/class-newspack-newsletters-constant-contact.php
dkoo: It is, but the `publish` action gets triggered anytime a published post is updated, too, so I was running into errors when updating meta for newsletters that I had already sent (the action would try to re-send the newsletters every time I updated the post). So using `transition_post_status` and only triggering the send if the post is transitioning from unpublished to published fixed that and also allows the post to be updated after being sent.

dkoo:
> The post publish button is set to "Update" after being briefly set to "Send". On master, this button is set to "Send" and stays that way. It's content does not change after marking the Newsletter as public in the sidebar.

Addressed this in c15f2e6. The problem was that I was relying on WP core's `isEditedPostDirty` selector, which supposedly checks whether the editor is in a dirty state. But because we're setting some meta based on async ops from the service providers, `isEditedPostDirty` always returns `true`. Now, I only check whether the `is_public` meta value has changed from the post's saved state before changing the button state.

This means that if you edit other post attributes or post content, the "Update" button won't be enabled, but it's still possible to make those edits, change the "public" setting, and then save all changes. Kind of an edge case since I don't see much of a use case for editing newsletter content after it has been sent, but I could also add a comparison of post content vs. edited post content if you think that's warranted.

includes/class-newspack-newsletters.php
adekbadek: I don't have a strong opinion here, I think it wouldn't hurt. Plus, a use case for this public newsletters posts feature is to allow a potential subscriber to assess what kind of content appears in the newsletters – a Related Posts section would be a nice way of navigating to other newsletter posts to get a better overview.

includes/service-providers/constant_contact/class-newspack-newsletters-constant-contact.php
adekbadek: Got it, thanks for the clarification!

adekbadek:
> This means that if you edit other post attributes or post content [of a sent newsletter], the "Update" button won't be enabled […] Kind of an edge case

I agree it's an edge case, I'm fine with it staying that way

dkoo:
> I noticed a bug when I change the public status of a newsletter where the newsletter single view is still available even when setting has been changed to don't publish.

@thomasguillot, are you logged in as an admin user? If that's the case then newsletter posts will always be visible as a preview. Try viewing the same post in a new incognito window and see if the page still renders?

thomasguillot:
> > I noticed a bug when I change the public status of a newsletter where the newsletter single view is still available even when setting has been changed to don't publish.
> 
> @thomasguillot, are you logged in as an admin user? If that's the case then newsletter posts will always be visible as a preview. Try viewing the same post in a new incognito window and see if the page still renders?

Oh ok yeah it work as expected then. But why will it always be visible as a preview if I specifically un-toggled the option?

includes/class-newspack-newsletters.php
dkoo: I think you've convinced me that we shouldn't disable by default. I added an option in cbaa4d3 that lets editors decide one way or another—what do you think?

dkoo:
> But why will it always be visible as a preview if I specifically un-toggled the option?

It's the only way I could think of that would allow editors to preview a single newsletter post without making it public. It more or less works the same way as a draft post, but since the newsletter is already "published" (a.k.a. sent) regardless of whether the post is marked as public, we can't use the WP publish status in the same way as we would use it in a regular post.

PR #293, opened 2020-08-13
Remove release changes from master

adekbadek:
In order to make alpha releases easier, let's remove the release-generated stuff from the master. It was added in #197, but that thought to be needed because of GH misconfiguration (out-of-date warning on PRs).


adekbadek:
Since we can re-set `alpha` on top of `release` after a release, there's no requirement to get rid of this code.

PR #317, opened 2020-08-28
Assorted bugfixes

### LINKED ISSUE #300: Styles: default font doesn't match render
When you create a new newsletter, the body font rendered looks like Times/serif even though the default font is Arial. Does not seem to affect heading font.

To replicate:

1. Create a new newsletter and select any layout.
2. Observe that the body font looks like Times/serif:

Screen Shot 2020-08-18 at 10 04 39 PM

3. Send a test email (or send the campaign). Observe that the body font is still Times/serif.
4. Change the body font setting to update the font. The font should change as expected.
5. Change the body font back to the default Arial. It should now render as Arial in both the editor and a sent email.

### LINKED ISSUE #305: Public newsletters: use styles set in block editor
Currently, styles set in the block editor (background color, fonts, etc.) are not rendered on the front-end when viewing the newsletter as a public post.

adekbadek:
### All Submissions:

Closes #300
Closes #305

1. On `master`
2. Create two ads, one published and one draft
3. Create a newsletter, set custom fonts and background color in the Styling panel
4. Send a test email, observe that the draft ad is included in the email, the custom fonts are not applied, and that when the newsletter is viewed as a WP page, the custom styles are not applied
5. Switch to this branch, re-send the email
6. Observe the draft ad is not included, the custom styles are applied both in the email and in the WP page
7. Create a new newsletter, using a template with some headers. 
8. Send a test email, observe the headers have the default Arial font, while the body has the default Georgia font

### Other information:


includes/class-newspack-newsletters.php
```diff
@@ -68,13 +68,15 @@ public function __construct() {
...
+		add_action( 'wp_head', [ __CLASS__, 'public_newsletter_custom_style' ], 10, 2 );
 		add_filter( 'display_post_states', [ __CLASS__, 'display_post_states' ], 10, 2 );
 		add_action( 'pre_get_posts', [ __CLASS__, 'maybe_display_public_archive_posts' ] );
 		add_action( 'template_redirect', [ __CLASS__, 'maybe_display_public_post' ] );
 		add_filter( 'post_row_actions', [ __CLASS__, 'display_view_or_preview_link_in_admin' ] );
 		add_filter( 'newspack_newsletters_assess_has_disabled_popups', [ __CLASS__, 'disable_campaigns_for_newsletters' ], 11 );
 		add_filter( 'jetpack_relatedposts_filter_options', [ __CLASS__, 'disable_jetpack_related_posts' ] );
-		add_action( 'save_post_' . self::NEWSPACK_NEWSLETTERS_CPT, [ $this, 'save' ], 10, 3 );
+		add_action( 'save_post_' . self::NEWSPACK_NEWSLETTERS_CPT, [ __CLASS__, 'save' ], 10, 3 );
+		add_filter( 'posts_results', [ __CLASS__, 'filter_search' ] );
```
jeffersonrabb, line 79: Using this filter affects more than just search results and will have unexpected results. With this filtering in place, the Newsletters list page shows only Newsletters with `is_public` switched on.

includes/class-newspack-newsletters.php
```diff
@@ -780,6 +799,34 @@ public static function default_title( $post_title, $post ) {
 		return $post_title;
 	}
 
+	/**
+	 * Handle custom Newsletter styling when viewing the newsletter as a public post.
+	 */
+	public static function public_newsletter_custom_style() {
```
jeffersonrabb, line 805: These styles will be included in the Newsletter archive page. This doesn't seem to have any obvious effect but it could cause unexpected results. Suggested fix:

```
if ( ! is_single() ) {
	return;
}
```

includes/class-newspack-newsletters.php
```diff
@@ -86,6 +88,23 @@ public function __construct() {
 		}
 	}
 
+	/**
+	 * Filter search results, disallowing non-public newsletter posts.
+	 *
+	 * @param array $posts The found posts.
+	 */
+	public static function filter_search( $posts ) {
```
dkoo, line 96:
```suggestion
	public static function filter_search( $posts ) {
		if ( is_admin() ) {
			return $posts;
		}
```
Would this do the trick?

includes/class-newspack-newsletters.php
```diff
@@ -780,6 +799,34 @@ public static function default_title( $post_title, $post ) {
...
+	public static function public_newsletter_custom_style() {
+		$post = get_post();
+		if ( $post && self::NEWSPACK_NEWSLETTERS_CPT === $post->post_type ) {
+			$font_header      = get_post_meta( $post->ID, 'font_header', true );
+			$font_body        = get_post_meta( $post->ID, 'font_body', true );
+			$background_color = get_post_meta( $post->ID, 'background_color', true );
+			?>
+				<style>
+					.main-content {
+						background-color: <?php echo esc_attr( $background_color ); ?>;
```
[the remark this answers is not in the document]
thomasguillot: Yes we should.

includes/class-newspack-newsletters.php
adekbadek: Fixed in 9578e710708c9734db4cd8a7920180c2e899292f

includes/class-newspack-newsletters-renderer.php
```diff
@@ -652,6 +642,7 @@ function ( $block ) {
 				array(
 					'post_type'      => Newspack_Newsletters_Ads::NEWSPACK_NEWSLETTERS_ADS_CPT,
 					'posts_per_page' => -1,
+					'post_status'    => 'publish',
```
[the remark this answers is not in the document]
adekbadek: That's true – reverted in 29de87551b251b5e42a7e33b61c3109706cba0cd. I don't know what caused a draft ad to appear in a newsletter for me, but cannot reproduce now either.

includes/class-newspack-newsletters.php
adekbadek: Should be fixed in 12570a68b904b02d0138bc5e78756dd0d10bad28

includes/class-newspack-newsletters.php
adekbadek: I've added that and some more checks 12570a68b904b02d0138bc5e78756dd0d10bad28in , thanks!

includes/class-newspack-newsletters.php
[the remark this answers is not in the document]
adekbadek: Added in b59ce0e07c7d6db09aa2303e89369993da381c88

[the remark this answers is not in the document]
thomasguillot: Since Newspack theme uses `32px` for its margins, we should use the same here. Also probably want to avoid padding top and bottom since the blocks add a margin top and bottom already. So I suggest `padding: 0 32px;`

adekbadek:
@jeffersonrabb This is an issue with filtering search results – ideally meta query should be used here, but I did not have any success with that approach. Reverting these changes in c3e051a – to be addressed in another PR.

includes/class-newspack-newsletters.php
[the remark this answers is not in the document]
adekbadek: Updated in 1d136bed133d3e3aa25f1c2143195ed30f30af55

thomasguillot (approved):
Looks good on my end 👍

PR #320, opened 2020-09-01
Add plugin readme.txt

laurelfulford:
### All Submissions:

This PR adds a readme.txt file to the plugin, following the format from these sources:

https://developer.wordpress.org/plugins/wordpress-org/how-your-readme-txt-works/
https://developer.wordpress.org/plugins/wordpress-org/how-your-readme-txt-works/

Right now, the 'Changelog' part is copy-pasted from the release changelogs, down to v1.0 (which I replaced with 'Initial release', to save having to copy over all of the items). In the future it could be handy to automate this somehow, but this should work for the short-term. 

Note that the different sections ("Description", "Installation", "Installation") populate the different tabs in the plugin repo, like on the Jetpack page: https://wordpress.org/plugins/jetpack/

1. Read through the readme and check for any errors. 
2. Check for any formatting issues in the markdown eg. make sure the links are formatted correctly. 
3. Confirm if the required PHP, WordPress versions are correct. 

Open-ended questions: is anything missing? 

### Other information:


readme.txt
```diff
@@ -0,0 +1,237 @@
...
+= What newsletter providers are supported? =
+
+Mailchimp and Constant Contact are currently supported.
+
+== Screenshots ==
+
+1. Pick from one of the bundled templates to start your new newsletter, or start with a blank page. You can also create and save your own templates to create a consistent newsletter series.
+2. In the editor, you can add and remove blocks, save settings, and send yourself a test email before sending out your newsletter.
+
+== Changelog ==
```
adekbadek, line 67: Is that required? If we add a changelog here, it will have to be updated on release.

adekbadek (commented):
Just one question

readme.txt
```diff
@@ -0,0 +1,237 @@
...
+
+= Newspack Blocks =
+
+Newspack Newsletters comes bundled with the Newspack Post Inserter block. This block lets you insert excerpts from live posts on your site, and, once inserted, will turn those posts into static content using the Header and Paragraph blocks, linking back to your site.
+
+Need to target a post title to your newsletter audience? Not a problem! You can edit these blocks, and add others, to make sure they’ speak to your readers.
+
+= Add ads! =
+
+Monetize your newsletters by including advertising. The Newspack Newsletters plugin lets you create ads, which are stored as a custom post type. Ads are automatically appended to the end of your newsletters, and can also be toggled off on a newsletter-by-newsletter basis.
```
adekbadek, line 36: Ads can be set to a percentage position ;)

```suggestion
Monetize your newsletters by including advertising. The Newspack Newsletters plugin lets you create ads, which are stored as a custom post type. Ads are automatically inserted into your newsletters, and can also be toggled off on a newsletter-by-newsletter basis.
```

adekbadek (changes requested):

readme.txt
laurelfulford: I _assumed_ it was, but I don't think it actually is either! 

When running the readme through the [validator](https://wordpress.org/plugins/developers/readme-validator/) it will show a Note (rather than a Warning) when the Changelog is missing -- we also get a note for not including a Donate link. Warnings seem to be limited to the vital things like missing the 'Contributors' or 'Tested up to'. 

It'll definitely be nice to not have to manually maintain it here -- I've removed it in dd8cd31.

readme.txt
laurelfulford: 😮 I did not know this! That's awesome! 

Updated as suggested in cb7af53

adekbadek (approved):

laurelfulford:
Thanks Adam!

PR #342, opened 2020-09-22
Alpha release Sep 22

adekbadek:
A new alpha release.


adekbadek:
This will not trigger an new alpha release (just a single `test:` commit is here), but we'll see the test-and-release workflow during the actual release.

PR #350, opened 2020-09-30
fix: prevent draft ads from being inserted into newsletters

adekbadek:
### All Submissions:

We were thinking that WP query would return only published posts by default, but it's different in admin context:

> And if the query is run in an admin context (administration area or AJAX call), protected statuses are added too. By default protected statuses are ‘future‘, ‘draft‘ and ‘pending‘.

([source](https://developer.wordpress.org/reference/classes/wp_query/#status-parameters))

There's also a tiny tweak to PHPCS configuration here.

1. Create a draft ad
2. On `master`, observe it's inserted into the newsletter
3. Switch to this branch, observe the ad is not inserted into the newsletter

### Other information:


dkoo (approved):
I actually wasn't seeing draft ads getting inserted into sent Newsletters on `master`, but this extra check certainly can't hurt.

PR #352, opened 2020-09-30
Chore: remove author URI

adekbadek:
Because of this error while submitting to WPORG:

> Error: Your plugin and author URIs are the same. Your plugin headers in the main plugin file headers have the same value for both the plugin and author URI (Uniform Resource Identifier). A plugin URI is a webpage that provides details about this specific plugin. An author URI is a webpage that provides information about the author of the plugin. Those two must be different. You are not required to provide both, so pick the one that best applies to your situation.


claudiulodro (approved):
After the plugin goes live, we can change the plugin URI to the wporg page for the plugin, but this should work well for now.

PR #428, opened 2021-03-23
fix: respect excerpt length setting while not messing up site excerpts

### LINKED ISSUE #424: Posts Inserter: excerpt length slider length can't exceed a certain number
For the Posts Inserter, there's an option to set the "max number of words in excerpt." This setting defaults to the site's excerpt length (usually 55, sometimes lower or higher if reset by another plugin). However, the slider currently doesn't allow you to exceed this default—if you increase the setting beyond that number, the excerpts in the block will remain at the default length. Reducing the setting below the default length will correctly truncate the excerpts, though.

To replicate:

1. Publish a post with at least 100 words in the content. (Note that if the generated or custom excerpt only has a certain number of words, the setting will not expand the excerpt beyond what exists.) Do not add a custom excerpt.
2. Create or edit a newsletter post and add a Posts Inserter block. Try to increase the "max number of words in excerpt" to 100 and observe that it doesn't have any effect on the excerpt.
3. Add a long custom excerpt to the content (100+ words). Observe the same behavior in the Posts Inserter block.

Expected behavior:

Whatever number is shown by the "max number of words in excerpt" setting should be the max number of words in each post's excerpt, even if that exceeds the filtered excerpt length for the site.

dkoo:
### All Submissions:

Filters the excerpt length server-side so that the length set in Posts Inserter block attributes can override the site's excerpt length.

Because Newsletters uses standard REST API endpoints and not custom endpoints, we need to hook into `rest_post_query` and `the_posts` hooks to know when to set and reset the excerpt length, respectively.

Should behave similarly to https://github.com/Automattic/newspack-blocks/pull/701.

Closes #424.

1. Create or edit a newsletter, add a Posts Inserter block.
2. Turn on the "Display post excerpt" option and set an excerpt length to something much longer than the default (55), such as `100`.
3. On `master`, observe that in the block preview and a sent test email, the excerpts shown in the block don't exceed 55 words even though the block attribute is set to 100.
4. Check out this branch and run `npm run build`.
5. Refresh the newsletter and confirm that now, the excerpt lengths match the block attribute setting, even if it exceeds 55. Send the test letter and confirm the same is true in the sent email.
6. View some pages showing post excerpts, such as an archive page and a page with Homepage Posts blocks. Verify that the Posts Inserter excerpt length setting doesn't affect the excerpts in these other contexts, because the filter is removed as soon as the newsletter posts query returns posts.
7. Retest 4-6 with a very short excerpt length setting (such as `10`). Confirm again that the excerpts rendered in all contexts (newsletter editor, newsletter email, Homepage Posts block, archive pages) have the expected lengths.
8. (Optional) Follow testing instructions for https://github.com/Automattic/newspack-blocks/pull/701 and https://github.com/Automattic/newspack-blocks/pull/694 and confirm that excerpts still behave as expected in all contexts.

### Other information:

### DIFF
--- a/includes/class-newspack-newsletters-editor.php
+++ b/includes/class-newspack-newsletters-editor.php
@@ -21,6 +21,13 @@ final class Newspack_Newsletters_Editor {
 	protected static $instance = null;
 
 	/**
+	 * Closure for excerpt filtering that can be added and removed.
+	 *
+	 * @var newspack_newsletters_excerpt_length_filter
+	 */
+	public static $newspack_newsletters_excerpt_length_filter = null;
+
+	/**
 	 * Main Newspack Newsletter Editor Instance.
 	 * Ensures only one instance of Newspack Editor Instance is loaded or can be loaded.
 	 *
@@ -40,6 +47,8 @@ final class Newspack_Newsletters_Editor {
 		add_action( 'the_post', [ __CLASS__, 'strip_editor_modifications' ] );
 		add_action( 'enqueue_block_editor_assets', [ __CLASS__, 'enqueue_block_editor_assets' ] );
 		add_filter( 'allowed_block_types', [ __CLASS__, 'newsletters_allowed_block_types' ], 10, 2 );
+		add_action( 'rest_post_query', [ __CLASS__, 'maybe_filter_excerpt_length' ], 10, 2 );
+		add_filter( 'the_posts', [ __CLASS__, 'maybe_reset_excerpt_length' ] );
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


adekbadek (changes requested):
It works as expected, however I see the following error logged when the posts are fetched:

```
PHP Warning:  Invalid argument supplied for foreach() in […] wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php on line 987
```

Is the excerpt length capped at 55 by WP? If so, then WDYT about, instead of adding complexity, capping it at 55 in the block settings?

dkoo:
> Is the excerpt length capped at 55 by WP? If so, then WDYT about, instead of adding complexity, capping it at 55 in the block settings?

55 is just the default excerpt length in WP. This change came about because a publisher was trying to use longer excerpts in their newsletters but found they couldn't exceed the default—which in their case was actually 30 words, probably because it was being filtered elsewhere as well (a prime suspect would be the WooCommerce Memberships plugin, which filters the excerpt length. Blocks handles this, and I've replicated the handling here). So the problem might still occur even if we cap the setting at 55 without doing our own excerpt length filtering. For that reason I think it might be worth the extra complexity.

I've fixed the PHP warning with 8a5f5b8—can you take another look?

adekbadek (approved):

PR #454, opened 2021-04-28
feat: custom CSS

### LINKED ISSUE #417: Allow custom styles in email-template.mjml.php
Hi all!

I've been using MJML templates for our newsroom's emails for a year or so and have a series of templates that are rock-solid and are a big part of our newsroom branding. We just got access to this plugin via Lede, but right now there is no way to import our styles into the mjml template. We're currently not using the plugin at all for this reason and I would really like to bring it into our workflow.

An ideal execution could be as simple as as an optional field (with plenty of warnings about it being an advanced feature) under the styling menu that would allow a user to overwrite everything in the `<mj-style>` tag in email-template.mjml.php.

That feature would basically instantly take this plugin from unusable to invaluable in our newsroom.

dkoo:
### All Submissions:

Adds a new CodeMirror field to the Styling panel which lets you add custom CSS to the newsletter. The CSS entered here will be appended to the default MJML styles in sent emails (test or live) only (appended, this way all emails will still have certain base styling, but these can be extended or overwritten since the custom CSS cascades after the defaults).

Custom CSS has no effect on how the newsletter appears on the site as a public newsletter because this can be handled using the Customizer's Additional CSS field.

Screen Shot 2021-04-27 at 5 06 24 PM

Some quirks and questions:

* Is the behavior to append custom CSS the right call? This could bloat the email payload if a publisher wants to overwrite a lot of default styles, but it also helps prevent emails from looking totally broken if the publisher doesn't provide comprehensive styles.
* Relies on the [React-CodeMirror](https://uiwjs.github.io/react-codemirror/) library—any objections? I first tried using a.) WP's built-in CodeMirror, and b.) the standalone CodeMirror NPM package, but neither plays well with webpack and React. I also thought about providing just a plain `textarea` element, but the code editing experience from CodeMirror feels so much nicer.
* The custom CSS is shown in the preview by adding it to the document's `style` element. ~This works well, but it also affects elements outside the editor preview, which can have some funny effects. I can't think of another way to do this and only target elements within the editor preview element, though, can you?~ Fixed in 9fcdfe9, h/t @adekbadek 
* I use `esc_html` to escape the CSS—is this enough? We can also use `sanitize_text_field` as a callback on the meta field, but this destroys the pretty formatting that CodeMirror provides when entering CSS in the textarea.
* I haven't done extensive testing on different CSS properties and how they might render in the sent emails. Such an open feature might inevitably result in strange output, so we should encourage users to send test emails whenever possible to ensure the custom CSS renders as expected.

Happy to revisit the CodeMirror approach if you have suggestions or other feedback.

Closes #417.

1. Check out this branch, run `npm ci && npm run build` to rebuild with new dependencies.
2. Create or edit a newsletter.
3. Expand the "Styling" sidebar panel. Confirm there's a new "Custom CSS" code editor field.
4. Enter some text in the field. Confirm that the field provides inline linting to highlight invalid CSS (although it's still possible to save the post with invalid CSS—this is consistent with other WP "custom CSS" features).
5. Enter some valid CSS. Confirm that the custom styles are automatically applied to the editor preview.
6. Send the email as a test or live email. Confirm that the sent email is styled with your custom CSS.
7. Bonus: preview the newsletter post or view it as a public post after publishing. Confirm that the custom CSS is NOT rendered on the site.

### Other information:


adekbadek (commented):
1. 

> Is the behavior to append custom CSS the right call? 

I think so. Making the CSSes mutually exclusive would require the user to recreate base styles.

2.

> Relies on the React-CodeMirror library—any objections?

I was worried about the size of the released zip, but the difference is just ~2kB after zipping

3.

> The custom CSS is shown in the preview by adding it to the document's style element. This works well, but it also affects elements outside the editor preview

Yeah, this is an issue. There [_was_](https://stackoverflow.com/a/45692033) a `scoped` attribute proposal, but it was dropped. A stylesheet is basically a list of rules, and can be [parsed nativelly](https://stackoverflow.com/a/14865690/3772847). We could rebuild the stylesheet by prefixing each rule with editor selector to effectively scope the stylesheet.

4.

> I use esc_html to escape the CSS—is this enough? 

I think it's ok.


5. 

> I haven't done extensive testing on different CSS properties and how they might render in the sent emails. Such an open feature might inevitably result in strange output

We could just add a disclaimer saying that this is an advanced feature and CSS in email tends to misbehave.

dkoo:
> A stylesheet is basically a list of rules, and can be parsed nativelly. We could rebuild the stylesheet by prefixing each rule with editor selector to effectively scope the stylesheet.

TIL! Implemented the auto-scoping in 9fcdfe9.

> We could just add a disclaimer saying that this is an advanced feature and CSS in email tends to misbehave.

Also updated the Custom CSS help text in 9fcdfe9.

src/newsletter-editor/styling/index.js
```diff
@@ -178,7 +204,7 @@ export const Styling = compose( [
 				id={ `inspector-custom-css-control-${ instanceId }` }
 				label={ __( 'Custom CSS', 'newspack-newsletters' ) }
 				help={ __(
-					'Custom CSS will be appended to default styles in sent emails only. Refresh editor to preview styles.',
+					'This is an advanced feature and may result in unpredictable behavior. Custom CSS will be appended to default styles in sent emails only.',
```
adekbadek, line 207: This means that the styles will not be applied to the public newsletters viewed as posts? That would actually be a nice feature, WDYT?

adekbadek (approved):
Works as advertised 🙌

src/newsletter-editor/styling/index.js
dkoo: Correct—it was my hunch that the CSS entered for emails may not translate well to the site. In my experience, email CSS requires a lot of `!important` and aggressive style overrides, which could be too heavy-handed for the site, so I felt that if publishers wanted to apply their custom styles to public newsletters viewed as posts, they could do that via the Customizer and have a bit more control. I'm willing to try, though, maybe in a new PR?

adekbadek: > if publishers wanted to apply their custom styles to public newsletters viewed as posts, they could do that via the Customizer and have a bit more control. 

That makes sense, yeah.

PR #477, opened 2021-06-16
Update various packages, including colors to use latest WP base-styles 

thomasguillot:
### All Submissions:

The main change with this PR is the use of the latest `@wordpress/base-styles` package which means that the color variables needed to be updated.
I've also:
* Removed Material packages
* Updated calypso-build to 6.3.0 to match other plugins
* Small bug fixes

1. Switch to this branch
2. Check the various screens

### Other information:


dkoo (changes requested):
Changes look good! This isn't strictly related, but it's something that's always bothered me and it probably involves some of these same files. The social icons container has a left margin that makes it look off-center in the editor:

Screen Shot 2021-06-16 at 4 10 18 PM

Can we fix? 😬

dkoo (approved):
👍

dkoo (changes requested):
Whoops, accidentally approved this PR (wrong window!)

thomasguillot:
> Changes look good! This isn't strictly related, but it's something that's always bothered me and it probably involves some of these same files. The social icons container has a left margin that makes it look off-center in the editor:
> 
> Can we fix? 😬

Fixed in https://github.com/Automattic/newspack-newsletters/pull/477/commits/1fa89bb254cde2286c92b950e7f892a75f8c89a0

dkoo (changes requested):
Ah, much better, thanks! But now I just noticed (not sure if this was the case before) that if you click on the social icons container to select the block, the appender button appears invisible, so it's not possible to add more social media links:

Screen Shot 2021-06-17 at 9 52 05 AM

This is in contrast to the same block when used in a post or page, where the appender appears as a floated icon over the container:

Screen Shot 2021-06-17 at 9 54 53 AM

Wonder if our custom editor styles are conflicting with this button's styles, somehow?

thomasguillot:
> Wonder if our custom editor styles are conflicting with this button's styles, somehow?

I did notice this and was going to open a separate issue. I'm not sure what's happening here.

dkoo (approved):
I'm going to approve this since we can handle the social media icon appender bug in a separate PR. 👍

PR #489, opened 2021-07-02
Persist user defined test emails as user meta

### LINKED ISSUE #156: Prepopulate Test Email With User Account Email
A nice enhancement would be to have the "Send A Test To" field come pre-populated with the email associated with the current WordPress user.

miguelpeixe:
### All Submissions:

Populate the field for email testing with the address(es) previously set by the user. If no email has been previously set, populates the field with the user email address.

Closes #156.

1. Write a newsletter
2. Send a test to a custom email address (or multiple comma-separated addresses)
3. Refresh the page or create a new newsletter and see that the emails have persisted

### Other information:

### DIFF
--- a/src/newsletter-editor/testing/index.js
+++ b/src/newsletter-editor/testing/index.js
@@ -28,7 +28,7 @@ export default compose( [
 	} ),
 ] )( ( { apiFetchWithErrorHandling, inFlight, postId, savePost, setInFlightForAsync } ) => {
 	const [ testEmail, setTestEmail ] = useState(
-		window?.newspack_newsletters_data?.user_test_emails?.join( ',' )
+		window?.newspack_newsletters_data?.user_test_emails?.join( ',' ) || ''
 	);
 
 	const sendTestEmail = async () => {
--- a/src/newsletter-editor/testing/index.js
+++ b/src/newsletter-editor/testing/index.js
@@ -4,7 +4,7 @@
 import { __ } from '@wordpress/i18n';
 import { withSelect, withDispatch } from '@wordpress/data';
 import { compose } from '@wordpress/compose';
-import { Fragment, useState } from '@wordpress/element';
+import { Fragment } from '@wordpress/element';
 import { Button, Spinner, TextControl } from '@wordpress/components';
 import { hasValidEmail } from '../utils';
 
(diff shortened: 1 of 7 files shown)


dkoo (changes requested):
Works as described! The only issue I saw is that if you update the email addresses in the input field, then collapse the Testing sidebar panel, the value will revert to the original value if you re-expand the Testing panel (even after sending a test email). We'll need to figure out a way to update the state in that case.

The easiest way I can think of is to move the state value from the `Testing` component up to the `NewsletterEdit` component and then pass it as a prop to `Testing` so that the state will persist even if `Testing` is unmounted.

miguelpeixe:
I've also changed a little bit of the field behavior. Now the default email is the user email address. If the user sets a new email, it is saved as user meta.

src/newsletter-editor/testing/index.js
```diff
@@ -26,48 +26,48 @@ export default compose( [
 			savePost,
 		};
 	} ),
-] )( ( { apiFetchWithErrorHandling, inFlight, postId, savePost, setInFlightForAsync } ) => {
-	const [ testEmail, setTestEmail ] = useState(
-		window?.newspack_newsletters_data?.user_test_emails?.join( ',' ) || ''
-	);
+] )(
+	( { apiFetchWithErrorHandling, inFlight, postId, savePost, setInFlightForAsync, ...props } ) => {
+		const { testEmail, onChangeEmail } = props;
```
dkoo, line 31: Non-blocking nitpick: why not destructure these variables on the previous line along with the other props?

dkoo (approved):
Works well, and I like how the field now defaults to the user's email! (Which is what #156 originally suggested, so now we have the best of both worlds.) I have a non-blocking question below, but feel free to merge if you did it for a specific reason.

src/newsletter-editor/testing/index.js
miguelpeixe: You're right! I wanted to have them together for legibility but doesn't look like we have a problem with having them together with the other component props.

dkoo (approved):

src/newsletter-editor/testing/index.js
dkoo: Got it, looks good now!

PR #496, opened 2021-07-09
Add "manual" provider for email markup copy

### LINKED ISSUE #368: After initial setup, the pre-send modal displays wrong content
1. Active plugin afresh (so that the are no plugin options saved)
2. Create a new newsletter, observe the setting modal prompts for API keys and all
3. Fill in the necessary keys
4. Input the necessary stuff in the sidebar settings, so the newsletter is sendable
5. Click "Send", observe the pre-send modal displays the example-provider version:

image

6. Save draft and reload the site, observe the pre-send modal will now display proper content (dependent on the ESP):

image

---

This seems to be related to the value of `newspack_newsletters_data.is_service_provider_configured` global being `""`.

### LINKED ISSUE #470: "Generic" email provider integration
There are too many email providers out there for us to integrate with all of them. We could increase our compatibility by having a "generic" (better name TBD) email provider option which outputs valid e-mail markup to the clipboard, where it can then be manually pasted into any email provider. This would be a good compatibility feature and would certainly increase adoption of the plugin.

miguelpeixe:
### All Submissions:

Implements "manual" provider to accommodate features from #470.

Although the issue mentions a "generic" provider approach, this PR interprets the required features as a manual approach, as it does not consist in new provider in its concept nor in our [ESP service interface](https://github.com/Automattic/newspack-newsletters/blob/master/includes/service-providers/interface-newspack-newsletters-esp-service.php). The manual option is available as a provider option for the user, which is interpreted in the editor to implement the necessary features.

The implemented features/changes in the editor are:
 - Isolate the "preview text" input into `renderPreviewText` so it can be available outside of a provider sender context.
 - Added HTML markup copy to clipboard in pre-send info modal.
 - Added support for post-update modal (which  `manual` implements to provide the updated HTML)
 - Removed "test email" when the provider is set to "Manual".

Closes #470.
Closes #368 (see 2d70899d3fa4de6c7716cb9d7289e3e0c469a869 commit message).

1. Run `npm ci && npm run build`
2. Modify your provider settings to "Manual" (make sure to clear Mailchimp API Key, as it enforces the provider to Mailchimp due to legacy reasons)
3. Create a new Newsletter and select a layout
4. Click on **Send**
5. Observe the new modal with the "Copy to clipboard" option
6. Click on Mark as sent (which sets the newsletter to `publish`)

After the newsletter is marked as sent, you can still make changes and obtain the updated HTML by clicking on `Update and copy HTML` button.

### Other information:

### DIFF
--- a/includes/class-newspack-newsletters.php
+++ b/includes/class-newspack-newsletters.php
@@ -118,6 +118,22 @@ final class Newspack_Newsletters {
 	public static function register_meta() {
 		\register_meta(
 			'post',
+			'without_provider',
+			[
+				'object_subtype' => self::NEWSPACK_NEWSLETTERS_CPT,
+				'show_in_rest'   => [
+					'schema' => [
+						'context' => [ 'edit' ],
+					],
+				],
+				'type'           => 'boolean',
+				'single'         => true,
+				'auth_callback'  => '__return_true',
+			]
+		);
+
+		\register_meta(
+			'post',
 			'mc_campaign_id',
 			[
 				'object_subtype' => self::NEWSPACK_NEWSLETTERS_CPT,
@@ -447,6 +463,12 @@ final class Newspack_Newsletters {
 	public static function save( $post_id, $post, $update ) {
 		if ( ! $update ) {
 			update_post_meta( $post_id, 'template_id', -1 );
+
+			// Set "without provider" post meta if using manual.
+			$provider = self::service_provider();
+			if ( 'manual' === $provider ) {
+				update_post_meta( $post_id, 'without_provider', true );
+			}
 		}
 	}
 
@@ -941,6 +963,10 @@ final class Newspack_Newsletters {
 		}
 
 		if ( self::$provider && self::$provider->has_api_credentials() ) {
+			$response['status'] = true;
+		}
+
+		if ( 'manual' === $service_provider ) {
 			$response['status'] = true;
 		}
 
--- a/includes/class-newspack-newsletters.php
+++ b/includes/class-newspack-newsletters.php
@@ -118,22 +118,6 @@ final class Newspack_Newsletters {
 	public static function register_meta() {
 		\register_meta(
 			'post',
-			'without_provider',
-			[
-				'object_subtype' => self::NEWSPACK_NEWSLETTERS_CPT,
-				'show_in_rest'   => [
-					'schema' => [
-						'context' => [ 'edit' ],
-					],
-				],
-				'type'           => 'boolean',
-				'single'         => true,
-				'auth_callback'  => '__return_true',
-			]
-		);
-
-		\register_meta(
-			'post',
 			'mc_campaign_id',
 			[
 				'object_subtype' => self::NEWSPACK_NEWSLETTERS_CPT,
@@ -463,12 +447,6 @@ final class Newspack_Newsletters {
 	public static function save( $post_id, $post, $update ) {
 		if ( ! $update ) {
 			update_post_meta( $post_id, 'template_id', -1 );
-
-			// Set "without provider" post meta if using manual.
-			$provider = self::service_provider();
-			if ( 'manual' === $provider ) {
-				update_post_meta( $post_id, 'without_provider', true );
-			}
 		}
 	}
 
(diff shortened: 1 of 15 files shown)


includes/class-newspack-newsletters.php
```diff
@@ -447,6 +463,12 @@ public static function register_meta() {
 	public static function save( $post_id, $post, $update ) {
 		if ( ! $update ) {
 			update_post_meta( $post_id, 'template_id', -1 );
+
+			// Set "without provider" post meta if using manual.
+			$provider = self::service_provider();
+			if ( 'manual' === $provider ) {
+				update_post_meta( $post_id, 'without_provider', true );
```
dkoo, line 470: With this stored as post meta, it means that any newsletter drafts that are created while the provider is set to "manual" will always be manual-only. If I then change my provider to Mailchimp, those drafts will not be sendable via Mailchimp or any other ESP. Is this intended?

dkoo (changes requested):
I wasn't able to get far because I ran into an error at step 3: after setting my provider to "manual", when I create a newsletter, I see the UI that prompts me to select a provider:

Screen Shot 2021-07-12 at 5 08 37 PM

If I then set this UI to "Manual" and click "Save settings", I encounter a PHP fatal:

```
PHP Fatal error:  Uncaught Error: Call to a member function set_api_credentials() on null in /Users/dkoo/Local Sites/newspack/app/public/wp-content/plugins/newspack-newsletters/includes/class-newspack-newsletters.php:935
Stack trace:
#0 /Users/dkoo/Local Sites/newspack/app/public/wp-includes/rest-api/class-wp-rest-server.php(1140): Newspack_Newsletters::api_set_settings(Object(WP_REST_Request))
#1 /Users/dkoo/Local Sites/newspack/app/public/wp-includes/rest-api/class-wp-rest-server.php(987): WP_REST_Server->respond_to_request(Object(WP_REST_Request), '/newspack-newsl...', Array, NULL)
#2 /Users/dkoo/Local Sites/newspack/app/public/wp-includes/rest-api/class-wp-rest-server.php(414): WP_REST_Server->dispatch(Object(WP_REST_Request))
#3 /Users/dkoo/Local Sites/newspack/app/public/wp-includes/rest-api.php(370): WP_REST_Server->serve_request('/newspack-newsl...')
#4 /Users/dkoo/Local Sites/newspack/app/public/wp-includes/class-wp-hook.php(303): rest_api_loaded(Object(WP))
#5 /Users/dkoo/Local Sites/newspack/app/public/wp-includes/c in /Users/dkoo/Local Sites/newspack/app/public/wp-content/plugins/newspack-newsletters/includes/class-newspack-newsletters.php on line 935
```

I'm not able to proceed past this point into the editor.

includes/class-newspack-newsletters.php
miguelpeixe: It was intended until your comment made me think about drafts. First I thought only about published newsletters, once they are _marked as sent_ there wouldn't be a reason to handle this post with another provider.

Now I realize that a user could easily start drafting newsletters in "manual mode" before finishing setting up their provider. Because of that, I've removed the `without_provider` post meta and we now rely solely on the current provider data to display the "manual provider" features.

We are still not completely able to smoothly take over a draft with a new provider. Mailchimp, for example, always attempts to fetch the `mc_campaign_id`, which is not available for drafts that were initially made in manual provider:

image

Once you save the draft again, the sync method fixes the error message by creating the campaign and storing its ID. I'll open an issue for that.

miguelpeixe:
I missed an API check when we update the settings with the `manual` provider, which is fixed now.

While reviewing the recent changes, I stumbled on #368. The `getServiceProvider()` on `send-button.js` was moved to the component render so it can get the updated provider value from `initModal`, which fixes the issue.

dkoo:
> We are still not completely able to smoothly take over a draft with a new provider. Mailchimp, for example, always attempts to fetch the `mc_campaign_id`, which is not available for drafts that were initially made in manual provider

I've noticed this, too! I feel like we should handle this situation. I've filed an issue for that [here](https://github.com/Automattic/newspack-newsletters/issues/499).

dkoo (approved):
This works as described now, and thanks for fixing #368 too! This feature will be even more useful for users who don't want to use one of the supported ESPs.

The only thing that gives me pause is that "Mark as sent" might not be totally clear at first that it will actually publish the newsletter, which might be a it of a surprise if the draft is marked to be publicly visible. I don't think this is a blocker, though, because it makes sense—it's just not super obvious at first glance.

miguelpeixe:
I agree @dkoo! I've just updated the modal button so it displays "Mark as sent and publish" when in manual provider with the public option toggled on.

dkoo (approved):
Thanks for adding the extra copy—I think that makes it a bit clearer what's happening when you mark as sent.

PR #498, opened 2021-07-13
fix: change allowed_block_types filter name

adekbadek:
### All Submissions:

Changes `allowed_block_types` filter name to `allowed_block_types_all`, since the former is deprecated in WP 5.8.

1. With WP 5.8 and on `master` branch, obsere a PHP notice about `allowed_block_types` being deprecated on post editor
2. Switch to this branch, observe the notice is not shown
3. Switch to WP <5.8, ensure the blocks are still filtered on Newsletter editor, e.g. a Jetpack Mailchimp block can't be added

### Other information:

### DIFF
--- a/includes/class-newspack-newsletters-editor.php
+++ b/includes/class-newspack-newsletters-editor.php
@@ -46,7 +46,7 @@ final class Newspack_Newsletters_Editor {
 	public function __construct() {
 		add_action( 'the_post', [ __CLASS__, 'strip_editor_modifications' ] );
 		add_action( 'enqueue_block_editor_assets', [ __CLASS__, 'enqueue_block_editor_assets' ] );
-		add_filter( 'allowed_block_types', [ __CLASS__, 'newsletters_allowed_block_types' ], 10, 2 );
+		add_filter( 'allowed_block_types_all', [ __CLASS__, 'newsletters_allowed_block_types' ], 10, 2 );
 		add_action( 'rest_post_query', [ __CLASS__, 'maybe_filter_excerpt_length' ], 10, 2 );
 		add_action( 'rest_api_init', [ __CLASS__, 'add_newspack_author_info' ] );
 		add_filter( 'the_posts', [ __CLASS__, 'maybe_reset_excerpt_length' ] );
@@ -283,7 +283,7 @@ final class Newspack_Newsletters_Editor {
 
 	/**
 	 * Get current user test emails
-	 * 
+	 *
 	 * @return array List of user defined emails.
 	 */
 	public static function get_current_user_test_emails() {


miguelpeixe (changes requested):
The `allowed_block_types_all` filter does not exist prior to 5.8. When switching to WP < 5.8 all the block types become available.

We might have to duplicate this filter call for legacy reasons.

I'm still not sure why, but even with this change I still see the following notice:
```
Deprecated: block_categories is deprecated since version 5.8.0! Use block_categories_all instead.
```

dkoo (approved):
Nice, I was wondering where those PHP warnings were coming from!

dkoo (changes requested):
@miguelpeixe makes a good point about supporting WP 5.7.X, so I'll remove my approval for now until we decide one way or another. I wasn't seeing the PHP warning Miguel mentioned after switching to this branch, FWIW.

miguelpeixe:
> I wasn't seeing the PHP warning Miguel mentioned after switching to this branch, FWIW.

It's probably not related to newsletters. The plugin does not create any new block category that could trigger this warning.

miguelpeixe:
I misunderstood the usage of [`apply_filters_deprecated`](https://developer.wordpress.org/reference/functions/apply_filters_deprecated/) 🤦 . It's not a valid suggestion.

dkoo:
This is what [WP's documentation](https://developer.wordpress.org/block-editor/reference-guides/filters/block-filters/#allowed_block_types_all) says about continuing to support WP versions prior to 5.7:

> Before WordPress 5.8 known as `allowed_block_types`. In the case when you want to support older versions of WordPress you might need a way to detect which filter should be used – the deprecated one vs the new one. The recommended way to proceed is to check if the `WP_Block_Editor_Context` class exists.

The question in my mind is, do we care to support WP versions prior to 5.8? All managed Newspack sites will be moved to 5.8 automatically. Continuing to support 5.7 and below would potentially benefit self-hosted sites who might be waiting to upgrade for whatever reason.

adekbadek:
> Continuing to support 5.7 and below would potentially benefit self-hosted sites who might be waiting to upgrade for whatever reason.

I'd follow core in this respect - if the legacy filter is deprecated, let's not support it. If the block filter won't work, and the user inserts an unsupported block, they will likely catch this issue when viewing a test email - the potential for disruption is minimal.

miguelpeixe (approved):

dkoo (approved):
> I'd follow core in this respect - if the legacy filter is deprecated, let's not support it. If the block filter won't work, and the user inserts an unsupported block, they will likely catch this issue when viewing a test email - the potential for disruption is minimal.

I'm okay this this. Should we consider posting a blog post about things like this that might work differently in WP 5.7 vs 5.8 or no longer be supported in our plugins after the 5.8 release?

adekbadek:
> Should we consider posting a blog post about things like this that might work differently in WP 5.7 vs 5.8 or no longer be supported in our plugins after the 5.8 release?

Are there more features that would not work in <5.8?

In the case of this plugin, let's add a note in release notes.

Issue #503, opened 2021-07-15
Not able to create newsletter draft with Constant Contact provider


miguelpeixe:
While attempting to start a new newsletter with Constant Contact as my provider, I get the following fatal error:

```
Fatal error: Uncaught TypeError: Argument 3 passed to GuzzleHttp\Client::request() must be of the type array, string given, called in newspack-plugin/vendor/guzzlehttp/guzzle/src/Client.php on line 92 and defined in newspack-plugin/vendor/guzzlehttp/guzzle/src/Client.php on line 184

Call Stack
--
1 | 0.0001 | 375496 | {main}( ) | .../post-new.php:0
2 | 0.2890 | 12181408 | get_default_post_to_edit( ) | .../post-new.php:66
3 | 0.2890 | 12181784 | wp_insert_post( ) | .../post.php:690
4 | 0.2947 | 12249032 | do_action( ) | .../post.php:4373
5 | 0.2947 | 12249408 | WP_Hook->do_action( ) | .../plugin.php:484
6 | 0.2947 | 12249408 | WP_Hook->apply_filters( ) | .../class-wp-hook.php:316
7 | 0.2963 | 12244056 | Newspack_Newsletters_Constant_Contact->save( ) | .../class-wp-hook.php:292
8 | 0.2963 | 12244056 | Newspack_Newsletters_Constant_Contact->sync( ) | .../class-newspack-newsletters-constant-contact.php:374
9 | 0.2980 | 12402720 | Ctct\Services\AccountService->getAccountInfo( ) | .../class-newspack-newsletters-constant-contact.php:314
10 | 0.2981 | 12402816 | Ctct\Services\AccountService->createBaseRequest( ) | .../AccountService.php:94
11 | 0.2981 | 12402816 | GuzzleHttp\Client->createRequest( ) | .../BaseService.php:65
12 | 0.2981 | 12403192 | GuzzleHttp\Client->__call( ) | .../BaseService.php:65
13 | 0.2981 | 12403192 | GuzzleHttp\Client->request( ) | .../Client.php:92
```

miguelpeixe:
The Constant Contact PHP SDK is [pretty much unmaintained](https://github.com/constantcontact/php-sdk/releases) and we should probably connect to the API with our own methods.

dkoo:
[...]

PR #505, opened 2021-07-16
Fix: sync instead of erroring for missing campaign id

### LINKED ISSUE #499: Error for existing drafts after switching ESPs
For Mailchimp and Constant Contact ESPs, a campaign in the currently selected ESP is created when a new newsletter draft is created, then kept in sync upon update. However, if you change your selected ESP after creating a new draft, the plugin will assume that the draft has a corresponding campaign on the ESP side exists, and when it doesn't, it will throw an error. The only way to resolve this is to create a new newsletter draft.

### Expected behavior

If the plugin can't find a campaign to sync to in the currently selected ESP, it should silently create it instead of erroring out.

miguelpeixe:
### All Submissions:

Execute `sync()` method instead of return error when `retrieve()` method is unable to identify an existing campaign id.

Warning: Although the change in Constant Contact provider follows the same logic as for Mailchimp, I was not able to test it due to #503.

Closes #499.

1. In `master` branch, create a newsletter draft using the Manual provider
2. Change your provider to Mailchimp with valid API keys
3. Edit the created draft and see the "no campaign ID" error.
4. Switch to this PR branch, repeat steps and see that no error is displayed and the newsletter is synchronized with the ESP.

Steps should also be executed using Constant Contact instead of Mailchimp.

### Other information:

### DIFF
--- a/includes/service-providers/constant_contact/class-newspack-newsletters-constant-contact.php
+++ b/includes/service-providers/constant_contact/class-newspack-newsletters-constant-contact.php
@@ -469,7 +469,7 @@ final class Newspack_Newsletters_Constant_Contact extends \Newspack_Newsletters_
 	public function retrieve_campaign_id( $post_id ) {
 		$cc_campaign_id = get_post_meta( $post_id, 'cc_campaign_id', true );
 		if ( ! $cc_campaign_id ) {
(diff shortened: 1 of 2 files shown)


dkoo (approved):
Nice and simple solution. Works exactly as I would expect.

PR #537, opened 2021-08-05
fix: do not show mc_campaign_id in rest

### LINKED ISSUE #535: Mailchimp synchronization generates multiple draft campaigns
Using Mailchimp provider, If I edit a newsletter that has already been created with a different provider, the synchronization creates repeated campaign drafts.

The newsletter does not stick with a campaign ID, clearing and generating new campaigns every save.

miguelpeixe:
### All Submissions:

Similar to what has been experienced with Constant Contact provider, showing the `mc_campaign_id` in rest causes an issue with synchronization:

As shared on #519:
> I struggled a little bit to implement the ability to synchronize the newsletter with the provider after the draft was created (either by the post being created with a different provider or before the tokens were available). I wasn't able to identify the root of this sync issue, but I identified that it was caused by the cc_campaign_id meta being available on the rest API. At some save cycle, the meta was being deleted and the post was desynced from the provider. Removing the meta from the API (see 2b1061b) solved the issue. I figured it's not a problem, since we don't (and probably shouldn't) manage this data on the frontend.

Closes #535.

1. In the master branch, create a newsletter using the manual provider as save as draft.
2. Change your provider to Mailchimp with a valid API key
3. Edit the newsletter by making a few changes and saving the draft a couple of times
4. Observe on your Mailchimp dashboard that you have multiple duplicated drafts
4. Checkout this branch, repeat steps and verify that you only have 1 draft

### Other information:


adekbadek (approved):
Works as expected. Have you considered moving the meta field registration to `includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp-controller.php`? I think it's where it is only as a remnant of the single-provider architecture this plugin started with.

miguelpeixe:
> Have you considered moving the meta field registration

Good idea! 6a580c44577e8a07af55158d86d65d32bc56a7a7 should do it.

miguelpeixe:
Oops, I moved it to the provider class instead of the controller. 13bf763a5eb59546f2ee6043f22670c9df5389a8 is the actual fix.

adekbadek (approved):

PR #544, opened 2021-08-12
feat(post inserter): add post content option

### LINKED ISSUE #478: Feature request: Post block inserter - Support full post content, not just excerpt
Our newsletters are comprised of the full content of our (rather short) posts. Post inserter seems to only let us include a post excerpt. Would like to request the ability to include the full body content of the post.

kariae:
### All Submissions:

add post content as an option for the post inserter, it will disable the post excerpt if selected and
vice versa, since it doesn't make sense to have both post excerpt and content in the same post

fix #478

1. Pull the branch of this PR.
2. On a new newsletter, add a post-inserter block.
3. Observe the new option to add the post content.
4. Check the post content checkbox and observe that the post excerpt is removed and the post content is added instead.

### Other information:

### DIFF
--- a/src/editor/blocks/posts-inserter/utils.js
+++ b/src/editor/blocks/posts-inserter/utils.js
@@ -51,6 +51,16 @@ const getSubtitleBlockTemplate = ( post, { subHeadingFontSize, textColor } ) => 
 	return [ 'core/heading', assignFontSize( subHeadingFontSize, attributes ) ];
 };
 
+const getContentBlockTemplate = ( post, { textFontSize, textColor } ) => {
+	let content = post.content.rendered;
+	const contentElement = document.createElement( 'div' );
+	contentElement.innerHTML = content;
+	content = contentElement.textContent || contentElement.innerText || '';
+
+	const attributes = { content: content.trim(), style: { color: { text: textColor } } };
+	return [ 'core/paragraph', assignFontSize( textFontSize, attributes ) ];
+};
+
 const getExcerptBlockTemplate = ( post, { excerptLength, textFontSize, textColor } ) => {
 	let excerpt = post.excerpt.rendered;
 	const excerptElement = document.createElement( 'div' );
@@ -126,6 +136,9 @@ const createBlockTemplatesForSinglePost = ( post, attributes ) => {
 	}
 	if ( attributes.displayPostDate && post.date_gmt ) {
 		postContentBlocks.push( getDateBlockTemplate( post, attributes ) );
+	}
+	if ( attributes.displayPostContent ) {
+		postContentBlocks.push( getContentBlockTemplate( post, attributes ) );
 	}
 	if ( attributes.displayPostExcerpt ) {
 		postContentBlocks.push( getExcerptBlockTemplate( post, attributes ) );
--- a/src/editor/blocks/posts-inserter/block.json
+++ b/src/editor/blocks/posts-inserter/block.json
@@ -17,6 +17,10 @@
 			"type": "boolean",
 			"default": false
 		},
+	  	"displayPostContent": {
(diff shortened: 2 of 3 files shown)


src/editor/blocks/posts-inserter/utils.js
```diff
@@ -51,6 +51,16 @@ const getSubtitleBlockTemplate = ( post, { subHeadingFontSize, textColor } ) =>
 	return [ 'core/heading', assignFontSize( subHeadingFontSize, attributes ) ];
 };
 
+const getContentBlockTemplate = ( post, { textFontSize, textColor } ) => {
+	let content = post.content.rendered;
+	const contentElement = document.createElement( 'div' );
+	contentElement.innerHTML = content;
+	content = contentElement.textContent || contentElement.innerText || '';
```
miguelpeixe, line 58: The use of this strategy can result in a very long unformatted paragraph for posts with larger content. I also experienced a crash while attempting to display a post with custom shortcodes.

miguelpeixe (changes requested):
Due to what I commented inline, I believe we need to investigate what would be the strategy for displaying full post content, or even if this is a viable option for the purpose of the Posts Inserter block.

We can't display the post content as it's displayed on the website due to email limitations, hence the restriction of blocks in the newsletters editor. At the same time, I don't think it's ideal to display as a big chunk of text.

@dkoo @adekbadek, do you have thoughts on this?

src/editor/blocks/posts-inserter/utils.js
kariae: I do agree with you @miguelpeixe, I would suggest setting the article content (with the word limit that we have currently) only if the post doesn't have an excerpt, but again, would like to have more thoughts on this.

adekbadek: An auto-excerpting feature would be nice. If the post does not have an excerpt, the plugin could attempt to take a chunk of the text and use it as excerpt. This would not require any changes in the UI, but work under the hood.

dkoo: Not sure if this is helpful, but I built a function to build [auto-excerpts for Newspack Listings](https://github.com/Automattic/newspack-listings/blob/master/includes/newspack-listings-utils.php#L170-L254) that mimics the core excerpt functionality, but allows certain HTML elements as well. You could maybe adapt this with an absolute max length to avoid bloating the newsletter content.

kariae: Go to for the auto-excerpting then, @dkoo this will help a lot thank you. I'll adapt this to the already max length option that we have.

miguelpeixe: Also agreed! In that case, we wouldn't have the "content" as a toggle option on the settings, right?

kariae: Nope, I'll remove the toggle, if the post contains an excerpt it's all good if not I'll generate one from the post content, the excerpt generated can be toggled on/off via the excerpt toggle.

kariae: I just realized that this is the current behavior 😅; having posts without an excerpt, when added on the post inserter block, the excerpt is automatically generated from the content based on the excerpt max size. Can someone confirm this, please?

adekbadek: Yes, looks like it's just what WP does :D Well then, case closed?

miguelpeixe: Agreed! @kariae, feel free to close this PR :)

kariae:
Closed since it's not viable as discussed in the comments to add the post content in the newsletter posts.

PR #548, opened 2021-08-16
fix: custom namespace for Guzzle

### LINKED ISSUE #539: Change how to handle composer dependencies and version conflicts
#519 introduces the use of Jetpack Autoloader with the intention to avoid conflict with plugins that use an older version of matching dependencies, specially Guzzle that has a lot of [breaking changes from version 5 to 6](https://github.com/guzzle/guzzle/blob/master/UPGRADING.md#50-to-60).

What I've missed in its implementation is that it also relies on other plugins to also use the autoloader in order to load the latest version. As @kbrown9 kindly [shared with us](https://github.com/Automattic/newspack-newsletters/pull/519#issuecomment-892983578):

> The Jetpack Autoloader can only control which version is loaded when every plugin with that dependency uses the Jetpack Autoloader. If any plugin on the site uses the dependency but does not use the Jetpack Autoloader (for example, if it uses the Composer autoloader instead), the JP Autoloader can't fully control which version gets loaded. So it's possible that you will still run into situations where an older version of Guzzle is loaded first.

She also suggests we use a custom namespace to dependencies, like [Yoast does](https://plugins.trac.wordpress.org/browser/wordpress-seo/tags/16.8/vendor_prefixed?order=name), which I also believe is the better approach.

miguelpeixe:
### All Submissions:

Implements [`php-scoper`](https://github.com/humbug/php-scoper) to add a custom prefix to Guzzle namespace.

Closes #539.

1. In the master branch, install the plugin [Constant Contact Forms](https://wordpress.org/plugins/constant-contact-forms/) (it uses an older version of Guzzle that causes conflict)
2. With Constant Contact provider configured, attempt to create a new newsletter
3. Observe the fatal error
4. Switch to this branch, run `composer install` and confirm newsletter create/edit/publish works
5. Also confirm that the release build works:
    1. Clear the `vendor/` and `vendor_prefixed/` directories (**do not remove the `vendor_prefixed/` entirely, leave it empty instead**)
    2. Run `composer install --no-dev --no-scripts && composer dump-autoload && npm run release:archive`
    3.  Install generated zip file in a new WordPress installation and confirm the plugin works with Constant Contact set as provider and with the plugin mentioned in the first step

### Other information:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -75,7 +75,7 @@ jobs:
       - checkout_with_workspace
       - run:
           name: Install PHP dependencies
-          command: composer install --no-dev --no-scripts
+          command: composer install --no-dev --no-scripts && composer dump-autoload
       - run:
           name: Install rsync
           command: sudo apt install rsync
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -41,7 +41,7 @@ jobs:
 
   test-php:
     docker:
-      - image: circleci/php:7.2-node-browsers
+      - image: circleci/php:7.2
       - image: circleci/mysql:5.6
     environment:
       - WP_TESTS_DIR: '/tmp/wordpress-tests-lib'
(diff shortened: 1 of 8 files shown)


claudiulodro (approved):
Works well locally and built on a new site in my testing. Good solution!

PR #552, opened 2021-08-17
Prefix classes before generating autoload files

claudiulodro:
### All Submissions:

This PR fixes and issue @miguelpeixe noticed while building the releases: the namespaced Guzzle classes weren't getting included in the composer autoload. This is because `post-composer-autoload` runs after the autoload file is generated. Switching to `pre-composer-autoload` solves the issue.

1. Before checking out this PR, run `rm -r vendor; rm -r vendor_prefixed/*`, then run `composer install --no-dev --no-scripts && composer dump-autoload`, then run `cat vendor/composer/autoload_classmap.php`. You should not see the Guzzle classes in the classmap.
2.  Check out this PR, run all those above commands again. You should see the Guzzle classes in the classmap.

### Other information:

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,7 +33,7 @@
 		"post-update-cmd": [
 			"vendor/bin/cghooks update"
 		],
-		"post-autoload-dump": [
+		"pre-autoload-dump": [
 			"composer prefix-guzzle"
 		]
 	},


claudiulodro:
This causes issues building Newspack Newsletters. From the prefixer docs:

> Warning: After prefixing the files, if you are relying on Composer for the autoloading, dumping the autoloader again is required.

We'll need to re-approach

PR #556, opened 2021-08-19
Refactor Mailchimp recipients and segmentation options

### LINKED ISSUE #22: Support For Mailchimp Segments
Add support for Mailchimp segments: https://us7.admin.mailchimp.com/lists/members?id=453669&previewName=Contacts#p:1-s:10-sa:last_update_time-so:false

miguelpeixe:
### All Submissions:

This PRs proposes a change on how we support segmentation with the Mailchimp provider. Currently, we have support for selecting a group and multiple tags. The tags implementation also included saved segments, although not supported due to the segmentation options for `saved` segments being different for `static` segments (tags).

The redesign fixes the segmentation support and proposes a simpler approach, similar to how Mailchimp's wizard works, in which you are able to select one of the available groups, tags, or segments. For what we generally propose in terms of form complexity, this approach avoids scenarios in which the user would select a group and different segments that don't match existing recipients.

It also removes the `Use Mailchimp tags` field. It looks like an unnecessary block for accessing segmentation options.

image

image

Closes #22.

1. Switch to this branch and run `npm run build`
2. Test selecting different segmentation options and publishing campaigns

### Other information:

### DIFF
--- a/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php
+++ b/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php
@@ -546,12 +546,15 @@ final class Newspack_Newsletters_Mailchimp extends \Newspack_Newsletters_Service
 	 */
 	public function audience_segments( $post_id, $target_id ) {
 
+		$interest_id = false;
+		$segment_id  = false;
+
 		// Determine if we're dealing with an interest or a segment.
 		if ( false !== strpos( $target_id, ':' ) ) {
 			$exploded    = explode( ':', $target_id );
 			$field       = count( $exploded ) ? $exploded[0] : null;
 			$interest_id = count( $exploded ) > 1 ? $exploded[1] : null;
-		} else {
+		} elseif ( '' !== $target_id ) {
 			$segment_id = $target_id;
 		}
 
--- a/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php
+++ b/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php
@@ -170,13 +170,19 @@ final class Newspack_Newsletters_Mailchimp extends \Newspack_Newsletters_Service
 				}
 			}
 
-			$tags = [];
+			$segments = [];
 			if ( $list_id ) {
-				$tags_response = $this->validate(
-					$mc->get( "lists/$list_id/segments?count=1000", [], 20 ),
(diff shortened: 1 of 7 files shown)


includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php
```diff
@@ -561,6 +562,14 @@ public function audience_segments( $post_id, $compound_interest_id, $tag_ids ) {
 				__( 'Mailchimp campaign ID not found.', 'newspack-newsletters' )
 			);
 		}
+
+		if ( '' !== $target_id && ! $interest_id && ! $segment_id ) {
```
adekbadek, line 566: If the condition on line 550 is not met, `$interest_id` will not be defined. It should either be defined before the condition, or its existence checked here.

src/service-providers/mailchimp/ProviderSidebar.js
```diff
@@ -3,75 +3,71 @@
...
+	}
 
 	return (
 		<Fragment>
-			{ availableInterests.length ? (
+			{ options.length ? (
 				<SelectControl
-					label={ __( 'Groups', 'newspack-newsletters' ) }
-					value={ segmentsData.interest_id }
+					label={ __( 'Group, segment or tag', 'newspack-newsletters' ) }
```
adekbadek, line 61: Oxford-comma nitpick ;) : 
```suggestion
					label={ __( 'Group, segment, or tag', 'newspack-newsletters' ) }
```

adekbadek (changes requested):
Works well! Just two nitpicks.

Also, perhaps for a later PR - the pre-send modal could indicate that a segment or group was chosen. Here, I know my list has more contacts than 1 (but with the chosen segment the newsletter will be sent to just one contact), so it's weird to see `1 subscriber`:

image

includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php
miguelpeixe: Nice catch! 8aba70be868cc7d36a43c1c05786838e9a95d8f5 fixes it.

miguelpeixe:
> Also, perhaps for a later PR - the pre-send modal could indicate that a segment or group was chosen. Here, I know my list has more contacts than 1 (but with the chosen segment the newsletter will be sent to just one contact), so it's weird to see `1 subscriber`

That would certainly be a nice addition, I filed #558 for this.

adekbadek (approved):

PR #583, opened 2021-09-01
feat: add support for embed

### LINKED ISSUE #386: Support embeds
Some embeds, like video, don't translate well to emails. A solution might be to produce a screenshot of an embed, and insert is as an image linking to the original embed.

miguelpeixe:
### All Submissions:

Add support for the embed block. There are 4 types of embeddable content, as [specified on oEmbed](https://oembed.com/). For each available type we'll have the following behavior:

#### Video

Displays thumbnail with caption, which is the video title by default or the inline caption manually inserted. If no thumbnail is provided by the service, attempts to create a link with the title or caption.

#### Image

Displays image with caption, which is the image title by default or the inline caption manually inserted.

#### Rich

Validates HTML content, allowing `blockquote`, `p`, `a`, `br`, `em` and `strong`. If no valid HTML is returned after filtering, a link is generated with the resource title.

Examples are:

 - Twitter: returns a `blockquote`, which renders nicely
 - Amazon: returns an `iframe`, which becomes a link with the resource title

#### Link

Displays text with a link. The text is the resource title by default, or the inline caption manually inserted. I was not able to find a valid oembed provider that uses link as type but should work without problems.

Closes #386.

1. In this branch, create a new newsletter
2. Create several embed blocks with content from different services:
   1. YouTube
   2. Vimeo
   3. VideoPress
   3. Twitter
   4. Amazon
   5. Flickr
   6. Reddit
3. Send a test email and make sure the content is displayed as described above for each embed type

### Other information:

### DIFF
--- a/includes/class-newspack-newsletters-editor.php
+++ b/includes/class-newspack-newsletters-editor.php
@@ -97,6 +97,7 @@ final class Newspack_Newsletters_Editor {
 			'core/block',
 			'core/group',
 			'core/paragraph',
+			'core/embed',
 			'core/heading',
 			'core/column',
 			'core/columns',
--- a/includes/class-newspack-newsletters-embed.php
+++ b/includes/class-newspack-newsletters-embed.php
@@ -1,0 +1,72 @@
+<?php
+/**
+ * Newspack Newsletter Embed
+ *
+ * @package Newspack
+ */
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Newspack Newsletters Embed Class.
+ */
+final class Newspack_Newsletters_Embed {
+	/**
+	 * Allowed HTML tags for rich embeds.
+	 *
+	 * @var array
+	 */
+	public static $allowed_html = array(
+		'a'          => array(
+			'href'  => array(),
+			'title' => array(),
+		),
+		'blockquote' => array(),
+		'br'         => array(),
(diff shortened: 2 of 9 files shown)


adekbadek (commented):
Apart from `video` and `photo` types, the embeds are most likely to look much different in editor vs. in the email. WDYT about making it more WYSIWYG and displaying the embeds in the editor as they will appear in the email?

miguelpeixe:
Good idea! I spent some time experimenting with different possibilities and I think we have 2 options to better integrate the WYSIWYG aspect:

 - Create a new custom block that makes use of the already implemented oEmbed API so the WYSIWYG also reflects the web output for published newsletters.
 - Modify how the current `core/embed` block displays the email result but maintain rich embeds for the published web version.

What do you think?

adekbadek:
> Modify how the current core/embed block displays the email result but maintain rich embeds for the published web version.

This sounds like the best solution, maintaining WYSIWYG while enriching the public version.

miguelpeixe:
Agreed! The email preview has been implemented with new toolbar buttons so the user can switch between email and web previews:

https://user-images.githubusercontent.com/820752/132248704-743f7c15-8b87-428e-9af6-0dc8a749e9df.mov

adekbadek (approved):
Great idea with the preview mode!

PR #591, opened 2021-10-05
feat: block visibility attribute

### LINKED ISSUE #581: Conditional blocks for email/site newsletter content
Newsletters can have greetings and other information that is convenient and important for email content, but not so much for the version published on the website. This problem becomes more pressing with the implementation of #575.

A conditional block that groups blocks to configure which output to display (email or website) should solve it.

miguelpeixe:
### All Submissions:

Creates a "Visibility" block attribute so that blocks can be configured to be displayed for email or web versions only.

The "Support" newsletter template has also been updated so that the `Dear *|FNAME|*,` paragraph is automatically set to "Email only".

The new attribute is available for every block in the newsletter editor and is filtered without distinction for both web and email rendering.

Closes #581.

1. Check out this branch and run `npm run build`
2. Create a new newsletter and modify different blocks to be "Web only" and "Email only"
3. Make sure to test with Group, Image and Paragraph blocks
4. Set your newsletter public by clicking on "Make newsletter page public?" option
5. Send and publish
6. View both versions and confirm that the blocks are visible according to your selection

### Other information:

### DIFF
--- a/src/editor/blocks/visibility-attribute/index.js
+++ b/src/editor/blocks/visibility-attribute/index.js
@@ -136,13 +136,19 @@ const withVisibilityNotice = createHigherOrderComponent(
 					>
 						<span className="newsletters-block-visibility-label">
 							<Icon icon={ warning } size={ 15 } />
-							{ sprintf(
-								__( 'Only visible on the "%s" version.', 'newspack-newsletters' ),
-								value
+							{ shouldBePublic ? (
+								__(
+									'Newsletter is not public, this block will not be visible.',
+									'newspack-newsletters'
+								)
+							) : (
+								<Fragment>
+									{ value === 'web' &&
+										__( 'Only visible on public newsletter page.', 'newspack-newsletters' ) }
+									{ value === 'email' &&
+										__( 'Only visible in the sent email.', 'newspack-newsletters' ) }
+								</Fragment>
 							) }
-							{ shouldBePublic
-								? ' ' + __( 'Newsletter is not public.', 'newspack-newsletters' )
-								: null }
 						</span>
 						<BlockListBlock { ...props } />
 					</div>


src/editor/blocks/visibility-attribute/index.js
```diff
@@ -0,0 +1,84 @@
...
+					>
+						<SelectControl
+							label={ __( 'Select which output to display', 'newspack-newsletters' ) }
+							value={ value }
+							options={ visibilityOptions }
+							onChange={ selected => {
+								setAttributes( { [ ATTRIBUTE_NAME ]: selected } );
+							} }
+							help={ __(
+								"If your newsletter is going to be published on the web, select here if you'd like this block to be visible in a particular version",
```
adekbadek, line 62: Suggestion: I find "published on the web" somewhat confusing - sounds like it's about some _other_ site publishing it. 

```suggestion
								"If the newsletter is going to be viewable publicly on this site, select here if you'd like this block to be visible in a particular version.",
```

src/editor/blocks/visibility-attribute/index.js
```diff
@@ -0,0 +1,84 @@
...
+	return settings;
+};
+
+const withVisibilityControl = createHigherOrderComponent( BlockEdit => {
+	return props => {
+		const { attributes, setAttributes } = props;
+		const value = attributes[ ATTRIBUTE_NAME ];
+		return (
+			<Fragment>
+				<BlockEdit { ...props } />
```
adekbadek, line 48: WDYT about adding a notice above the component – if it's web- or email-only? With a more sophisticated layout, clicking each block to verify the sidebar setting is correct would be cumbersome.

src/editor/blocks/visibility-attribute/index.js
```diff
@@ -0,0 +1,84 @@
...
+				<InspectorControls>
+					<PanelBody
+						title={ __( 'Visibility options', 'newspack-newsletters' ) }
+						initialOpen={ true }
+					>
+						<SelectControl
+							label={ __( 'Select which output to display', 'newspack-newsletters' ) }
```
adekbadek, line 55: Hm, this sounds odd. How about

```suggestion
							label={ __( 'Where should this block be visible?', 'newspack-newsletters' ) }
```

src/editor/blocks/visibility-attribute/index.js
adekbadek: I'm not sure – it's a tough cookie. This feature has a lot of impact on the WYSIWYG aspect of the editing and I see a lot of potential for user error. Someone might mark the block as web-only, forget about it, and wonder why something did not make it the recipient. This is obviously more problematic with web-only, since you can't edit an already sent email. I think the layout disruption is less important compared to the potential for "missing" content. 
WDYT @thomasguillot ?

thomasguillot: This is really tricky because it can be extremely complex and it can easily look incredibly busy.

I was thinking, can we append the visibility status to the list view?

Screenshot 2021-10-13 at 09 33 00

miguelpeixe: This is an attempt to be as frictionless as possible. The all-around border is so it can visually cover grouped blocks and the label is `position:absolute;` so it doesn't interfere with different layouts.

https://user-images.githubusercontent.com/820752/137503284-a791714c-f5ad-4e55-a75e-04f2d125140d.mov

thomasguillot: > This is an attempt to be as frictionless as possible. [...]

This is quite nice. Are you using box-shadow? I can see the content slightly shifting when it doesn't have the yellow border.

src/editor/blocks/visibility-attribute/index.js
```diff
@@ -0,0 +1,172 @@
...
+							[ `newsletters-block-visibility__${ value }` ]: !! value,
+							'newsletters-block-selected': props.isSelected,
+							'newsletters-block-error': shouldBePublic,
+						} ) }
+						data-align={ props.attributes?.align || null }
+					>
+						<span className="newsletters-block-visibility-label">
+							<Icon icon={ warning } size={ 15 } />
+							{ sprintf(
+								__( 'Only visible on the "%s" version.', 'newspack-newsletters' ),
```
adekbadek, line 140: Not sure about this phrasing. How about two separate versions:

- "Only visible on public newsletter page."
- "Only visible in the sent email."

?

src/editor/blocks/visibility-attribute/index.js
```diff
@@ -0,0 +1,172 @@
...
+								value
+							) }
+							{ shouldBePublic
+								? ' ' + __( 'Newsletter is not public.', 'newspack-newsletters' )
```
adekbadek, line 144: I'd propose to display this message instead of the "Only visible…" message, and expand it a bit, e.g.:

> Newsletter is not public, but this block is set to be displayed on the public newsletter page only.

src/editor/blocks/visibility-attribute/index.js
thomasguillot: Nice ✌️

src/editor/blocks/visibility-attribute/index.js
adekbadek: That's true.

thomasguillot (approved):

PR #597, opened 2021-10-18
fix: change MC API key option name

kariae:
### All Submissions:

https://github.com/Automattic/newspack-plugin/pull/1168 introduced a new name for the MC API key
option, which will be managed from the `newspack-plugin` instead of here. We do keep the old name for
backwards compatibility.

1. Set your MC API Key from `Newsletters > Settings > Mailchimp API Key`
2. Create a new newsletter.
3. Connect to your Mailchimp account, and see that the newsletter created is there as a draft campaign, which means that the API is working.
4. Pull this PR, and repeat 2. and 3. to create a second newsletter. Observe the newly created campaign is in your Mailchimp account as a draft campaign, which means the backward compatibility is 👌 
5. Create a new API Key from your Mailchimp account and set it in `Newsletters > Settings > Mailchimp API Key`
6. Create a third newsletter by repeating 2. and 3.
7. Observe that the third newsletter is in your Mailchimp account as a draft campaign, which means that the new API Key name is working.

### Other information:


includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php
```diff
@@ -79,7 +81,9 @@ public function set_api_credentials( $credentials ) {
 			$ping = null;
 		}
 		return $ping ?
-			update_option( 'newspack_newsletters_mailchimp_api_key', $api_key ) :
+			// 'newspack_mailchimp_api_key' is a new option introduced to manage MC API key accross Newspack plugins.
+			// Keeping the old option for backwards compatibility.
```
adekbadek, line 85: Nit: this comment is irrelevant in this case, there's no legacy option name handling here.

includes/class-newspack-newsletters-settings.php
```diff
@@ -56,7 +56,7 @@ public static function get_settings_list() {
 			),
 			array(
 				'description' => __( 'Mailchimp API Key', 'newspack-newsletters' ),
-				'key'         => 'newspack_newsletters_mailchimp_api_key',
+				'key'         => 'newspack_mailchimp_api_key',
```
adekbadek, line 59: This is a little bit problematic, because a user will see empty value here when they do have the legacy option set, but not the new one. The `default` key could be set here to mitigate this:

```
'default'     => get_option( 'newspack_newsletters_mailchimp_api_key', '' ),
```

adekbadek (changes requested):

includes/class-newspack-newsletters-settings.php
kariae: Ah, you're right, I forgot the default value here 👍

adekbadek (approved):

PR #609, opened 2021-10-29
Fix/non public newsletters in search results new

### LINKED ISSUE #313: Public posts – newsletters appear in search results regardless of their public status
As is: When using the site's search, it's possible to get the newsletter CPT in the search results. This is regardless of whether they were marked as public in the editor. 

To be: The newsletters which were not marked as public should not appear in search results.

eddiesshop:
### All Submissions:

* [X] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [X] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [X] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

Newspack Newsletters were showing up in search results regardless of whether they were marked as Public or non-Public. This fix attempts to correct that behavior by ensuring that non-public posts do not show up in search results.

Closes #313 .

1. Create new newsletter.
2. Mark as sent. Do not make public.
3. Go to site search, and search for some content which is in the newsletter you created.
4. You should not see the newsletter in the search results.
5. Go back and make the newsletter public.
6. Go to site search again and search for some content which is in the newsletter.
7. You should see the newsletter in the search results.
8. Create a regular post and publish.
9. Search for content within that post.
10. You should not see that post in the search results.

*Negative testing*

1. create a post and publish.
2. For that post ID, add a new record to postmeta table with key = is_public and value = 1
3. Search for content within the post you created
4. The post should not be visible despite having the target content.
5. Create a post and publish
6. For that post ID, add a new record to postmeta table with key = is_public
7. Search for content within the post you created
8. The post should not be visible.

### Other information:

* [X] Have you added an explanation of what your changes do and why you'd like us to include them?
* [X] Have you successfully ran tests with your changes locally?


claudiulodro (approved):
The logic looks solid and it works well in my testing. Thanks for investigating and fixing!

dkoo:
Only thing I would suggest is that we as a team normally prefer the shorthand syntax for arrays: `$thing = [ 'array' ]` instead of `$thing = array( 'array' )`. Is it possible to update your IDE rules to use the shorthand syntax? 

Acknowledging that there may still be instances of `array()` throughout our code, but those are probably holdovers from early days.

PR #622, opened 2021-11-19
Fix/public newsletter queries

dkoo:
### All Submissions:

Ensure that public newsletters always appear alongside regular posts, and non-public newsletters never appear in any queries.

Test the following conditions:

- [x] Homepage Posts: should only display public newsletters if Newsletters are selected in Post Types
- [x] Blog page: should display both posts and public newsletters (set this page in Settings > Reading)
- [x] Category archives: should should display both posts and public newsletters
- [x] Tag archives: should should display both posts and public newsletters
- [x] Date archives: should should display both posts and public newsletters
- [x] Author archives: should should display both posts and public newsletters
- [x] Newsletter post type archives: should contain only public newsletters (default URL: https://<site_domain>/newsletter)
- [x] Other post type archives: should only contain the post type being queried (no newsletters or posts)
- [x] Search results: should contain all post types, but only public newsletters
- [x] Singular newsletter posts when logged in: should be visible regardless of public setting
- [x] Singular newsletter posts when not logged in: should only be visible if set as public

### Other information:


includes/class-newspack-newsletters.php
```diff
@@ -479,37 +479,44 @@ public static function should_apply_filter_to_query( $query ) {
...
-		$meta_query        = $query->get( 'meta_query', array() ); // phpcs:ignore WordPressVIPMinimum.Hooks.PreGetPosts.PreGetPosts
-		$meta_query_params = array(
-			array(
-				'key'     => 'is_public',
-				'value'   => true,
-				'compare' => '=',
-			),
-		);
+		$contains_newsletters   = 'any' === $post_types || self::NEWSPACK_NEWSLETTERS_CPT === $post_types || ( is_array( $post_types ) && in_array( self::NEWSPACK_NEWSLETTERS_CPT, $post_types ) );
+		$contains_regular_posts = empty( $post_types ) || 'any' === $post_types || 'post' === $post_types || ( is_array( $post_types ) && in_array( 'post', $post_types ) );
```
claudiulodro, line 496: I really like the way you structured this. It's very readable

claudiulodro (approved):
This looks solid and works well in my testing!

The only situation this approach would cause issues in that I see is if a regular post had `is_public` meta, which would be a weird situation and highly unlikely.

dkoo:
> The only situation this approach would cause issues in that I see is if a regular post had `is_public` meta, which would be a weird situation and highly unlikely.

@eddiesshop @claudiulodro 3524a0f restores the custom `join` and `where` clauses, which should help us avoid this situation, and also removes the meta query from the `pre_get_posts` query filtering, which is redundant if used with the custom clauses. Confirmed via functional and unit testing that all test cases outlined above still pass.

miguelpeixe:
> The CI build task is still failing due to this issue, but I think we should merge anyway if you can also confirm that npm ci && npm run build work locally.

It could be from recent image changes, but not exactly the same issue. Newsletters doesn't install any extra PHP extensions from PPA that would fail from the identified breaking change.

From CI logs looks like a `node-sass` with node v16 conflict. What is weird to me is that [this image has been using lts node for a while now](https://github.com/CircleCI-Public/cimg-php/blame/master/7.4/node/Dockerfile#L8-L10). 🤔

claudiulodro (approved):
Latest revision works well 👍 I like that you have tests also!

PR #640, opened 2021-12-07
feat: add newsletter post type to supported post types in Campaigns

dkoo:
### All Submissions:

Adds support for the newsletter post type to Newspack Campaigns. Requires the new filter applied in https://github.com/Automattic/newspack-popups/pull/726 to work.

h/t @kariae for helping me debug.

This doesn't necessarily fix a bug, but it does address some behavior that might be unexpected. Public newsletters are displayed in archives, alongside regular posts if they share the same categories, tags, or authors. Newspack Campaigns prompts can be set to display in post archives, but if the archive contains any post type other than `post` and `page`, those extra post types aren't used for determining whether to display the archive prompt. Adding support for newsletters to Campaigns by default seems a reasonable default behavior—editors can now _opt out_ of showing specific prompts on singular newsletters or archives that contain newsletters, instead of having to _opt in_ to showing specific prompts in these cases.

1. Publish a newsletter and assign it a category so that it shows up as the first post in that category's archive page.
2. In Newspack Campaigns, publish an archive prompt that should display after 1 post in category archives. Don't edit the Post Types options (but do observe that it shows newsletters, but it's not selected by default).
3. On `master`, observe that the prompt is not displayed on the category archive (because the first post shown is a newsletter, not a post).
4. Check out this branch, confirm that the archive prompt and any new prompt now have newsletters selected under Post Types by default.
5. Confirm that the category archive now shows the prompt after the first post.

### Other information:

### DIFF
--- a/includes/class-newspack-newsletters.php
+++ b/includes/class-newspack-newsletters.php
@@ -85,6 +85,7 @@ final class Newspack_Newsletters {
 		add_filter( 'newspack_theme_featured_image_post_types', [ __CLASS__, 'support_featured_image_options' ] );
 		add_filter( 'gform_force_hooks_js_output', [ __CLASS__, 'suppress_gravityforms_js_on_newsletters' ] );
 		add_filter( 'render_block', [ __CLASS__, 'remove_email_only_block' ], 10, 2 );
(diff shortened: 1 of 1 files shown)


dkoo:
No longer needed because of https://github.com/Automattic/newspack-popups/pull/726#issuecomment-990083616.

PR #733, opened 2022-02-08
Simplify visibility option

thomasguillot:
### All Submissions:

This PR simplifies the `ToolbarDropdownMenu` of the Visibility option.

* Remove the use of icons (except for the "check")
* Display the all the options if the Newsletter is public (default: "Email and Web")

1. Switch to this branch
2. Enable Newsletter Public
3. Play with visibility settings
4. Also check if you disable Newsletter Public, the settings stay

### Other information:


miguelpeixe:
The current strategy maintains the toolbar if there are any set values exclusively to handle the exception of a block set to web without the newsletter having a web version. Removing the toolbar for such case would require the user to reenable the web version just so they can update the attribute.

To tackle that, 34dcc54ff5e3180f0694558b86e679ad8fe424c3 adds a button shortcut to clear the visibility attribute when the block is not going to be displayed.

PR #743, opened 2022-02-14
Add all accepted keys to get_post_type_labels()

thomasguillot:
### All Submissions:

When registering the CPT we currently don't translate all the labels which means in the editor we might see things like "Post updated." or "Post published.". This PR fixes it 🙂

1. Go to the editor and send a new newsletter
2. Notice the snackbar
3. Switch to this branch
4. Repeat 1 and 2

### Other information:


dkoo (changes requested):
Good call on this! We're missing one keys according to the [docs](https://developer.wordpress.org/reference/functions/get_post_type_labels/), though:

* `view_items`

Guessing we don't need the following because the default strings apply to Newsletters, too—is that correct?

* `featured_image`
* `set_featured_image`
* `remove_featured_image`
* `use_featured_image`
* `filter_by_date`

thomasguillot:
> Good call on this! We're missing one key according to the [docs](https://developer.wordpress.org/reference/functions/get_post_type_labels/), though:

Oops fixed in https://github.com/Automattic/newspack-newsletters/pull/743/commits/48f71a76fe097860a6f1419df96e51ee0a4e4eb6

> Guessing we don't need the following because the default strings apply to Newsletters, too—is that correct?

Correct ☺️

dkoo (approved):
💯

PR #746, opened 2022-02-17
fix(mailchimp): check send status

### LINKED ISSUE #620: Changing post status causes Mailchimp error
The current implementation uses `post_status` changes as the strategy to trigger the "send newsletter" action on the ESP as well as use the `publish` state to identify if the newsletter has been sent.

Although it is a good assumption, it is a faulty relationship since WordPress allows the newsletter post status to change back to draft and republish. This action causes an error with Mailchimp because it attempts to resend an already sent campaign.

It would be best to not use the `post_status` for this control strategy and use a custom meta that the plugin would have sole and full control over. The custom meta should also include the revision that was sent to track possible after-send changes.

miguelpeixe:
### All Submissions:

Verify the campaign status before attempting to send a Mailchimp newsletter.

Closes #620 

1. On the master branch, create and send a new newsletter
2. Use the quick edit feature to edit the newsletter status to `Draft`, than back to `Publish`
3. Observe the error than delete this newsletter
4. Switch to this branch, create and send another newsletter
5. Repeat the second step and observe you are able to switch the status without attempting to send or errors

### Other information:

### DIFF
--- a/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php
+++ b/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php
@@ -536,6 +536,16 @@ final class Newspack_Newsletters_Mailchimp extends \Newspack_Newsletters_Service
 
 		if ( 'publish' === $new_status && 'publish' !== $old_status ) {
 			try {
+				// Check if campaign has already been sent and if so, don't attempt to
+				// send again.
+				$campaign_data = $this->retrieve( $post_id );
+				if (
+					isset( $campaign_data['campaign']['status'] ) &&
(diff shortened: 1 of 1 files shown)


dkoo (changes requested):
This works as described!

We may want to handle this in a separate PR if it'll take a while, but we should also implement the same fix for Campaign Monitor and Constant Contact.

miguelpeixe:
It'd be best to handle separately since each ESP provides a different API, thus a different way of handling the "sent" status that may require different approaches.

dkoo (approved):
Makes sense. This fix works well for Mailchimp, so 👍

PR #757, opened 2022-03-04
fix: color palette css rules

miguelpeixe:
### All Submissions:

The email editor strips all default editor styles, this also removes the `.has-{color-name}-color` used by the "Highlight" (`<mark />`) attribute used for inline coloring of texts.

image

1. While on the master branch, create a new newsletter and apply "highlight" to a selected text
2. Observe the background color works as expected but not the text color (background color uses inline CSS rules)
3. Check out this branch, apply the highlight text color and observe the changes are applied
4. Send the newsletter with a highlighted text and confirm it is also highlighted on the sent email

### Other information:

### DIFF
--- a/includes/class-newspack-newsletters-editor.php
+++ b/includes/class-newspack-newsletters-editor.php
@@ -114,7 +114,8 @@ final class Newspack_Newsletters_Editor {
 			}
 		}
 		if ( $container_selector ) {
-			$rules = array_map(
+			$container_selector = esc_html( $container_selector );
+			$rules              = array_map(
 				function( $rule ) use ( $container_selector ) {
 					return $container_selector . ' ' . $rule;
 				},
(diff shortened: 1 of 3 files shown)


dkoo (changes requested):
Hmm, this is strange because when I inspect the email HTML saved to the `newspack_email_html` value, I see the inline styles in the markup. But when I send a test email to myself, I'm not able to see the background or text color, and the `mark` element with the inline styles and color-based class name seem to be stripped.

miguelpeixe:
Good catch! Apparently `<mark />` is not a supported HTML tag for email clients and Mailchimp strips it. I've added a `preg_replace()` to use `<span />` instead. The styles should be preserved and produce the same desired visual effect.

I've also made a small change to `<figcaption />` parsing to add the `LIBXML_HTML_NOIMPLIED | LIBXML_HTML_NODEFDTD` flags. This allows for the output of `saveHTML()` to not include `<html><body/></html>` on its output. It shouldn't produce any visible change to the behavior of rendering image captions.

PR #764, opened 2022-03-15
feat: minify and notify clipable email body

miguelpeixe:
### All Submissions:

[Email content larger than 102kb gets clipped by Gmail](https://mailchimp.com/pt-br/help/gmail-is-clipping-my-email/). This PR implements a notice to let the user know their content may get clipped if we detect a resulting body larger than 100,000 bytes. This approximation is assuming that the ESP will add extra tags for tracking, footer, etc.

image

It also implements the minifying options to MJML to reduce the chance of clipping. My local tests produced a reduction of 25% of an email with ~100k bytes.

1. Check out this PR and run `npm run build`
2. Send an email using multiple different blocks to make sure the markup is not altered by the minification
3. Add multiple text blocks in a newsletter draft until it reaches > 100k bytes
4. Save the draft and confirm you see the notice in the screenshot above

### Other information:

### DIFF
--- a/src/newsletter-editor/editor/index.js
+++ b/src/newsletter-editor/editor/index.js
@@ -6,6 +6,7 @@ import { get, isEmpty } from 'lodash';
 /**
  * WordPress dependencies
  */
+import { __ } from '@wordpress/i18n';
 import { compose } from '@wordpress/compose';
 import { withDispatch, withSelect } from '@wordpress/data';
 import { createPortal, useEffect, useState } from '@wordpress/element';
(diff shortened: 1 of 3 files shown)


src/newsletter-editor/editor/index.js
```diff
@@ -133,6 +142,23 @@ const Editor = compose( [
 		}
 	}, [ props.status ] );
 
+	// Notify if email content is larger than ~100kb.
+	useEffect( () => {
+		const noticeId = 'newspack-newsletters-email-content-too-large';
+		const message = __(
+			'Email content is too large and may get clipped by email clients.',
```
dkoo, line 149: Nit: "long" sounds to me like it better suggests what action the editor should take to fix this error.

```suggestion
			'Email content is too long and may get clipped by email clients.',
```

miguelpeixe:
> except the new WP 5.9 "Row" block seems to not render correctly

Nice catch! I just pushed #772 for this.

> Inspecting the meta value, it does appear to be minified, but I wonder if the above indicates support for this feature might go away at some point?

I didn't notice that warning, but [unfortunately yes](https://github.com/mjmlio/mjml/pull/2204) ☹️ 
We could use the minify option until the next stable major release is out and until then figure out if we'll use a different library for the task. WDYT?

dkoo (approved):
>  We could use the minify option until the next stable major release is out and until then figure out if we'll use a different library for the task. We can also not minify at all and just use the notice. WDYT?

This sounds good to me, as long as continuing to use the `minify` option doesn't end up causing an error on the MJML side. I'm assuming after the option is removed from the library behavior will revert back to what's happening on the current `master` HEAD, right? Basically the meta value will be saved as unminified HTML again?

dkoo:
That matches my understanding, too—thanks!

PR #772, opened 2022-03-22
fix: disable row group block variation

miguelpeixe:
### All Submissions:

[As pointed out](https://github.com/Automattic/newspack-newsletters/pull/764#pullrequestreview-916140528) by @dkoo, the "Row" block, which is a Group block variation, is not supported by our renderer. Instead of a row, the result is a regular group.

This PR unregisters the block variation.

1. While on the master branch, send a newsletter using the "Row" block
2. Confirm the unexpected layout
3. Check out this branch and confirm you are no longer able to use the "Row" block
4.

### Other information:

### DIFF
--- a/src/editor/index.js
+++ b/src/editor/index.js
@@ -1,7 +1,7 @@
 /**
  * WordPress dependencies
  */
-import { unregisterBlockStyle } from '@wordpress/blocks';
+import { unregisterBlockStyle, unregisterBlockVariation } from '@wordpress/blocks';
 import domReady from '@wordpress/dom-ready';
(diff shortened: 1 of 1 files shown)


dkoo (approved):
This works as described! 👍 

Interestingly enough, if you've already previously added a Row block to a newsletter before this change, it will continue to behave as a Row block in the block editor (with alignment/justification controls) but be named as a "Group" in the block list. New Group blocks will lack the Row alignment/justification controls. Probably not an issue, just wanted to note the interesting fallback behavior.

PR #786, opened 2022-04-12
fix(layouts): ensure selected layout is saved

miguelpeixe:
### All Submissions:

`savePost()` is intended to be called after the selection of the layout ID but instead it's assuming the `editPost()` action will be fulfilled after 1 ms, which is not good practice.

This PR updates the `savePost()` action to be executed as a callback to the `setNewsletterMeta()`.

Confirm layout selection and draft saving behaves as expected and the selected layout is saved (the layout picker is not displayed again after refreshing the editor).

### Other information:


dkoo (approved):
Nice catch—it's hard to replicate the original issue, but the promise seems to work great. 👍

PR #796, opened 2022-04-20
Reorganise layouts and add header group to all of them

thomasguillot:
### All Submissions:

* Reorganisation of the layouts and removal of the titles (still available for saved layouts though!)
* All layouts now are sharing more/less the same header that extracts the color from Newspack theme
* Adjusted some of the blocks

1. Switch to this branch
2. Test all the different layouts

### Other information:


dkoo (approved):
New layouts look good to me.

Approving, because this doesn't relate to this PR. But one thing I noticed that's also happening on `master` is that the "Support our newsroom" blocks don't render well when viewed on the site as a public post. Looks like it loses the alignment, margins and button colors:

Screen Shot 2022-04-20 at 4 40 19 PM

Is there a way we can ensure that the styles match a bit more closely or are at least readable on the site?

thomasguillot:
@dkoo Hmmm this is a problem... In the meantime, I might just add `"newsletterVisibility":"email"` to the `Group`. The donation footer is not really need on the website.

PR #798, opened 2022-04-20
refactor: sending newsletters on status changes

### LINKED ISSUE #785: Disable "Move to trash" on sent newsletters
If a sent newsletter is inadvertently moved to the trash, it's not possible to republish it without attempting to send again:

Annotation on 2022-04-11 at 08-37-23

There should never be a need to delete a sent newsletter. The option to toggle the public-facing version off is sufficient for removing it from the site. So let's remove the option to trash a sent newsletter.

miguelpeixe:
### All Submissions:

This PR refactors how the plugin handles sending Newsletters on status changes by moving the hook to the abstract class, avoiding duplicate logic for improved readability and maintenance. It also simplifies each provider's `send()` method to only perform the send action and return `true|WP_Error`  to be handled by the `send_newsletter()` abstract class method.

The hook was changed from `post_status_changes` to `pre_post_update` so it can prevent the status change in case of a failed send. The current method requires a `wp_update_post()` to revert the post back to draft, which is not ideal.

The new `pre_post_update()` hooked method also verifies if a sent newsletter is changing status after being sent and prevents such action - unless it's changing to/from trash, which it should always allow.

With the plan of detaching the post status value to the newsletter "sent" status in mind, this PR also implements the `newsletter_sent` meta. It's populated with a timestamp in case the final provider `send()` method succeeds but also has backward compatibility, being populated with the `post_date` if the newsletter is `publish`.

By using the meta, changing a trashed sent newsletter to publish will not trigger a send attempt. A hook also ensures that restoring a trashed sent newsletter automatically moves it to publish instead of the default "draft" (closes #785)

1. Send a newsletter using each provider and confirm you receive the email and it's marked as sent as expected. (Constant Contact is currently unavailable due to #802)
2. Using the "Quick edit", attempt to modify a sent newsletter status to "Draft" and confirm you see an appropriate error message
3. Confirm you are able to move newsletters to the trash (sent or drafts)
4. Restore a trashed sent newsletter and confirm its status is `publish`
5. Restore a trashed draft newsletter and confirm its status is `draft`

### Other information:

### DIFF
--- a/includes/service-providers/campaign_monitor/class-newspack-newsletters-campaign-monitor.php
+++ b/includes/service-providers/campaign_monitor/class-newspack-newsletters-campaign-monitor.php
@@ -352,7 +352,6 @@ final class Newspack_Newsletters_Campaign_Monitor extends \Newspack_Newsletters_
 
 		$cm_campaigns = new CS_REST_Campaigns( null, [ 'api_key' => $api_key ] );
 		$args         = $this->format_campaign_args( $post_id );
-		$send_date    = 'future' === $new_status ? $post->post_date : 'Immediately';
 
 		// Set the current user's email address as the email to receive sent confirmation.
 		$current_user       = wp_get_current_user();
@@ -364,7 +363,7 @@ final class Newspack_Newsletters_Campaign_Monitor extends \Newspack_Newsletters_
 		if ( ! $new_campaign->was_successful() ) {
 			return new WP_Error(
 				'newspack_newsletter_error',
-				__( 'Failed sending Campaign Monitor campaign: ', 'newspack-newsletters' ) . $new_campaign->response->Message
+				__( 'Failed creating Campaign Monitor campaign: ', 'newspack-newsletters' ) . $new_campaign->response->Message
 			);
 		}
 
@@ -374,7 +373,7 @@ final class Newspack_Newsletters_Campaign_Monitor extends \Newspack_Newsletters_
 			$campaign_to_send->send(
 				[
 					'ConfirmationEmail' => $confirmation_email,
-					'SendDate'          => $send_date,
+					'SendDate'          => 'Immediately',
 				]
 			);
 		} catch ( Exception $e ) {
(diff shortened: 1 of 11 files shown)


includes/service-providers/campaign_monitor/class-newspack-newsletters-campaign-monitor.php
```diff
@@ -328,75 +327,64 @@ public function content( $post_id ) {
...
-				$cm_campaigns = new CS_REST_Campaigns( null, [ 'api_key' => $api_key ] );
-				$args         = $this->format_campaign_args( $post_id );
-				$send_date    = 'future' === $new_status ? $post->post_date : 'Immediately';
-
-				// Set the current user's email address as the email to receive sent confirmation.
-				$current_user       = wp_get_current_user();
-				$confirmation_email = $current_user->user_email;
+		$cm_campaigns = new CS_REST_Campaigns( null, [ 'api_key' => $api_key ] );
+		$args         = $this->format_campaign_args( $post_id );
+		$send_date    = 'future' === $new_status ? $post->post_date : 'Immediately';
```
adekbadek, line 355: `$new_status` is not defined anymore.

adekbadek (changes requested):
Nice! This simplifies things a lot. 

Nit: since the meta is a timestamp, I'd expect it to be named rather `newsletter_sent_at` and the related variables also along these lines (e.g. `$sent_at`).

includes/service-providers/campaign_monitor/class-newspack-newsletters-campaign-monitor.php
miguelpeixe: Good catch! a0936876f1dc807d77a93cc37842f396747b660a updates the method to always send immediately. We don't have to manage scheduling if the method is only called when changing from draft to publish, which can be scheduled through WordPress.

miguelpeixe:
A few other updates to this PR:

91097fd08ef538c599da1d85ac1696728df47ccd fixes a bug to properly construct `DateTime()` when using a timestamp.

8fa527d2900367e3449341c4a2fbdd06cda57525 updates the middleware to let core (`next()`) catch potential errors, there's no need for our interception there. You can test the error handling by adding:

```php
wp_die( new WP_Error( 'newspack_newsletters_error', 'A test error' ) );
```

to line `105` of `/includes/service-providers/class-newspack-newsletters-service-provider.php`

adekbadek (approved):
👐

PR #801, opened 2022-04-22
Add another layout and add scroll effect to layout picker

thomasguillot:
### All Submissions:

Added an 8th layout plus a tiny scroll effect when selecting a layout

1. Switch to this branch
2. Do you see the effect when hovering?
3. Select Layout 8 and publish/test

### Other information:


src/components/init-modal/style.scss
```diff
@@ -318,12 +318,22 @@
...
+
 				.block-editor-block-preview__container,
 				.block-editor-block-preview__content-iframe {
 					cursor: pointer;
 					position: absolute;
 					top: 0;
 				}
+
+				.block-editor-block-preview__container {
+					transition: margin 1s ease-in-out;
```
dkoo, line 335: Non-blocking: Personal preference, maybe, but I think it looks better with a 0.5s duration—WDYT?

src/components/init-modal/style.scss
```diff
@@ -318,12 +318,22 @@
 					border-color: var( --wp-admin-theme-color );
 				}
 
+				&:hover {
+					.block-editor-block-preview__container {
+						margin-top: -25%;
```
dkoo, line 323: Non-blocking: Is the intention to show more of the layout as you're hovering over the thumbnail? If so, maybe we should bump this to 50% or even 100%?

dkoo (approved):
Works as described. Approving because the comments below are personal preference rather than issues.

src/components/init-modal/style.scss
thomasguillot: It is, yes. The reason I used 25% is because the "Support" layout is very short and if I increase the margin then it looks half empty when I'm hovering.

src/components/init-modal/style.scss
thomasguillot: Good idea. https://github.com/Automattic/newspack-newsletters/pull/801/commits/1d780e373e3a00a74ee3363b9c0d961ee35dfdd1

PR #831, opened 2022-05-25
fix: sending scheduled newsletters

miguelpeixe:
### All Submissions:

The refactor from #798 was missing the integration with scheduled posts. WP scheduling uses `wp_publish_post()`, which does not fire `pre_post_update` or any other hook before updating the database 😢 

This requires us to use a hook after the post is published (`transition_post_status`) to handle sending scheduled newsletters.

This PR implements that case and also updates some phrasing on the UI. The sending confirmation modal for scheduled newsletters should state "Schedule newsletter" instead of "Send".

The error handling of scheduled newsletters should later be adjusted, along with the other error transients, to have a better UI integration. Currently it only serves to allow the failed send to return to draft after publishing. It should properly display the error message regardless of current user ID or a time frame, which is how the transients are currently implemented.

1. Check out this branch and set up a valid ESP
2. Start a new newsletter and click to send without scheduling
5. Confirm the modal have appropriate messaging and close
4. Schedule the newsletter for a couple of minutes from now and click to schedule sending
5. Confirm the modal have appropriate messaging and confirm the scheduled sending
7. After the scheduled time passes, confirm the newsletter is sent and marked as sent on your dashboard
8. Simulate an error by adding `return new WP_Error( 'newspack_newsletters_error', 'Temporary error' );` to line 201 of `includes/service-providers/class-newspack-newsletters-service-provider.php`
9. Create a new newsletter, schedule the send and after the time passes confirm the newsletter is back to draft

### Other information:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## [1.46.1-hotfix.1](https://github.com/Automattic/newspack-newsletters/compare/v1.46.0...v1.46.1-hotfix.1) (2022-05-25)
+
+
+### Bug Fixes
+
+* sending scheduled newsletters ([2a25225](https://github.com/Automattic/newspack-newsletters/commit/2a25225f79e032ae454fe3e47fd40708f3b8c6e9))
+
 # [1.46.0](https://github.com/Automattic/newspack-newsletters/compare/v1.45.0...v1.46.0) (2022-05-18)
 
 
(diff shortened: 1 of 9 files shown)


dkoo (approved):
This works as described! Approving.

One thing I don't like (which is also true in `master`, so perhaps the subject of a future issue) is that after scheduling a post, the "Send" button changes to say "Scheduled." But it's not clear to me what should happen when you click that. What does happen is that it prompts you to schedule the send again, which I suppose makes sense if you were to make changes before it gets sent and want to ensure those changes get sent at the scheduled time.

miguelpeixe:
Thanks, @dkoo!

> But it's not clear to me what should happen when you click that. 

I agree! Let's file an issue for this.

PR #835, opened 2022-05-25
fix: ensure controlled statuses after a post update

miguelpeixe:
### All Submissions:

#831 recently released a hotfix for scheduled newsletters that need to be adjusted to our new "controlled statuses" strategy, implemented by #824.

The `wp_publish_post()` function used by WP scheduling does not go through the `wp_post_insert_data` hook either, which would adjust the status before the update. For that reason, we need to add an extra check to adjust the post status after an update.

Repeat steps from #831, confirming a scheduled send is set to private when not enabling "Public newsletter".

### Other information:


dkoo (approved):
Confirmed that scheduled newsletters get set to `private` when the "public" option is disabled, and `publish` when enabled.

PR #836, opened 2022-05-26
fix: ensure scheduled newsletter is sent through a lock meta

miguelpeixe:
### All Submissions:

Sending scheduled newsletters is still not working appropriately due to how we approach the `is_newsletter_sent()` method. For legacy reasons, it has to assume that `publish` newsletters are sent. By the time `send_newsletter()` checks that on scheduled newsletters, WP scheduling has already called `wp_publish_post()`, so our check method will assume the newsletter is sent and bail.

This PR implements a lock mechanism through a disposable meta, just to guarantee that the scheduled sending window goes through the legacy check.

Repeat steps from #831 and confirm the newsletter is sent from the ESP dashboard.

### Other information:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## [1.46.2-hotfix.1](https://github.com/Automattic/newspack-newsletters/compare/v1.46.1...v1.46.2-hotfix.1) (2022-05-26)
+
+
+### Bug Fixes
+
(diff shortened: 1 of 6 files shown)


dkoo (approved):
This works for me—confirmed that a scheduled newsletter remains in scheduled status until the scheduled time, then transitions to `publish` and triggers the ESP send at that point.

Noting for future reference that @miguelpeixe and I discovered a strange bug (maybe from WP core?) in which a post scheduled to send very soon in the future (~2 minutes or less) sometimes gets sent immediately.

PR #892, opened 2022-07-19
feat(subscribe-block): prefill email input

adekbadek:
### All Submissions:

Pre-filling of email input of the Subscribe block, if an email is available. 

cc @miguelpeixe in case that was something you had planned :) 

1. Insert a Subscribe block
2. As a logged in user, observe the email address is pre-filled
3. In a new session, use the Reader Registration block from the Newspack plugin to pre-authenticated as an existing reader
4. Observe the email is pre-filled, despite user not being logged in

### Other information:

### DIFF
--- a/src/blocks/subscribe/index.php
+++ b/src/blocks/subscribe/index.php
@@ -69,6 +69,18 @@ function render_block( $attrs ) {
 		$available_lists = [ $lists[0] ];
 	}
 
+	if ( \is_user_logged_in() ) {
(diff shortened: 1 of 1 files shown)


miguelpeixe (approved):
Nice!

The `\Newspack\Reader_Activation::get_auth_intention_value()` value may not always return due to page caching. The alternative would be to use the `ReaderActivation.getReaderEmail()` from the frontend library, but this can be enhanced on another PR.

PR #895, opened 2022-07-21
feat: add contact on reader activation

adekbadek:
Handle reader activation via hook. See https://github.com/Automattic/newspack-plugin/pull/1781#discussion_r924733952

### DIFF
--- a/includes/class-newspack-newsletters-subscription.php
+++ b/includes/class-newspack-newsletters-subscription.php
@@ -19,6 +19,7 @@ class Newspack_Newsletters_Subscription {
 	 */
 	public static function init() {
 		add_action( 'rest_api_init', [ __CLASS__, 'register_api_endpoints' ] );
+		add_action( 'newspack_registered_reader', [ __CLASS__, 'newspack_registered_reader' ], 10, 5 );
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


includes/class-newspack-newsletters-subscription.php
```diff
@@ -301,5 +302,30 @@ public static function add_contact( $contact, $lists = [] ) {
...
+	 * @param bool           $authenticate  Whether to authenticate after registering.
+	 * @param false|int      $user_id       The created user id.
+	 * @param false|\WP_User $existing_user The existing user object.
+	 * @param false|any      $metadata Metadata.
+	 */
+	public static function newspack_registered_reader( $email, $authenticate, $user_id, $existing_user, $metadata ) {
+		if ( $existing_user ) {
+			// Only sign up if they've just created an account.
+			return;
+		}
```
miguelpeixe, line 319: We shouldn't mind if it's an existing user. Reader registration is optimistic and the subscription is upserted to that email, so we would want that to happen.

miguelpeixe (changes requested):

includes/class-newspack-newsletters-subscription.php
```diff
@@ -301,5 +302,30 @@ public static function add_contact( $contact, $lists = [] ) {
...
 	}
+
+	/**
+	 * Handle Newspack's reader registration – add contact to list.
+	 *
+	 * @param string         $email         Email address.
+	 * @param bool           $authenticate  Whether to authenticate after registering.
+	 * @param false|int      $user_id       The created user id.
+	 * @param false|\WP_User $existing_user The existing user object.
+	 * @param false|any      $metadata Metadata.
```
miguelpeixe, line 313:
```suggestion
	 * @param array          $metadata      Metadata.
```

Type and linting

includes/class-newspack-newsletters-subscription.php
adekbadek: Done in 9bedd8aa04daf81c30156878b0f97ae9a11746bb. So it looks like the `add_contact` method should be called `upsert_contact`.

miguelpeixe (approved):

PR #900, opened 2022-07-29
feat: enable contact adding w/out lists, as opposed to w/ empty lists

adekbadek:
### All Submissions:

Enables signalling that the contact should be upserted without any change in list selection. Before, the list selection was an empty array by default, implicitly meaning that the upserted contact should not be subscribed to any list in the absence of this parameter. 

With this change, when an empty array of lists is supplied, it means the contact is unsubscribed from any lists. When a `false` value of the list ids parameter means that the contact has no change in lists selection.

_Follow instructions in https://github.com/Automattic/newspack-plugin/pull/1816_

### Other information:

### DIFF
--- a/includes/class-newspack-newsletters-subscription.php
+++ b/includes/class-newspack-newsletters-subscription.php
@@ -308,19 +308,19 @@ class Newspack_Newsletters_Subscription {
 	/**
 	 * Upserts a contact to lists.
 	 *
-	 * @param array    $contact {
-	 *    Contact information.
+	 * @param array          $contact {
+	 *          Contact information.
 	 *
 	 *    @type string   $email    Contact email address.
 	 *    @type string   $name     Contact name. Optional.
 	 *    @type string[] $metadata Contact additional metadata. Optional.
 	 * }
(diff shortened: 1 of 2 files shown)


includes/class-newspack-newsletters-subscription.php
```diff
@@ -377,7 +377,7 @@ public static function add_contact( $contact, $lists = [] ) {
 		 *    @type string   $name     Contact name. Optional.
 		 *    @type string[] $metadata Contact additional metadata. Optional.
 		 * }
-		 * @param string[]      $lists    Array of list IDs to subscribe the contact to.
+		 * @param string[]|false      $lists    Array of list IDs to subscribe the contact to.
 		 * @param bool|WP_Error $result   True if the contact was added or error if failed.
```
miguelpeixe, line 381: Nit: lint the inline doc spacing

miguelpeixe (approved):

includes/class-newspack-newsletters-subscription.php
adekbadek: Hm, this was not reported by the linter. But adjusted in 202ceb7e26e53d12a281da60bfe570f224c015a6

PR #907, opened 2022-08-01
fix: updating a list member

adekbadek:
### All Submissions:

Fixes a bug. `PATCH` will only update, but `PUT` [will update or create](https://mailchimp.com/developer/marketing/api/list-merges/).

1. Set up MC w/ two audiences
2. Using Reader Activation, register and subscribe to both lists
3. On `master`, the reader will be subscribed to only the first list, on this branch to both

### Other information:

### DIFF
--- a/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php
+++ b/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php
@@ -862,7 +862,7 @@ final class Newspack_Newsletters_Mailchimp extends \Newspack_Newsletters_Service
 				$result = $mc->post( "lists/$list_id/members", $update_payload );
 			} else {
 				$member_id = $found_subscribers[0]['id'];
-				$result    = $mc->patch( "lists/$list_id/members/$member_id", $update_payload );
+				$result    = $mc->put( "lists/$list_id/members/$member_id", $update_payload );
 			}
 			if (
 				! $result ||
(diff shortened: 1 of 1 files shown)


miguelpeixe (changes requested):
We should either use the `get_contact_data()` method to appropriately identify which lists the contact exists or always upsert regardless of "found subscribers".

The current fix can be misleading for future debugging.

I believe using our contact data method with our own upsertion strategy (instead of `put`) is better because it gives us more control over how it's done. See `update_contact_lists()` for patching subscription status.

adekbadek:
@miguelpeixe  – done in 43236fa660b750a2edf6898713caee18c1db5cdf. I think using `PUT` here is fine – `add_contact` should just upsert, it does not need to be concerned about the payload (which lists to assign/unassign the contact to/from).

miguelpeixe (approved):
If we're using `PUT`, we shouldn't care for `$existing_contact` to use `POST`. The current detection of existing contact only tells that the email exists in a list. Each list is its own database. Although it works, we're misusing the logic here.

I'm approving because it's non-blocking, but I think we should use a more strict logic to avoid confusion in the future.

adekbadek:
> If we're using PUT, we shouldn't care for $existing_contact to use POST. […] I think we should use a more strict logic  […] Either by using only PUT […]

That's not possible – the `PUT` call requires a contact identifier. It's an upsertion of contact-list pair, not of a contact, so to speak.

> Each list is its own database.

I wouldn't describe it so in the context of these two API calls here. There's a single DB of contacts, each of which can be assigned to list(s) – that would be my description. The logic fixed by this PR deals with either inserting (`POST`) or updating (`PUT`) a contact. The structure of the API endpoints suggests something else, but that's effectively what it does.

PR #914, opened 2022-08-04
fix(active-campaign): format date fields

miguelpeixe:
### All Submissions:

Implement proper data type for date fields

In order to test it, I created a new custom field called `NP_Test Date` to not disrupt the current sandbox. Existing fields cannot have their type changed, so we'll need to delete them in order to implement the new format.

1. Checkout this branch and https://github.com/Automattic/newspack-plugin/pull/1855
2. Either create a new dummy field at https://github.com/Automattic/newspack-plugin/pull/1855 or delete the existing fields so they can be recreated
3. Make an anonymous donation with Stripe
4. Confirm the created contact has the proper date values for `NP_Registration Date`, `NP_Last Payment Date`, `NP_Next Payment Date`
5. Edit the value and confirm it triggers a date picker with calendar

### Other information:

### DIFF
--- a/includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign.php
+++ b/includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign.php
@@ -622,6 +622,34 @@ final class Newspack_Newsletters_Active_Campaign extends \Newspack_Newsletters_S
 	}
 
 	/**
+	 * Get data type ID for a given field.
+	 *
+	 * Possible values:
+	 *  1 = Text Field,
+	 *  2 = Text Box (textarea),
+	 *  3 = Checkbox,
+	 *  4 = Radio,
+	 *  5 = Dropdown,
+	 *  6 = Hidden field,
+	 *  7 = List Box,
+	 *  9 = Date
+	 *
+	 * @param string $field_name The field name.
+	 *
+	 * @return int Data type ID.
+	 */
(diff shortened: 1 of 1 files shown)


includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign.php
```diff
@@ -621,6 +621,34 @@ public function trash( $post_id ) {
...
+	 *  6 = Hidden field,
+	 *  7 = List Box,
+	 *  9 = Date
+	 *
+	 * @param string $field_name The field name.
+	 *
+	 * @return int Data type ID.
+	 */
+	private static function get_metadata_type( $field_name ) {
+		switch ( $field_name ) {
```
adekbadek, line 642: There will be more date fields, and they'll all end with `Date`. WDYT about a regex check (`/NP_.*Date$/`) instead of a `switch`?

miguelpeixe: I think it's a fragile assumption to put in code

adekbadek (approved):

includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign.php
adekbadek: It's an assumption we can make if we set the field names. But there's also readability value in an explicit list.

PR #924, opened 2022-08-09
feat: send verification email upon new account registration

dkoo:
### All Submissions:

Along with https://github.com/Automattic/newspack-plugin/pull/1876, sends a verification email instead of the standard WooCommerce "welcome" email to readers who register a new account via direct email input.

Will not send a verification email to readers who register via SSO, which is considered a secure login method and doesn't require verification.

Closes https://github.com/Automattic/newspack-plugin/issues/1872.

1. Check out this branch and https://github.com/Automattic/newspack-plugin/pull/1876.
2. Register a new email address via the Register block (via direct email input, not SSO) and confirm that you receive a verification email (with slightly tweaked language) instead of WooCommerce's "Your Newspack account has been created!" email.
3. Visit the verification URL in the same session where you registered the email address and confirm that you're redirected to the main My Account dashboard page with a message that your email has been verified. Also confirm that you're able to manage newsletter subscriptions at this point.
4. In a new session, register a new account via SSO and confirm that you do not receive a verification email, and that you're immediately able to manage newsletter subscriptions in My Account > Newsletters.

### Other information:

### DIFF
--- a/includes/class-newspack-newsletters-subscription.php
+++ b/includes/class-newspack-newsletters-subscription.php
@@ -12,7 +12,12 @@ defined( 'ABSPATH' ) || exit;
  */
 class Newspack_Newsletters_Subscription {
 
-	const API_NAMESPACE       = 'newspack-newsletters/v1';
+	const API_NAMESPACE = 'newspack-newsletters/v1';
+
+	const EMAIL_VERIFIED_META    = 'newspack_newsletters_email_verified';
+	const EMAIL_VERIFIED_REQUEST = 'newspack_newsletters_email_verification_request';
+	const EMAIL_VERIFIED_CONFIRM = 'newspack_newsletters_email_verification';
+
 	const WC_ENDPOINT         = 'newsletters';
 	const SUBSCRIPTION_UPDATE = 'newspack_newsletters_subscription';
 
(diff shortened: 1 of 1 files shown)


miguelpeixe:
If we're disabling WC welcome email and replacing it with our own RAS email, I think this should be considered a core feature and be included in `newspack-plugin`, not newsletters. WDYT?

dkoo:
> If we're disabling WC welcome email and replacing it with our own RAS email, I think this should be considered a core feature and be included in `newspack-plugin`, not newsletters. WDYT?

I thought the same, but opted for the smaller changeset at first. 😄 I'll refactor to move most of this functionality to the main plugin.

miguelpeixe:
Considering that newsletters picks up the `newspack_reader_verified` action hook, we don't need to use newsletters feature for this. Sending a RAS' magic link should be enough for verification as well.

miguelpeixe:
We must still preserve newsletters' agnostic verification for 2 reasons:

 - It should be able to work stand-alone
 - It per-email basis. Once RAS start supporting email address updates newsletters must still have its own verification strategy to not allow an email change access subscriptions

dkoo:
afbae2d moves all email verification methods to the main plugin.

> Considering that newsletters picks up the `newspack_reader_verified` action hook, we don't need to use newsletters feature for this. Sending a RAS' magic link should be enough for verification as well.

I like the idea of keeping a separate verification flow so that we can use different messaging in each case. As a reader, it doesn't make a ton of sense to me to receive a magic "authentication" link after registering a new account and being authenticated on the site.

miguelpeixe:
> As a reader, it doesn't make a ton of sense to me to receive a magic "authentication" link after registering a new account and being authenticated on the site.

The magic link is just the URL strategy for verifying. The email message is what tells the reader what the link does.

dkoo:
I think I know what to do here, then.

dkoo:
Closing this PR as it's no longer needed—we'll use magic links instead.

PR #932, opened 2022-08-15
feat: honeypot field for Subscribe block form

dkoo:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Similar to https://github.com/Automattic/newspack-plugin/pull/1896, implements a honeypot trap field for the Subscribe block's form. This is implemented fully in this plugin so that it has no dependencies on the main plugin.

### How to test the changes in this Pull Request:

1. Check out this branch. Confirm that the Subscribe block looks and functions as normal.
2. Temporarily comment out or delete [this SCSS block](https://github.com/Automattic/newspack-newsletters/pull/932/files#diff-bb09ed8cb7e0f6604c1d72c36e6bfb866010cc2e5d8dc319c3c1c8238af7e6f3R58-R69) to unhide the honeypot field.
3. Refresh the Subscribe block and confirm there's a new email input field alongside the usual one.
4. Rebusmit the form, this time filling out both email inputs, and confirm that the form resolves with a success message.
5. Confirm that no subscription event happens with the ESP on the back-end.

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [ ] Have you successfully ran tests with your changes locally?

PR #935, opened 2022-08-17
feat: handle reCAPTCHA on Subscribe form

dkoo:
### All Submissions:

Adds reCAPTCHA to Subscribe form submissions as implemented in https://github.com/Automattic/newspack-plugin/pull/1910, if available.

Also fixes a bug from #932 where the honeypot field is prefilled with an email address if one is known, which causes the form to always trigger the honeypot trap.

See testing instructions in https://github.com/Automattic/newspack-plugin/pull/1910.

### Other information:

### DIFF
--- a/src/blocks/subscribe/index.php
+++ b/src/blocks/subscribe/index.php
@@ -36,10 +36,10 @@ function enqueue_scripts() {
 		filemtime( NEWSPACK_NEWSLETTERS_PLUGIN_FILE . 'dist/subscribeBlock.css' )
 	);
 
-	$use_captcha  = method_exists( '\Newspack\Reader_Activation', 'can_use_captcha' ) && \Newspack\Reader_Activation::can_use_captcha();
+	$use_captcha  = method_exists( '\Newspack\Recaptcha', 'can_use_captcha' ) && \Newspack\Recaptcha::can_use_captcha();
 	$dependencies = [ 'wp-polyfill', 'wp-i18n' ];
 	if ( $use_captcha ) {
-		$dependencies[] = \Newspack\Reader_Activation::RECAPTCHA_SCRIPT_HANDLE;
+		$dependencies[] = \Newspack\Recaptcha::RECAPTCHA_SCRIPT_HANDLE;
 	}
 
 	\wp_enqueue_script(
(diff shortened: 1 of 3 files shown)


src/blocks/subscribe/index.php
```diff
@@ -241,6 +248,15 @@ function process_form() {
 		return send_form_response( [ 'email' => \sanitize_email( $_REQUEST['email'] ) ] );
 	}
 
+	// reCAPTCHA test.
+	if ( ! empty( $_REQUEST['captcha_token'] ) && method_exists( '\Newspack\Reader_Activation', 'verify_captcha' ) ) {
```
miguelpeixe, line 252: This check allows for the captcha test to be skipped if not sent. It must fail if it's enabled and the request doesn't provide a token.

miguelpeixe (changes requested):
As mentioned in https://github.com/Automattic/newspack-plugin/pull/1910, I'm getting a javascript error on the block `view.js`:

```
Uncaught ReferenceError: regeneratorRuntime is not defined
```

Probably related to webpack being unable to properly compile the declared dependencies?

src/blocks/subscribe/index.php
dkoo: Updated in 37f658

miguelpeixe (changes requested):
I'm still having issues with the compiled `view.js`. It looks related to the polyfill and the usage of `async`:

```
Uncaught ReferenceError: regeneratorRuntime is not defined
```

I'm not sure how to resolve it dependency-wise. I got it working locally by refactoring to use `Promise` only and not `async/await`.

dkoo:
[03b142f](https://github.com/Automattic/newspack-newsletters/pull/935/commits/03b142f4fd97e0937e546bfbedc3fbc3f623f9d0) replaces `async`/`await` with promises only.

src/blocks/subscribe/view.js
```diff
@@ -13,37 +15,86 @@ import './style.scss';
...
+						return res( '' );
+					}
+
+					const captchaSiteKey = reCaptchaScript.getAttribute( 'src' ).split( '?render=' ).pop();
+					if ( ! grecaptcha?.ready || ! captchaSiteKey ) {
+						rej( __( 'Error loading the reCaptcha library.', 'newspack-newsletters' ) );
+					}
+
+					grecaptcha.ready( async () => {
+						const token = await grecaptcha
```
miguelpeixe, line 36: There's still `async/await` here

miguelpeixe (changes requested):

src/blocks/subscribe/view.js
dkoo: Bah, sorry for the back and forth. b75f03a should fix it once and for all.

miguelpeixe (approved):
🙌

PR #948, opened 2022-09-05
fix(constant-contact): update api for oauth2

### LINKED ISSUE #802: Constant Contact API no longer supported
Constant Contact API has changed its authorization method and deprecated the method we currently rely on.

https://developer.constantcontact.com/api_guide/auth_update_apps.html

The applications should be updated on Constant Contact's dashboard and our custom SDK should be updated to use the new method.

miguelpeixe:
### All Submissions:

This PR fixes the OAuth2 API integration and improves error handling. The refresh token is now optional because the request for an access token wasn't always returning a refresh token value.

Closes #802.

#### Updating the app

You might have to update your app to the new OAuth2 API, follow instructions from this page: https://developer.constantcontact.com/api_guide/auth_update_apps.html

On the app settings, change the redirect URI to: `https://yoursite.com/wp-admin/index.php?cc_oauth2=1`

1. Check out this branch and connect to a valid and reviewed* Constant Contact account
2. Create a draft newsletter and attempt to send a test email
3. If you have a free account, you might see the error message: "Constant Contact: Your current plan does not allow the use of Test Send."
4. If you have access to the test email feature you should see a success message. Confirm you've received the email
5. Select a list and send the campaign
6. Confirm the campaign is sent

*Accounts can be under "compliance review" and must be approved before using some functionality.

### Other information:

### DIFF
--- a/includes/service-providers/constant_contact/class-newspack-newsletters-constant-contact-sdk.php
+++ b/includes/service-providers/constant_contact/class-newspack-newsletters-constant-contact-sdk.php
@@ -74,7 +74,7 @@ final class Newspack_Newsletters_Constant_Contact_SDK {
 	 */
 	private function request( $method, $path, $options = [] ) {
 		$url = $this->base_uri . $path;
-		if ( $options['query'] ) {
+		if ( isset( $options['query'] ) ) {
 			$url = add_query_arg( $options['query'], $url );
 			unset( $options['query'] );
 		}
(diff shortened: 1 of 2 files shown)


adekbadek (changes requested):
The connection to the API itself works, but when receiving the "Your current plan does not allow the use of Test Send" error, it does not show up in the editor notice (but I see it in the devtools).

Also, I've pushed a small change in f3d8f1490b95df2bb9bbd1b9c609d53914d5a248.

miguelpeixe:
I'm not able to reproduce that error. This is how it's rendered for me:

image

Can you provide steps and/or screenshots?

adekbadek:
Here's what I see after authorising the API and then trying to send a test email:

image

miguelpeixe:
I have filed #964 to further investigate the reason for this intermittent error. I was able to reproduce on other occasions with different types of errors. I couldn't yet identify the reason behind it and it seems unrelated to this PR.

Can we unblock this?

adekbadek (approved):

PR #949, opened 2022-09-06
fix: adding contact w/out lists handling

adekbadek:
### All Submissions:

Reader Activation introduces adding contact w/out lists. This PR ensures this is handled gracefully by all ESPs.

1. Enable Reader Activation
2. Set the ESP to Constanct Contact 
3. On `master`, observe the post-donation webhook callback errors out, if it won't pass a list ID to `add_contact`
4. Observe it handled gracefully on this branch

### Other information:

### DIFF
--- a/includes/class-newspack-newsletters-subscription.php
+++ b/includes/class-newspack-newsletters-subscription.php
@@ -377,7 +377,7 @@ class Newspack_Newsletters_Subscription {
 		$lists = apply_filters( 'newspack_newsletters_contact_lists', $lists, $contact, $provider->service );
 
 		$errors = new WP_Error();
-		$result = true;
+		$result = [];
 
 		if ( empty( $lists ) ) {
 			try {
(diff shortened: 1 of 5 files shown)


includes/class-newspack-newsletters-subscription.php
```diff
@@ -377,6 +377,7 @@ public static function add_contact( $contact, $lists = false ) {
 		$lists = apply_filters( 'newspack_newsletters_contact_lists', $lists, $contact, $provider->service );
 
 		$errors = new WP_Error();
+		$result = true;
```
miguelpeixe, line 380: The returned value should either be an `array` or `WP_Error` and this variable should always have a value. Why create it early?

miguelpeixe (changes requested):

includes/class-newspack-newsletters-subscription.php
adekbadek: Assigment of `$result` happens in a try/catch. If it errors out, the `$result` will never be assigned and so `is_wp_error( $result )` check will cause a PHP warning. I've updated the initial value to an empty array in d5339693d684c044c5943ce973da7f625eaff736

miguelpeixe (approved):

PR #965, opened 2022-09-20
feat(subscribe-block): button text color

adekbadek:
### All Submissions:

Fixes a styling issue.

1. On `master`, insert a Newsletter Subscription block
2. Observe that the color of the text is black, which results in a poor contrast
3. Switch to this branch, observe the text color is white

### Other information:


miguelpeixe (approved):
At first, I couldn't reproduce the issue but noticed that the value being applied to button text color by the theme is the contrast to the secondary color, which will vary per site.

Since the background color is fixed, its text color should be as well.

PR #992, opened 2022-10-21
fix: standardize api error status code

### LINKED ISSUE #964: Intermittent editor error message
A failed API request sometimes shows the proper error message and sometimes `The response is not a valid JSON response.`:

image

image

It is not yet clear the reason for such behavior.

miguelpeixe:
### All Submissions:

The default status code for REST API responses when returning a `WP_Error` is `500`. This in combination with the [PWA](https://wordpress.org/plugins/pwa/) plugin triggers the service worker's cached template, which is HTML content instead of the standard JSON response.

The returned HTML is not parseable, resulting in the error seen in #964.

This error is intermittent because it only occurs if the PWA's service worker is installed and active, which is most of the time.

This PR updates the error response of API requests to change the status to `400`, ensuring that the SW cached template doesn't take over.

Closes #964

1. Make sure you have [PWA](https://wordpress.org/plugins/pwa/) plugin installed
2. Setup Newsletters with ActiveCampaign
3. Modify line `167` of `includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign-controller.php` to: `return new \WP_Error( 'test_error', 'This is a test error' );` to ensure that an attempt to send a test campaign will return an error
4. Start a new newsletter draft, open devtools and send a test
5. Confirm the request is handled by the service worker and the error message is displayed correctly
6. In case the service worker did not handle the request, refresh the page without cache clearing and attempt to send a test again:

image

### Other information:

### DIFF
--- a/includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign-controller.php
+++ b/includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign-controller.php
@@ -150,6 +150,9 @@ class Newspack_Newsletters_Active_Campaign_Controller extends Newspack_Newslette
 	 */
 	public function api_retrieve( $request ) {
 		$response = $this->service_provider->retrieve( $request['id'], true );
+		if ( is_wp_error( $response ) ) {
+			$response->add_data( [ 'status' => 400 ] );
+		}
 		return \rest_ensure_response( $response );
 	}
 
@@ -169,6 +172,9 @@ class Newspack_Newsletters_Active_Campaign_Controller extends Newspack_Newslette
 			$request['id'],
 			$emails
 		);
+		if ( is_wp_error( $response ) ) {
+			$response->add_data( [ 'status' => 400 ] );
+		}
 		return \rest_ensure_response( $response );
 	}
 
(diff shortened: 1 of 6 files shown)


leogermani (commented):
It tests as described

IMO all the repetition to add the 400 code doesn't look very good.

What do you think of adding the status code to the WP_Error object when they are created (as you did in some of the cases).

For example
* https://github.com/Automattic/newspack-newsletters/blob/master/includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign.php#L146
* https://github.com/Automattic/newspack-newsletters/blob/master/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php#L127
* and other similar places...

Another option is to have a `prepare_api_response` in the `Newspack_Newsletters_Service_Provider_Controller` class and pass all responses through this method

I also left one comment of one specific method that, unlike all the others, is not returning a WP_Error where it should

miguelpeixe:
> Another option is to have a prepare_api_response in the Newspack_Newsletters_Service_Provider_Controller class and pass all responses through this method

I like this approach better. Implemented in 580ff5d7039af9fa7101af7b30fed840bc3345df

miguelpeixe:
The `verify_token()` method? It's not supposed to return `WP_Error`:

https://github.com/Automattic/newspack-newsletters/blob/fdb7d7f09311c7aceef8b5d96694f478b804403b/includes/service-providers/constant_contact/class-newspack-newsletters-constant-contact.php#L68-L75

miguelpeixe:
The introduced rule is only for handling methods that return `WP_Error`. This method returns `['valid' => false]` on error and is handled differently.

PR #1007, opened 2022-11-07
feat: bootstrapping lists CPT

leogermani:
### All Submissions:

Creates a CPT to manage Subscription Lists

1. Check the code
2. Verify the Lists menu added under Newsletters
3. Try adding/editing Lists
4. Verify the content only shows a simple textarea

### Other information:

### DIFF
--- a/includes/class-newspack-newsletters.php
+++ b/includes/class-newspack-newsletters.php
@@ -963,7 +963,7 @@ final class Newspack_Newsletters {
 		if (
 			self::NEWSPACK_NEWSLETTERS_CPT !== $screen->post_type &&
 			Newspack_Newsletters_Ads::NEWSPACK_NEWSLETTERS_ADS_CPT !== $screen->post_type &&
-			Newspack\Newsletter\Subscription_Lists::NEWSPACK_NEWSLETTERS_LIST_CPT !== $screen->post_type
+			Newspack\Newsletters\Subscription_Lists::NEWSPACK_NEWSLETTERS_LIST_CPT !== $screen->post_type
 		) {
 			return;
 		}
--- a/includes/class-newspack-newsletters.php
+++ b/includes/class-newspack-newsletters.php
@@ -962,7 +962,8 @@ final class Newspack_Newsletters {
 		$screen = get_current_screen();
 		if (
 			self::NEWSPACK_NEWSLETTERS_CPT !== $screen->post_type &&
-			Newspack_Newsletters_Ads::NEWSPACK_NEWSLETTERS_ADS_CPT !== $screen->post_type
+			Newspack_Newsletters_Ads::NEWSPACK_NEWSLETTERS_ADS_CPT !== $screen->post_type &&
+			Newspack\Newsletter\Subscription_Lists::NEWSPACK_NEWSLETTERS_LIST_CPT !== $screen->post_type
 		) {
 			return;
 		}
--- a/includes/class-subscription-lists.php
+++ b/includes/class-subscription-lists.php
@@ -5,7 +5,7 @@
  * @package Newspack
  */
 
-namespace Newspack\Newsletter;
+namespace Newspack\Newsletters;
 
 defined( 'ABSPATH' ) || exit;
 
--- a/includes/class-subscription-lists.php
+++ b/includes/class-subscription-lists.php
@@ -10,7 +10,7 @@ namespace Newspack\Newsletters;
 defined( 'ABSPATH' ) || exit;
 
 /**
- * Main Newspack Subscription Lists class.
+ * Main Newspack Newsletters Subscription Lists class.
  *
  * Subscriptions Lists are Lists which readers can subscribe to. AKA Newsletters.
  *
--- a/includes/class-subscription-lists.php
+++ b/includes/class-subscription-lists.php
@@ -1,0 +1,121 @@
+<?php
+/**
+ * Newspack Subscription Lists
+ *
+ * @package Newspack
+ */
+
+namespace Newspack\Newsletter;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Main Newspack Subscription Lists class.
+ *
+ * Subscriptions Lists are Lists which readers can subscribe to. AKA Newsletters.
+ *
+ * Each List is associated with a Audience/List in the Provider and can be associated to one or more tags in the provider
+ */
+class Subscription_Lists {
+
+	/**
+	 * CPT for Newsletter Lists.
+	 */
+	const NEWSPACK_NEWSLETTERS_LIST_CPT = 'newsletter_list';
+
+	/**
+	 * Initialize this class and register hooks
+	 *
+	 * @return void
+	 */
+	public static function init() {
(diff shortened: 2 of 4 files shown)


dkoo (changes requested):
This works as described:

Screen Shot 2022-11-08 at 4 08 34 PM

However, when running `npm ci` I get an error that I don't see while on `master`:


```
npm ERR! `npm ci` can only install packages when your package.json and package-lock.json or npm-shrinkwrap.json are in sync. Please update your lock file with `npm install` before continuing.
```

I don't see any changes to `package.json` or the lockfile, so not sure if it's related or if this branch just needs to be updated.

includes/class-subscription-lists.php
```diff
@@ -0,0 +1,121 @@
+<?php
+/**
+ * Newspack Subscription Lists
+ *
+ * @package Newspack
+ */
+
+namespace Newspack\Newsletter;
```
miguelpeixe, line 8:
```suggestion
namespace Newspack\Newsletters;
```

The plugin is called "Newsletters"

includes/class-subscription-lists.php
```diff
@@ -0,0 +1,121 @@
+<?php
+/**
+ * Newspack Subscription Lists
```
miguelpeixe, line 3:
```suggestion
 * Newspack Newsletters Subscription Lists
```

includes/class-newspack-newsletters.php
```diff
@@ -962,7 +962,8 @@ public static function branding_scripts() {
 		$screen = get_current_screen();
 		if (
 			self::NEWSPACK_NEWSLETTERS_CPT !== $screen->post_type &&
-			Newspack_Newsletters_Ads::NEWSPACK_NEWSLETTERS_ADS_CPT !== $screen->post_type
+			Newspack_Newsletters_Ads::NEWSPACK_NEWSLETTERS_ADS_CPT !== $screen->post_type &&
+			Newspack\Newsletter\Subscription_Lists::NEWSPACK_NEWSLETTERS_LIST_CPT !== $screen->post_type
```
miguelpeixe, line 966:
```suggestion
			Newspack\Newsletters\Subscription_Lists::NEWSPACK_NEWSLETTERS_LIST_CPT !== $screen->post_type
```

includes/class-newspack-newsletters.php
```diff
@@ -962,7 +962,8 @@ public static function branding_scripts() {
 		$screen = get_current_screen();
 		if (
 			self::NEWSPACK_NEWSLETTERS_CPT !== $screen->post_type &&
-			Newspack_Newsletters_Ads::NEWSPACK_NEWSLETTERS_ADS_CPT !== $screen->post_type
+			Newspack_Newsletters_Ads::NEWSPACK_NEWSLETTERS_ADS_CPT !== $screen->post_type &&
```
miguelpeixe, line 965: Will the namespace change be made in another PR?

includes/class-subscription-lists.php
```diff
@@ -0,0 +1,121 @@
...
+ *
+ * @package Newspack
+ */
+
+namespace Newspack\Newsletter;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Main Newspack Subscription Lists class.
```
miguelpeixe, line 13:
```suggestion
 * Main Newspack Newsletters Subscription Lists class.
```

includes/class-subscription-lists.php
```diff
@@ -0,0 +1,121 @@
...
+ * Subscriptions Lists are Lists which readers can subscribe to. AKA Newsletters.
+ *
+ * Each List is associated with a Audience/List in the Provider and can be associated to one or more tags in the provider
+ */
+class Subscription_Lists {
+
+	/**
+	 * CPT for Newsletter Lists.
+	 */
+	const NEWSPACK_NEWSLETTERS_LIST_CPT = 'newsletter_list';
```
miguelpeixe, line 24: We should use a proper plugin prefix for this. Perhaps `newspack_newsletters_list`?

miguelpeixe (changes requested):
The npm issue that @dkoo mentioned was resolved with 6947a698063cce879d5959cd3265563f1da3820b

The CPT is working well, some further discussion is needed in the comments below.

includes/class-newspack-newsletters.php
leogermani: what namespace change? In Ads?

includes/class-subscription-lists.php
leogermani: :+1: done

includes/class-newspack-newsletters.php
miguelpeixe: This is an ad feature inside Newsletters (`includes/class-newspack-newsletters-ads.php`). I assume the new namespace will be applied to all of the plugin classes eventually, correct?

includes/class-subscription-lists.php
leogermani: Actually no, post type names shouldn't be longer than 20 characters

includes/class-newspack-newsletters.php
leogermani: I'd love if we started using more namespaces across the codebase... and also that we relied on the autoloader so we don't have to keep `require_once` everything. But I figured this is out of the scope of this PR and this project... let's do baby steps.

BUT, what I considered doing there was to not have all these `if`s there. Instead, each of the 3 classes would enqueue the same script, each one in its own page... so we don't have this cross reference. There's a better hook to use. I could do that

includes/class-subscription-lists.php
miguelpeixe: The newsletters post type is called `newspack_nl_cpt`. We could borrow the `_nl_` and make `newspack_nl_list`?

leogermani: yep. done

newspack-newsletters.php
```diff
@@ -55,3 +55,5 @@
 require_once NEWSPACK_NEWSLETTERS_PLUGIN_FILE . '/includes/class-newspack-newsletters-quick-edit.php';
 require_once NEWSPACK_NEWSLETTERS_PLUGIN_FILE . '/includes/class-newspack-newsletters-embed.php';
 require_once NEWSPACK_NEWSLETTERS_PLUGIN_FILE . '/includes/class-newspack-newsletters.php';
+
+\Newspack\Newsletters\Subscription_Lists::init();
```
miguelpeixe, line 59: Until we fully change the strategy on how things are required and initialized, can we keep the `init()` inside the class file as we do for all other classes in the plugin?

miguelpeixe (changes requested):

newspack-newsletters.php
leogermani: Any particular reason for that?

I think we should start moving forward doing this the "right" way... meaning relying on the autoloader (so the classes are onlyu loaded if needed) and initializing things from outside the file. But a huge refactor might be traumatic, so we can so it step by step...

miguelpeixe (approved):

PR #1016, opened 2022-11-10
fix: nested list items and quote blocks in email

dkoo:
### All Submissions:

Fixes missing nested list items and Quote blocks from rendered emails. Asana reference: `/0/1200550061930446/1203326737036715`

1. Add a Quote block and a List block and then indent some of the list items like so:

Screen Shot 2022-11-10 at 12 00 01 PM

2. On `master`, send a test email and observe that the Quote block and the nested list items are dropped.
3. On this branch, send another test and confirm that all items are rendered.

### Other information:

### DIFF
--- a/includes/class-newspack-newsletters-renderer.php
+++ b/includes/class-newspack-newsletters-renderer.php
@@ -335,8 +335,8 @@ final class Newspack_Newsletters_Renderer {
 		$block_name    = $block['blockName'];
 		$attrs         = $block['attrs'];
 		$inner_blocks  = $block['innerBlocks'];
-		$inner_content = $block['innerContent'];
 		$inner_html    = $block['innerHTML'];
+		$inner_content = isset( $block['innerContent'] ) ? $block['innerContent'] : [ $inner_html ];
 
 		if ( ! isset( $attrs['innerBlocksToInsert'] ) && self::is_empty_block( $block ) ) {
 			return '';
@@ -378,7 +378,6 @@ final class Newspack_Newsletters_Renderer {
 			 */
 			case 'core/paragraph':
 			case 'core/heading':
-			case 'core/quote':
 			case 'core/site-title':
 			case 'core/site-tagline':
 			case 'newspack-newsletters/share':
@@ -715,9 +714,12 @@ final class Newspack_Newsletters_Renderer {
 				break;
 
 			/**
-			 * List block.
+			 * List, list item, and quote blocks.
+			 * These blocks may or may not contain innerBlocks with their actual content.
 			 */
 			case 'core/list':
+			case 'core/list-item':
+			case 'core/quote':
 				$text_attrs = array_merge(
 					array(
 						'padding'     => '0',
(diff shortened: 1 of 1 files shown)


includes/class-newspack-newsletters-renderer.php
```diff
@@ -730,8 +732,8 @@ public static function render_mjml_component( $block, $is_in_column = false, $is
 
 				$block_mjml_markup = '<mj-text ' . self::array_to_attributes( $text_attrs ) . '>' . $inner_content[0];
 				if ( ! empty( $inner_blocks ) && 1 < count( $inner_content ) ) {
-					foreach ( $inner_blocks as $block ) {
-						$block_mjml_markup .= $block['innerHTML'];
+					foreach ( $inner_blocks as $inner_block ) {
+						$block_mjml_markup .= self::render_mjml_component( $inner_block );
```
miguelpeixe, line 736: `<mj-text />` components are not supposed to contain other elements. Although rendered it appears fine, the markup is broken with `mj-*` elements:

```html
<ul>
  <mj-section padding="0"
    ><mj-column padding="12px" width="100%"
      ><mj-text
        padding="0"
        line-height="1.5"
        font-size="16px"
        font-family="Georgia"
        ><li>
          List<mj-section padding="0"
            ><mj-column padding="12px" width="100%"
              ><mj-text
                padding="0"
                line-height="1.5"
                font-size="16px"
                font-family="Georgia"
                ><ul>
                  <mj-section padding="0"
                    ><mj-column padding="12px" width="100%"
                      ><mj-text
                        padding="0"
                        line-height="1.5"
                        font-size="16px"
                        font-family="Georgia"
                        ><li>Sub item</li></mj-text
                      ></mj-column
                    ></mj-section
                  >
                </ul></mj-text
              ></mj-column
            ></mj-section
          >
        </li></mj-text
      ></mj-column
    ></mj-section
  >
</ul>
```

miguelpeixe (changes requested):

includes/class-newspack-newsletters-renderer.php
dkoo: [5b5b80c](https://github.com/Automattic/newspack-newsletters/pull/1016/commits/5b5b80c55aede86ec55c8f046520bd38997d862f) should avoid nesting MJML elements inside `mj-text`.

dkoo: [b62bd7e](https://github.com/Automattic/newspack-newsletters/pull/1016/commits/b62bd7efd76793d761b9a897367f22237e102244) fixes a regression caused by the previous commit which affected Group blocks and reusable blocks.

miguelpeixe (changes requested):
The nested list items are working as expected!

The blockquote generated markup is still incorrect because of a paragraph inside of `<mj-text />`:

```html
<blockquote class="wp-block-quote">
  <mj-text padding="0" line-height="1.5" font-size="16px" font-family="Georgia">
    <p>A quote</p>
  </mj-text>
  <cite>John Doe</cite>
</blockquote>
```

dkoo:
@miguelpeixe according to [MJML's documentation](https://documentation.mjml.io/#mj-text), `mj-text` elements can contain HTML tags, but not other `mj-` elements, so this should be fine.

> `mj-text` is an "ending tag", which means it can contain HTML code which will be left as it is, so it can contain HTML tags with attributes, but it cannot contain other MJML components. More information about ending tags [in this section](https://documentation.mjml.io/#ending-tags).

miguelpeixe:
We should then figure out why the `mj-text` element is part of the final output of the MJML render. It's not valid HTML markup and should be converted entirely. No MJML markup should be present in the final output.

miguelpeixe (changes requested):
We should then figure out why the mj-text element is part of the final output of the MJML render. It's not valid HTML markup and should be converted entirely. No MJML markup should be present in the final output.

dkoo:
> We should then figure out why the mj-text element is part of the final output of the MJML render.

Ah, it was because of the recursive call to `self::render_mjml_component` in the inner blocks' handling—the `core/paragraph` blocks inside quote blocks were getting wrapped in `mj-text`. 8ec5cb8 should fix this.

miguelpeixe (approved):
Thank you for the revisions! The HTML markup looks good now.

PR #1019, opened 2022-11-15
Tags support/add provider attributes

leogermani:
### All Submissions:

This PR adds new attributes to the providers classes
* labels that will be used in the UI
* a flag indicating whether the provider supports Tag

It also adds a condition and only loads the Lists CPT if the provider is set and support tags.

1. Make sure tests pass
2. Configure Constant Contact as a provider
3. Confirm that the Newsletters > Lists submenu is not added
4. Try to access `wp-admin/edit.php?post_type=newspack_nl_list` and make sure you get the "Invalid post type" message
5. Configure Mailchimp or Active Campaing as a provider
6. Confirm you now see and are able to use the Newsletters > Lists post type

### Other information:

### DIFF
--- a/includes/class-subscription-lists.php
+++ b/includes/class-subscription-lists.php
@@ -51,7 +51,7 @@ class Subscription_Lists {
 		}
 		
 		// If Service Provider is not configured yet.
-		if ( ! Newspack_Newsletters::is_service_provider_configured() ) {
+		if ( 'manual' === Newspack_Newsletters::service_provider() || ! Newspack_Newsletters::is_service_provider_configured() ) {
 			return false;
 		}
 
--- a/includes/class-subscription-lists.php
+++ b/includes/class-subscription-lists.php
@@ -7,6 +7,8 @@
 
 namespace Newspack\Newsletters;
 
+use Newspack_Newsletters;
+
 defined( 'ABSPATH' ) || exit;
 
 /**
@@ -29,9 +31,35 @@ class Subscription_Lists {
 	 * @return void
 	 */
 	public static function init() {
+		if ( ! self::should_initialize_lists() ) {
+			return;
+		}
 		add_action( 'init', [ __CLASS__, 'register_post_type' ] );
 		add_action( 'admin_menu', [ __CLASS__, 'add_submenu_item' ] );
 		add_filter( 'wp_editor_settings', [ __CLASS__, 'filter_editor_settings' ], 10, 2 );
+	}
+
+	/**
+	 * Check if we should initialize the Subscription lists
+	 *
+	 * @return boolean
+	 */
+	public static function should_initialize_lists() {
+		// We only need this on admin.
+		if ( ! is_admin() ) {
+			return false;
+		}
+		
+		// If Service Provider is not configured yet.
+		if ( ! Newspack_Newsletters::is_service_provider_configured() ) {
+			return false;
+		}
+
+		$provider = Newspack_Newsletters::get_service_provider();
+
+		// Only init if current provider supports tags.
+		return $provider::$support_tags;
+
 	}
 
 	/**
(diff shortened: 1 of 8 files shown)


miguelpeixe (changes requested):
While looking at Constant Contact's dashboard I do see [support for `tags`](https://app.constantcontact.com/pages/contacts/ui#tags). Is there an API limitation that prevents us from using it?

In any case, do we have a plan for when a provider doesn't support tags but still supports list configuration? As we currently do for Constant Contact:

Captura de Tela 2022-11-18 às 10 08 26

includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php
```diff
@@ -1061,4 +1068,21 @@ public function get_contact_data( $email, $return_details = false ) {
...
+	 *
+	 * @return array
+	 */
+	public static function get_labels() {
+		return [
+			'name'  => 'Mailchimp', // The provider name.
+			'list'  => __( 'audience', 'newspack-newsletters' ), // "list" in lower case singular format.
+			'lists' => __( 'audiences', 'newspack-newsletters' ), // "list" in lower case plural format.
+			'List'  => __( 'Audience', 'newspack-newsletters' ), // "list" in uppercase case singular format.
+			'Lists' => __( 'Audiences', 'newspack-newsletters' ), // "list" in uppercase case plural format.
```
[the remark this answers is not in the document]
leogermani: It will :)

leogermani:
> While looking at Constant Contact's dashboard I do see [support for tags](https://app.constantcontact.com/pages/contacts/ui#tags). Is there an API limitation that prevents us from using it?

No limitation. We just discussed and decided to implement the rest of the project and then come back and add support to the remaining providers (pamTN9-5Em-p2#comment-6937).

We might be able to add it for Campaign Monitor as well using custom fields

includes/class-subscription-lists.php
```diff
@@ -29,11 +31,37 @@ class Subscription_Lists {
...
+		
+		// If Service Provider is not configured yet.
+		if ( ! Newspack_Newsletters::is_service_provider_configured() ) {
+			return false;
+		}
+
+		$provider = Newspack_Newsletters::get_service_provider();
+
+		// Only init if current provider supports tags.
+		return $provider::$support_tags;
```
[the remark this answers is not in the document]
leogermani: hmmm good to know. I'll check other places I'm using the provider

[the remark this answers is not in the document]
leogermani: Fixed

miguelpeixe (approved):

includes/service-providers/class-newspack-newsletters-service-provider.php
```diff
@@ -318,4 +325,32 @@ public function send_newsletter( $post ) {
...
+
+	/**
+	 * Get one specific label for the current provider
+	 *
+	 * @param string $key The label key.
+	 * @return string Empty string in case the label is not found.
+	 */
+	public static function label( $key ) {
+		$labels = static::get_labels();
+		return $labels[ $key ] ?? '';
```
miguelpeixe, line 354:
```suggestion
		return $labels[ $key ] ?? $key;
```

Since `$key` is not slugified (`Lists` is a key) I wonder if this isn't a better practice for this method.

miguelpeixe: Oh, that should also check `isset()` to not throw warnings

leogermani: > Oh, that should also check isset() to not throw warnings

`??` does exactly that. It's a PHP 7 feature.

```PHP
return $labels[ $key ] ?? '';
return isset( $labels[ $key ] ) ? $labels[ $key ] : ''; // It's the same thing
```

That's why I don't know if it's a good idea to return `$key`... this would mean that I can pass anything and it will just return my string if the label is not found... Maybe if we had many labels this could be useful...

miguelpeixe: TIL!

> Maybe if we had many labels this could be useful.

Sounds good, we can revisit it later if it's the case.

PR #1028, opened 2022-11-22
Tags support: use tag ids

leogermani:
### All Submissions:

This PR changes how we are dealing with tags in the ESPs and linking them to our local Subscription Lists

* Create abstract method to make sure all providers implement the support tags in the same way
* Create methods to create tags, get tags id and others
* Save the ESP's tag ID as metadata to the Subscription List.

First, let's test the new methods.

* Configure both Mailchimp and Active Campaign with valid credentials
* Open their dashboards
* Go to `wp shell` and test the methods below (test it with both providers):

```PHP
$tag_name = 'Test tag';
$list_id = null; // required only for Mailchimp
$contact_email = 'test@test.com'; // The email of an existing contact

$provider = Newspack_Newsletters_Mailchimp::instance();
// $provider = Newspack_Newsletters_Active_Campaign::instance();

$tag_id = $provider->get_tag_id($tag_name, true, $list_id);
// Check it returns an ID and the tag was created in the ESP

$provider->add_tag_to_contact($contact_email, $tag_id, $list_id);
// Confirm that the tag was added to the contact (using ID)

$provider->remove_tag_from_contact('renato2@leo.com', $tag_name, $list_id);
// Confirm the tag was removed from the user

$provider->add_tag_to_contact('renato2@leo.com', $tag_name, $list_id);
// Confirm that the tag was added to the contact (using name)

$provider->remove_tag_from_contact('renato2@leo.com', $tag_id, $list_id);
// Confirm the tag was removed from the user

$provider->get_tag_by_id($tag_id, $list_id)
// Confirm you get the tag name

$another_tag = $provider->create_tag('Another tag', $list_id);
// Confirm $another_tag is an array and it has id and name
```
* Read the doc block of each new method and try to use them with many different values. Confirm they behave as you would expect, including error handling.

Now let's test the UI

1. Create a new Subscription List
2. Confirm you no longer have the option to leave the "List/Audience" field blank
3. Save it
4. Confirm you see the tag that was created in the metabox
5. Confirm the tag was actually created in the ESP
6. Change the title of the List and save it again. Confirm the Tag DOES NOT change
7. Go to the ESP and rename the Tag
8. Go back to wp-admin and save the list again. Confirm the Tag is updated to the value in ESP

### Other information:

### DIFF
--- a/includes/class-subscription-list.php
+++ b/includes/class-subscription-list.php
@@ -73,6 +73,15 @@ class Subscription_List {
 	}
 
 	/**
+	 * Generate the tag name that will be added to the ESP based on the post title
+	 *
+	 * @return string
+	 */
+	public function generate_tag_name() {
+		return 'Newspack: ' . $this->get_title();
+	}
+
+	/**
 	 * Returns the settings stored for a provider
 	 *
 	 * @param string $provider_slug The provider slug.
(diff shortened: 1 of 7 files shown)


miguelpeixe (approved):
Working like a charm 👌 

Do you think we need to worry about deleted lists and their reminiscent tags on the ESP?

leogermani:
> Working like a charm ok_hand
> 
> Do you think we need to worry about deleted lists and their reminiscent tags on the ESP?

No, I don't think we should delete tags in the ESP.

1. They do no harm there.. it's just a tag that can be ignored
2. Deleting them means losing data that could be useful.. "Who are the people that were subscribed to that old newsletter we deleted?"

PR #1049, opened 2022-12-12
feat(contact-add-ac): handle phone field

adekbadek:
### All Submissions:

Adds `Phone` field handling to AC contact upsertion.

`Phone` is one of the default fields on a contact, so when it is passed as metadata, it should be set as the default field instead. Trying to create a `Phone` metadata field will result in an error. 

1. Follow testing steps in https://github.com/Automattic/newspack-blocks/pull/1330, and add a field named `Phone` in the donate block
2. Observe that the value supplied in the field during the donation is reflected as the "Phone" datum on the contact

### Other information:


dkoo (changes requested):
Based on the code, should the field be named `PHONE` and not `Phone`?

I tried with fields named both ways and I see the Phone field in AC, but no data seems to be synced.

Screen Shot 2022-12-22 at 3 41 30 PM
Screen Shot 2022-12-22 at 3 41 18 PM

dkoo:
This custom billing field UI is no longer needed since we've pivoted to recommending WooCommerce for handling custom fields.

Issue #1022, opened 2022-11-16
Improve Constant Contact error handling


miguelpeixe:
The ESP methods [throw exceptions](https://github.com/Automattic/newspack-newsletters/blob/4e1cdc79af8fa1ad0baf78e391c39bc0112bd4e4/includes/service-providers/constant_contact/class-newspack-newsletters-constant-contact.php#L269-L271). These should be replaced with `WP_Error()` as it's the expected error handling already used for other providers.

leogermani:
Maybe related to that, if you save CC as the provider without credentials, the site throws errors and the settings page becomes unusable

leogermani:
Also, we could have only one instance of the SDK in the main provider class instead of creating an instance in every method

PR #1054, opened 2022-12-20
Add tags support to Constant Contact

leogermani:
### All Submissions:

Adds support to Tags for the Constant Contact provider. This will enable the new Lists feature for this provider.

1. Set Constant Contact as the provider with proper credentials
2. Go to Newsletter > Lists
3. Create a new list
4. Confirm that a correspondent tag is created in the provider
5. Go to Newsletter > Settings and enable the list in the Subsctiption Lists section
6. Add a registration or Newsletter subscription block to your site with that list as an option
7. As a visitor, register to a newsletter selecting that list
8. confirm the contact is created with the correct tag in the provider
9. confirm also that a Segment was created with users tagged with this tag
10. (Note: Constant Contact is not supported in the "My account" page for managing subscriptions, this is unrelated with this PR)
11. As a failsafe, make sure the same steps above are still working with other providers

Also note that the ability to select this segment in the editor when sending newsletters will be added in another PR (UPDATE: It was added in https://github.com/Automattic/newspack-newsletters/pull/1055)

### Other information:

### DIFF
--- a/includes/service-providers/constant_contact/class-newspack-newsletters-constant-contact-sdk.php
+++ b/includes/service-providers/constant_contact/class-newspack-newsletters-constant-contact-sdk.php
@@ -638,16 +638,16 @@ final class Newspack_Newsletters_Constant_Contact_SDK {
 	 * @return stdClass|WP_Error The tag object or a WP_Error if the tag was not found.
 	 */
 	public function get_tag_by_name( $name ) {
-		$limit_attempts     = 10;
-		$itemps_per_attempt = 100;
-		$cursor             = '';
+		$limit_attempts    = 10;
+		$items_per_attempt = 100;
+		$cursor            = '';
 		for ( $attempt = 1; $attempt <= $limit_attempts; $attempt++ ) {
 			$res = $this->request(
 				'GET',
 				'/contact_tags',
 				[
 					'query' => [
-						'limit'  => $itemps_per_attempt,
+						'limit'  => $items_per_attempt,
 						'cursor' => $cursor,
 					],
 				]
--- a/includes/service-providers/constant_contact/class-newspack-newsletters-constant-contact-sdk.php
+++ b/includes/service-providers/constant_contact/class-newspack-newsletters-constant-contact-sdk.php
@@ -474,7 +474,7 @@ final class Newspack_Newsletters_Constant_Contact_SDK {
 				'query' => [
 					'email'   => $email_address,
 					'status'  => 'all',
-					'include' => 'custom_fields,list_memberships',
+					'include' => 'custom_fields,list_memberships,taggings',
 				],
 			]
 		);
@@ -614,6 +614,9 @@ final class Newspack_Newsletters_Constant_Contact_SDK {
 					}
 				}
 			}
+			if ( isset( $data['taggings'] ) ) { // Using isset and not empty because this can be an empty array.
+				$body['taggings'] = $data['taggings'];
+			}
 		}
 		$res = $this->request(
 			$contact ? 'PUT' : 'POST',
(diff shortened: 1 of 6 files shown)


includes/service-providers/constant_contact/class-newspack-newsletters-constant-contact-sdk.php
```diff
@@ -622,4 +625,138 @@ public function upsert_contact( $email_address, $data = [] ) {
...
+	 *
+	 * If there are too many tags, we might not go over them all might fail in finding it, but that is a known limitation for now.
+	 * That's why we chose to do a case insensitive search, to increase the chances of finding the tag.
+	 *
+	 * @param string $name The tag name you are looking for. Case insensitive.
+	 * @return stdClass|WP_Error The tag object or a WP_Error if the tag was not found.
+	 */
+	public function get_tag_by_name( $name ) {
+		$limit_attempts     = 10;
+		$itemps_per_attempt = 100;
```
dkoo, line 642:
```suggestion
		$items_per_attempt = 100;
```

Typo?

includes/service-providers/constant_contact/class-newspack-newsletters-constant-contact-sdk.php
```diff
@@ -622,4 +625,138 @@ public function upsert_contact( $email_address, $data = [] ) {
 		);
 		return $this->get_contact( $email_address );
 	}
+
+	/**
+	 * Fetches a tag by its name
+	 *
+	 * Constant Contact API does not offer an endpoint to search for a tag by its name, so we have to query for all tags and look through them.
```
dkoo, line 632: Doesn't sound like there's anything we can do about this right now, but does this mean it's possible to create multiple lists with the same or similar names and then have this method fetch the wrong one? e.g.

My Tag 1
My tag

Searching for the second tag might return the first one if it appears in the response first.

The risk seems low, but we might want to consider some help text on the WP side to warn users about this. Maybe in a future PR

includes/service-providers/constant_contact/class-newspack-newsletters-constant-contact.php
```diff
@@ -825,4 +832,111 @@ public static function get_labels() {
...
+	/**
+	 * Retrieve the ESP's tag ID from its name
+	 *
+	 * @param string  $tag_name The tag.
+	 * @param boolean $create_if_not_found Whether to create a new tag if not found. Default to true.
+	 * @param string  $list_id The List ID.
+	 * @return int|WP_Error The tag ID on success. WP_Error on failure.
+	 */
+	public function get_tag_id( $tag_name, $create_if_not_found = true, $list_id = null ) {
+		$cc  = new Newspack_Newsletters_Constant_Contact_SDK( $this->api_key(), $this->api_secret(), $this->access_token() );
```
dkoo, line 844: Non-blocking nitpick since it doesn't seem like it will result in any real performance savings, but should we create and cache a single instance of the SDK class instead of creating a new instance each request?

includes/service-providers/constant_contact/class-newspack-newsletters-constant-contact-sdk.php
dkoo: Alternatively, it appears that the [endpoint to create tags](https://developer.constantcontact.com/api_guide/tags_create.html#) in the CC API returns the ID in the response. Could we store that ID as post meta and query using that, with the querying by name used only as a fallback?

dkoo (changes requested):
Works well as described. I have some questions that may or may not be relevant.

includes/service-providers/constant_contact/class-newspack-newsletters-constant-contact-sdk.php
leogermani: > Doesn't sound like there's anything we can do about this right now, but does this mean it's possible to create multiple lists with the same or similar names and then have this method fetch the wrong one? e.g.
> 
> My Tag 1 My tag
> 
> Searching for the second tag might return the first one if it appears in the response first.
> 
> The risk seems low, but we might want to consider some help text on the WP side to warn users about this. Maybe in a future PR

No, the search is for the exact full name. The only risk is that if you have more than 1000 tags in your account, we might not find an existing tag. We might then try to create it (assuming it doesn't already exists) but then the API will return an error saying the tag already exists - and they don't give us the ID, only the error message... so we end up in a dead end.

We could suggest them to add the existing ID to the error message when you try to create a tag that already exists.

> Alternatively, it appears that the [endpoint to create tags](https://developer.constantcontact.com/api_guide/tags_create.html#) in the CC API returns the ID in the response. Could we store that ID as post meta and query using that, with the querying by name used only as a fallback?

That's what we do! The only time we do that Search is when you are configuring a list to the provider for the first time and we check if a tag with that name already exists to use it instead of creating a new one.

includes/service-providers/constant_contact/class-newspack-newsletters-constant-contact.php
leogermani: YES! I thought the same thing, but I just followed the pattern there and didn't engage in a refactor to not inflate the PR.

I added a comment about this in [this issue](https://github.com/Automattic/newspack-newsletters/issues/1022), so we remember of it when we do some needed refactoring.

includes/service-providers/constant_contact/class-newspack-newsletters-constant-contact-sdk.php
leogermani: fixed in https://github.com/Automattic/newspack-newsletters/pull/1054/commits/d71973c7cf98c46e271efb617dee2b0ab977a681

leogermani:
We are holding this until we get a resolution whether we are shipping this feature or not.

leogermani:
We can get a new review on this one now @dkoo , thanks!

dkoo (changes requested):
This is still working as described, but something I noticed that seems true of all ESPs: updates to the local list names in WP don't persist to the corresponding ESP tag/group. Is this known/intentional? To test:

1. Create a local list, observe that it's created in the ESP
2. Update the name of the local list in WP, observe that the name no longer matches in the ESP
3. Subscribing to the local list as a reader still subscribes to the correct tag/group in the ESP, so it's more of a semantic/UX issue for publishers than a technical issue.

Another nitpick: when editing a single local list, the Newsletter > Lists menu item isn't expanded (although "Lists" is highlighted when you hover over "Newsletters").

leogermani:
> This is still working as described, but something I noticed that seems true of all ESPs: updates to the local list names in WP don't persist to the corresponding ESP tag/group. Is this known/intentional?

At some point this was intentional, but we discussed that we should change this behavior. There's a task for that already: 
1200550061930446-as-1203598613652135

> Another nitpick: when editing a single local list, the Newsletter > Lists menu item isn't expanded (although "Lists" is highlighted when you hover over "Newsletters").

Yeah I noticed that. I'll try to see if there's a workaround for it.

leogermani:
Fixed the menu behavior in https://github.com/Automattic/newspack-newsletters/pull/1092

dkoo (approved):
Thanks for the explanation! Since we're handling those two issues in other tasks/PRs, this one looks good 👍

PR #1055, opened 2022-12-20
Allow constant contact to send to segments

leogermani:
### All Submissions:

Allow users to send Newsletters to Segments using Constant Contact provider

1. Set Constant Contact as the provider with proper credentials
2. Create a new Newsletter
3. Verify the UI on the Sidebar Newsletter settings where you select the option to "SEND TO"
4. If you select one or more Lists, the Segments Select should be emptied
5. If you select one Segment, the Lists should all be unchecked
6. Play with these changes and verify that the Campaign in the Constant Contact dashboard is being updated with the correct values
7. Change the value on the CC dashboard and refresh the editor. Verify that the new values are correctly set up

### Other information:

### DIFF
--- a/includes/service-providers/constant_contact/class-newspack-newsletters-constant-contact-controller.php
+++ b/includes/service-providers/constant_contact/class-newspack-newsletters-constant-contact-controller.php
@@ -147,6 +147,42 @@ class Newspack_Newsletters_Constant_Contact_Controller extends Newspack_Newslett
 				],
 			]
 		);
+		\register_rest_route(
+			$this->service_provider::BASE_NAMESPACE . $this->service_provider->service,
+			'(?P<id>[\a-z]+)/segment/(?P<segment_id>[\a-z]+)',
+			[
+				'methods'             => \WP_REST_Server::EDITABLE,
+				'callback'            => [ $this, 'api_segment' ],
+				'permission_callback' => [ $this->service_provider, 'api_authoring_permissions_check' ],
+				'args'                => [
+					'id'         => [
+						'sanitize_callback' => 'absint',
+						'validate_callback' => [ 'Newspack_Newsletters', 'validate_newsletter_id' ],
+					],
+					'segment_id' => [
+						'sanitize_callback' => 'esc_attr',
+					],
+				],
+			]
+		);
+		\register_rest_route(
+			$this->service_provider::BASE_NAMESPACE . $this->service_provider->service,
+			'(?P<id>[\a-z]+)/segment/',
+			[
+				'methods'             => \WP_REST_Server::DELETABLE,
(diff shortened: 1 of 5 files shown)


dkoo (changes requested):
This works well as described and I don't see any issues with the code. However, I find the UI to be a bit confusing. Since lists and segments are mutually exclusive, could we implement something closer to what we have for Campaign Monitor (which shares the same restriction)?

The UI lets you select either lists or segments:

Screen Shot 2022-12-22 at 2 55 36 PM

Only after selecting an option do you see the fields to choose a list or segment:

Screen Shot 2022-12-22 at 2 55 44 PM
Screen Shot 2022-12-22 at 2 55 52 PM

I also feel like the control used to display and select lists vs. segments should be the same. I feel like `SelectControl` would be appropriate for lists rather than checkboxes.

cc @thomasguillot

leogermani:
> could we implement something closer to what we have for Campaign Monitor (which shares the same restriction)?

Sure, that's a good idea. Will do.

> I also feel like the control used to display and select lists vs. segments should be the same. I feel like SelectControl would be appropriate for lists rather than checkboxes.

Constant Contact allows you to send campaigns to more than one list. If we did that we would remove that possibility that the ESP offers.

leogermani:
I have added the Send Mode selection in the same way we do in Campaign Monitor

Captura de tela de 2022-12-23 15-14-04

I've left the Lists as checkboxes because, as I said, the provider allows that (and that's how it currently is presented, so removing this would degrade the experience).

I added a wrapper to limit the height and add a scrollbar in case there are too many lists:

Captura de tela de 2022-12-23 15-25-39

thomasguillot:
Dialog

Could we change the "send mode" to a `ButtonGroup` instead, with List being selected by default?
Also, I don't think we need the labels.

dkoo:
@leogermani thanks for clarifying about CC being able to send to multiple lists at once. I think @thomasguillot's proposed UI looks good—could we implement that?

leogermani:
Thanks @thomasguillot for looking at this. It looks much better

Captura de tela de 2023-01-09 16-09-20
Captura de tela de 2023-01-09 16-09-26

One little thing. The default style of the component leaves the label text for the secondary button blue, and not black. Is it ok or should we override it?

thomasguillot:
@leogermani thank you for the changes 😍
I think it'd look better if we could have the text and the border to use `$gray-900`

leogermani:
@thomasguillot done in https://github.com/Automattic/newspack-newsletters/pull/1055/commits/dc1bda8030dadf46d90a2d9211255ddab9e04275!

Captura de tela de 2023-01-09 19-47-36

dkoo (approved):
The new UI looks great and is much clearer to me!

PR #1063, opened 2023-01-12
feat: add caching to mailchimp

leogermani:
### All Submissions:

Closes 1203536997639429-as-1203697115315826 .

This PR adds a cache layer to the mailchimp provider to improve the performance of the editor.

It caches Segments, Groups and Merge fileds when the `retrieve` method is called to retrieve a campaign. Since this method is called every time you make a change to a Newsletter, the editor gets very slow especially if you have too many segments or groups.

#### Why so complicated?

Why not simply caching the return of the api call for a certain amount of time? What's all this?

It's a tricky situation. We can't cache this information for too long, because users can make changes directly in the MC dashboard and we want to reflect them asap.

On the other hand, if we cached this for just a couple of minutes, in practice the user will experience a slow editor every time, because Newsletters are not something you are constantly editing.

So the approach I took was to keep a cache for 10 minutes and, when the cache expires, serve the cache for the last time and trigger an async request to refresh the cache in the background.

Details in the implementation are in the doc blocks of the class (and I hope the code is readable)

NOTE: ~Before merging it, I want to test this build on a site with known performance issues to see if it delivers its promise.~ Done! It speeds up the editor to a normal/acceptable speed again.

1. Define Mailchimp as the ESP in Newsletters > Settings
2. Make sure you have loggin enabled by adding this to your wp-config:
```
define( 'NEWSPACK_LOG_LEVEL', 2);
```
3. Create a new Newsletter
4. On the Sidebar, go to "SEND TO" and select an Audience
5. Observe the logs, hte data will be fetched from the remote server:
```
[NEWSPACK-NEWSLETTERS][Newspack_Newsletters_Mailchimp_Cached_Data::get_data]: Mailchimp cache: getting data for list xxxxxxxxxx
[NEWSPACK-NEWSLETTERS][Newspack_Newsletters_Mailchimp_Cached_Data::get_data]: Mailchimp cache: serving from remote service
[NEWSPACK-NEWSLETTERS][Newspack_Newsletters_Mailchimp_Cached_Data::refresh_cached_data]: Mailchimp cache: Refreshing cache
```
6. From this point on, data will be served from the cache
7. Select a different Group or segment and you should see this in the logs:
```
[NEWSPACK-NEWSLETTERS][Newspack_Newsletters_Mailchimp_Cached_Data::get_data]: Mailchimp cache: getting data for list xxxxxxxxxx
[NEWSPACK-NEWSLETTERS][Newspack_Newsletters_Mailchimp_Cached_Data::get_data]: Mailchimp cache: serving from cache
```
8. The cache will last 10 minutes, so let's go ahead and manually clear it using
```
wp cache flush
wp transient delete newspack_nl_mailchimp_data_xxxxxx // where xxxxx is the list ID you see in the logs

```
9. Change the segment again, note that the response should be quick (take as long as when we were using the cache)
10. Check the logs that a async request was dispatched to refresh the cache, while the last_cache option was served
```
[NEWSPACK-NEWSLETTERS][Newspack_Newsletters_Mailchimp_Cached_Data::get_data]: Mailchimp cache: getting data for list xxxxxxxxxx
[NEWSPACK-NEWSLETTERS][Newspack_Newsletters_Mailchimp_Cached_Data::get_data]: Mailchimp cache: Dispatching refresh
[NEWSPACK-NEWSLETTERS][Newspack_Newsletters_Mailchimp_Cached_Data::get_data]: Mailchimp cache: serving from last_cache option
```
11. After a few moments, you should see the cache being refreshed by the async request
```
[NEWSPACK-NEWSLETTERS][Newspack_Newsletters_Mailchimp_Cached_Data::handle_dispatch_refresh]: Mailchimp cache: Handling ajax request to refresh cache
[NEWSPACK-NEWSLETTERS][Newspack_Newsletters_Mailchimp_Cached_Data::refresh_cached_data]: Mailchimp cache: Refreshing cache
```
12. Continue using the editor and confirm the cache version is being served

### Other information:

### DIFF
--- a/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp-cached-data.php
+++ b/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp-cached-data.php
@@ -129,6 +129,8 @@ final class Newspack_Newsletters_Mailchimp_Cached_Data {
 			return self::$memoized_data;
 		}
 
+		Newspack_Newsletters_Logger::log( 'Mailchimp cache: getting data for list ' . $list_id );
+
 		$data = get_transient( self::get_cache_key( $list_id ) );
 		if ( $data ) {
 			Newspack_Newsletters_Logger::log( 'Mailchimp cache: serving from cache' );
--- a/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp-cached-data.php
+++ b/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp-cached-data.php
@@ -17,8 +17,8 @@ use \DrewM\MailChimp\MailChimp;
  * The purpose of this class is to implement a non-obstrusive chat, in which refreshing the cache will happen in the background in an async request
  * and will never keep the user waiting.
  *
- * 1. It will check for the information in cache
- * 2. If it does not exist, it will check for the last_cached value, stored as an option
+ * 1. It will check for the information in cache (favor wp_cache because it's faster than options, especially with memcached)
+ * 2. If it does not exist, it will check for the last_cached value, stored as an option (with auto_load as false, to avoid unnecessary queries)
  * 3. It will then dispatch an async request to refresh the cache, while returning the last cached data immediately
  * 4. It will clear the last cached data, to make sure cache will be refreshed eventually, even if the async request fails for any reason
  * 5. The async request will fetch data from the Mailchimp server and pouplate the cache
--- a/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp-cached-data.php
+++ b/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp-cached-data.php
@@ -14,14 +14,14 @@ use \DrewM\MailChimp\MailChimp;
  *
  * This class handles fetching and caching segments and interests data from Mailchimp
  *
- * The purpose of this class is to implement a non-obstrusive chat, in which refreshing the cache will happen in the background in an async request
+ * The purpose of this class is to implement a non-obstrusive cache, in which refreshing the cache will happen in the background in an async request
  * and will never keep the user waiting.
  *
- * 1. It will check for the information in cache (favor wp_cache because it's faster than options, especially with memcached)
+ * 1. It will check for the information in cache
  * 2. If it does not exist, it will check for the last_cached value, stored as an option (with auto_load as false, to avoid unnecessary queries)
  * 3. It will then dispatch an async request to refresh the cache, while returning the last cached data immediately
  * 4. It will clear the last cached data, to make sure cache will be refreshed eventually, even if the async request fails for any reason
- * 5. The async request will fetch data from the Mailchimp server and pouplate the cache
+ * 5. The async request will fetch data from the Mailchimp server and populate the cache
  * 6. The first time we ask for data in a given list, there will be no cache nor last_cached option, so we'll fetch the data synchronously
  */
 final class Newspack_Newsletters_Mailchimp_Cached_Data {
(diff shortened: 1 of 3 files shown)


includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp-cached-data.php
```diff
@@ -0,0 +1,349 @@
...
+defined( 'ABSPATH' ) || exit;
+
+use \DrewM\MailChimp\MailChimp;
+
+/**
+ * Mailchimp cached class data
+ *
+ * This class handles fetching and caching segments and interests data from Mailchimp
+ *
+ * The purpose of this class is to implement a non-obstrusive chat, in which refreshing the cache will happen in the background in an async request
```
adekbadek, line 17: chat?

includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp-cached-data.php
```diff
@@ -0,0 +1,349 @@
...
+ * and will never keep the user waiting.
+ *
+ * 1. It will check for the information in cache (favor wp_cache because it's faster than options, especially with memcached)
+ * 2. If it does not exist, it will check for the last_cached value, stored as an option (with auto_load as false, to avoid unnecessary queries)
+ * 3. It will then dispatch an async request to refresh the cache, while returning the last cached data immediately
+ * 4. It will clear the last cached data, to make sure cache will be refreshed eventually, even if the async request fails for any reason
+ * 5. The async request will fetch data from the Mailchimp server and pouplate the cache
```
adekbadek, line 24:
```suggestion
 * 5. The async request will fetch data from the Mailchimp server and populate the cache
```

includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp-cached-data.php
```diff
@@ -0,0 +1,349 @@
...
+		$data = self::get_data( $list_id );
+		return $data['merge_fields'] ?? null;
+	}
+
+	/**
+	 * Retrieves the main Mailchimp instance
+	 *
+	 * @return Newspack_Newsletters_Mailchimp
+	 */
+	private static function mc() {
```
adekbadek, line 105: A more descriptive function name would make it more understandable :)

includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp-cached-data.php
```diff
@@ -0,0 +1,349 @@
...
+		];
+
+		wp_remote_post(
+			$url,
+			[
+				'timeout'   => 0.01,
+				'blocking'  => false,
+				'body'      => $body,
+				'sslverify' => apply_filters( 'https_local_ssl_verify', false ),
+                'cookies'   => $_COOKIE, // phpcs:ignore
```
adekbadek, line 189: Indentation is missing here.

adekbadek (changes requested):
Will this work without memchached? On my local instance, the MC API is hit (cache is refreshed) whenever the segment/group is switched in step 7.

includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp-cached-data.php
leogermani: fixed

includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp-cached-data.php
leogermani: changed

includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp-cached-data.php
leogermani: fixed

includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp-cached-data.php
leogermani: cache. fixed

leogermani:
> Will this work without memchached? On my local instance, the MC API is hit (cache is refreshed) whenever the segment/group is switched in step 7.

You're right, it won't. I was confused for a moment.

I changed the approach to use transients instead. Even though it will write to the options table by default, it will memcached (or other persistent cache option) if present.

leogermani:
FTR, I've just tested this on a site that was experiencing a severe slowness and it worked fine!

adekbadek (approved):

PR #1081, opened 2023-02-14
feat: add groups support to mailchimp

leogermani:
### All Submissions:

This PR adds all the methods to allow the Mailchimp service provider to manipulate groups. In the next PR, we will use it to make Mailchimp create Groups instead of Tags for local subscription lists.

We are going to manually test each method added by this PR

Start with a site with the Mailchimp credentials set, and 'mailchimp' selected as the provider in Newsletters > Settings

Go to your Mailchimp dashboard: Go to Audience Dashboard. Click on "Manage Audience > Manage Contacts" on the dropdown at the top right corner of the page. Scroll down and click on "Groups". We are going to check the groups being created here in this page.

1. Open a wp shell `wp shell`
2. Get a new instance of the provider class: `$mc = Newspack_Newsletters_Mailchimp::instance();`
3. Now we need to get the ID of your main list:
4. `$lists = $mc->get_lists();`
5. `$list_id = $lists[0]['id'];`
6. Ok, now we can start testing! Let's start with the first method:
7. `$mc->get_groups_category_id( $list_id );`
8. Confirm it returns a string
9. Refresh the Groups page in the Mailchimp dashboard and confirm you see the "Newspack Newsletters" category was created.

Captura de tela de 2023-02-14 15-47-39

10. Call the same method again and confirm it returns the string instantly (no requests are made to the server).
11. Delete the option and try again, to make sure there is no error when the category already exists in the server:
12. `delete_option('newspack_newsletters_mailchimp_group_category_id');`
13. `$mc->get_groups_category_id( $list_id );`
14. Let's go to the next method:
15. `$mc->get_group_id( 'Test group', false, $list_id );
16. Confirm it return an error saying the group was not found. Now let's tell it to create it:
17. `$group1 = $mc->get_group_id( 'Test group', true, $list_id );`
18. Confirm it returns a string
19. Refresh the page on the Mailchimp dashboard. Click "View groups" in the "Newspack Newsletters" category and confirm you see the new group created.
20. Also, if you visit "All Contacts" in your Mailchimp dashboard, you should see a new "Newspack Newsletters" column in the contacts table
21. Run `$mc->get_group_id( 'Test group', true, $list_id );` again and confirm the same ID is returned.
22. Next text. Let's try to get this group we have just created:
23. `$mc->get_group_by_id( $group1, $list_id );`
24. Confirm "Test group" was returned.
25. Next: `$created = $mc->create_group( 'A new group', $list_id );`
26. Go to the mailchimp Groups dashboard and confirm the new group was created
27. `echo $created['id']; ` and confirm it is a string with the grtoup id.
28. `$mc->get_group_by_id( $created['id'], $list_id );` and confirm you see "A new group".
29. Now choose an email of a contact that already exists in Mailchimp. Ex: `$contact = 'existing@contact.com';`.
30. `$mc->add_group_to_contact( $contact, $group1, $list_id);`
31. Go to "All contacts" in the Mailchimp dashboard and confirm the group was added to the contact.
32. Do the same with the other list:
33. `$mc->add_group_to_contact( $contact, $created['id'], $list_id);`
34. Remove one of the lists:
35. `$mc->remove_group_from_contact( $contact, $group1, $list_id);`
36. Confirm the changes are reflected in the contact on the Mailchimp dashboard
37. Call `$mc->get_contact_groups_ids( $contact );` and confirm it shows the correct status of the list of group ids the contact has
38. Feel free to test the methods with different combinations and order

### Other information:

### DIFF
--- a/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp-groups.php
+++ b/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp-groups.php
@@ -1,0 +1,253 @@
+<?php
+/**
+ * Service Provider: Mailchimp Groups Implementation
+ *
+ * @package Newspack
+ */
+
+defined( 'ABSPATH' ) || exit;
+
+use \DrewM\MailChimp\MailChimp;
+
+/**
+ * This tratis adds the Mailchimp Groups implementation to the Mailchimp Service Provider.
+ *
+ * In Mailchimp, Groups are also called Interests. Interests are categorized in Interest Categories.
+ *
+ * In this implementation, we will always add groups/interests to a category called "Newspack Newsletters". If this category
+ * doesn't exist, it will be created.
+ *
+ * From that point on, groups will always be added to this category.
+ */
+trait Newspack_Newsletters_Mailchimp_Groups {
+
+	/**
+	 * Gets the Newspack Newsletter group category ID, and creates it if it doesn't exist.
+	 *
+	 * @param string $list_id The list ID.
+	 * @return string|WP_Error The group category ID on success . WP_Error on failure .
+	 */
+	public function get_groups_category_id( $list_id ) {
+		$option_name        = 'newspack_newsletters_mailchimp_group_category_id';
+		$group_category_ids = get_option( $option_name );
+		if ( ! array( $group_category_ids ) ) {
+			$group_category_ids = [];
+		}
+		$group_category_id = $group_category_ids[ $list_id ] ?? null;
+		if ( $group_category_id ) {
+			return $group_category_id;
+		}
+		$mc     = new Mailchimp( $this->api_key() );
+		$create = $mc->post(
+			sprintf( 'lists/%s/interest-categories', $list_id ),
+			[
+				'title' => 'Newspack Newsletters',
+				'type'  => 'checkboxes',
+			]
+		);
+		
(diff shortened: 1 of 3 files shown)


miguelpeixe (approved):
It is working flawlessly! I really like the use of PHP's trait here. Only a small typo in the inline docs.

Not sure if there's any way around this, but publishers might already have their own existing group category with groups and contacts and will want to connect that to Newspack Newsletters' lists. Could we have a way to select the group category and support lists with pre-existing groups as well?

leogermani:
> Not sure if there's any way around this, but publishers might already have their own existing group category with groups and contacts and will want to connect that to Newspack Newsletters' lists. Could we have a way to select the group category and support lists with pre-existing groups as well?

Well, I think we can consider this in the future if it becomes a regular demand. It would be a little tricky to add this as an option, because we would have to handle changes to that option, which would not be simple... or allow each subscription list to be placed in a different category.

In any case this would increase the effort needed for this feature considerably. On the other hand, the way we do it now will work in all cases and is non-destructive - There is no real harm in having the lists in a new category if they already using another one. It might be an inconvenience but I don't think it will be a major problem.

We can also allow publishers to change the category via a constant if really needed. Just keeping in mind that if they change the constant after using the lists for a while, they would end up with two different groups in two different categories, so this would need to be set up before they create the lists.

PR #1086, opened 2023-02-17
feat: enable local lists and mailchimp groups

leogermani:
### All Submissions:

Re-enables the tag based Lists (local lists) for Active Campaign and Mailchimp.

For mailchimp, make it work with Groups instead of Tags.

1. Set your ESP as Mailchimp
2. Create a couple of Local lists in Newsletter > Lists
3. Register as a new user to your site
4. Subscribe to a newsletter
5. In Mailchimp dashboard, confirm that the "Newspack Newsletter" Group category was created (a new column in the all contacts list)
6. In Mailchimp dashboard, confirm the user was created with the expected Group
7. Go to My Account > Newsletters
8. Confirm that the page correctly shows the newsletters you have subscribed to
9. Check and uncheck different newsletters and save
10. Confirm that every time the correct Groups are applied to the contact on Mailchimp and the checkboxes are properly updated
11. In the admin (as an admin) switch the ESP to Active Campaign
12. Back to your reader user, go to My Account > Newsletters again
13. All newsletters should be unchecked
14. This user does not exist in the ESP yet, let's see if it will be created on the fly
15. Check one or several newsletters and submit
16. Confirm the user is created in the ESP
17. Play again with checking and unchecking Lists and see if they are correctly updated
18. In Active Campaign, the Lists will be stored as Tags in the contact
19. Check you error log for warnings and errors.

### Other information:

### DIFF
--- a/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php
+++ b/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php
@@ -21,7 +21,7 @@ final class Newspack_Newsletters_Mailchimp extends \Newspack_Newsletters_Service
 	 *
 	 * @var boolean
 	 */
-	public static $support_tags = false;
+	public static $support_local_lists = true;
 
 	/**
 	 * Provider name.
@@ -1219,11 +1219,15 @@ final class Newspack_Newsletters_Mailchimp extends \Newspack_Newsletters_Service
 	 */
 	public static function get_labels() {
 		return [
-			'name'  => 'Mailchimp', // The provider name.
-			'list'  => __( 'audience', 'newspack-newsletters' ), // "list" in lower case singular format.
-			'lists' => __( 'audiences', 'newspack-newsletters' ), // "list" in lower case plural format.
-			'List'  => __( 'Audience', 'newspack-newsletters' ), // "list" in uppercase case singular format.
-			'Lists' => __( 'Audiences', 'newspack-newsletters' ), // "list" in uppercase case plural format.
+			'name'                    => 'Mailchimp', // The provider name.
+			'list'                    => __( 'audience', 'newspack-newsletters' ), // "list" in lower case singular format.
+			'lists'                   => __( 'audiences', 'newspack-newsletters' ), // "list" in lower case plural format.
+			'List'                    => __( 'Audience', 'newspack-newsletters' ), // "list" in uppercase case singular format.
+			'Lists'                   => __( 'Audiences', 'newspack-newsletters' ), // "list" in uppercase case plural format.
+			'tag_prefix'              => '',
+			'tag_metabox_before_save' => __( 'Once this list is saved, a Group will be created for it.', 'newspack-newsletters' ),
+			// translators: %s is the name of the group category. "Newspack newsletters" by default.
+			'tag_metabox_after_save'  => sprintf( __( 'Group created for this list under %s:', 'newspack-newsletters' ), self::get_group_category_name() ),
 		];
 	}
 }
--- a/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php
+++ b/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php
@@ -42,6 +42,8 @@ final class Newspack_Newsletters_Mailchimp extends \Newspack_Newsletters_Service
 		add_action( 'wp_trash_post', [ $this, 'trash' ], 10, 1 );
 		add_filter( 'newspack_newsletters_process_link', [ $this, 'process_link' ], 10, 2 );
 
+		add_action( 'newspack_newsletters_subscription_lists_metabox_after_tag', [ $this, 'lists_metabox_notice' ] );
+
 		parent::__construct( $this );
 	}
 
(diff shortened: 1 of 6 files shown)


dkoo (changes requested):
Working really well, for the most part. However, I ran into some unexpected behavior which may be a result of groups/tags being a subset of audiences/lists.

1. In both Mailchimp and ActiveCampaign, I enabled an audience/list as well as the local lists. These audiences/lists happen to be the ones to which the group/tag is added.
2. In step 4 ("Subscribe to a newsletter" to Mailchimp), I deselected the audience, but selected the local list and registered.
3. In My Account, the reader is subscribed to both the audience and the group. 
4. After switching to ActiveCampaign, in My Account I selected only the local list (and not any of the other lists) and updated preferences.
5. In the ActiveCampaign dashboard, I see that the reader is subscribed to all lists and the local list as well, and refreshing My Account as the reader confirms this.

So it seems like the reader can't easily subscribe only to a local list and not its parent list upon contact creation. After that it is possible to unsubscribe from the parent audience/list while remaining subscribed to a local list.

leogermani:
Hi @dkoo , thanks for testing.

I had a look at what you describe and I'm afraid that's part of how things work and we need to document it:

**On Mailchimp:**

Using Local lists and "Audience Lists" at the same time won't work very well because of how Mailchimp works (unless the Local lists are attached to a different Audience than the Audience you offer as a List to readers).

Since a group is part of an Audience, a contact added to a group will always be added to the Audience as well. In this case, if you have a List that maps directly to the Audience, the user will be there as well.

As you mention, you can later remove yourself from the Audience List while still be part of the Group. But that actually doesn't work very well and is misleading.

Mailchimp has this concept of a contact that is "subscribed" or "unsubscribed". So you can have a Contact that has been added to an audience but is not subscribed. For example, this contact could get transactional emails, but has not agreed to receive any other emails.

So, when you go to "My account" and remove the Audience from your newsletters, what happens is that the contact is marked as "Unsubscribed" in Mailchimp (while still being part of a group). From the point of view of Contact Management, this contact is NOT subscribed and will not be eligible to receive any Campaign. So even though it's part of a group, you won't be able to send Campaigns to it.

So, in short, you should not use an Audience and Groups within that Audience as lists at the same time in Mailchimp...

What do you think we should do? Do you want to ask for advise from someone else?

**On Active Campaign:**

In AC we don't have this limitation, as tags are not tied to Lists.

What is happening there is because we (Newspack, as part of RAS) always add a new contact to what we call the "Master List". See Newspack > User Engagement > Email Service Provider Settings.

So if you offer the AC Master List as one of the options of Lists the reader can subscribe to, you'll get the inconsistency you saw.

In fact, this was not introduced by the Local Lists. If you do the same test, adding several regular Lists and then try to subscribe to other lists other than the Master List, you'll see that you will be added to the Master List as well.

dkoo:
> that's part of how things work and we need to document it

I suspected this is the case.

> So, in short, you should not use an Audience and Groups within that Audience as lists at the same time in Mailchimp...

Could we encourage this best practice by making the selection in Register and Subscribe blocks mutually exclusive? e.g. you have to choose between offering Audiences or Local Lists in each block?

> In fact, this was not introduced by the Local Lists. If you do the same test, adding several regular Lists and then try to subscribe to other lists other than the Master List, you'll see that you will be added to the Master List as well.

I understand the behavior for the Master List, but in this test scenario I had several other non-master lists and subscribing to a local list via My Account resulted in being subscribed to all available lists.

leogermani:
> Could we encourage this best practice by making the selection in Register and Subscribe blocks mutually exclusive? e.g. you have to choose between offering Audiences or Local Lists in each block?

Not so simple because if we did that we would block a legitimate use case. If the Local lists are mapped to Groups inside one Audience, you can still offer other Audiences as Lists. The conflict will only happen if you use an Audience and local lists with groups within the same audience.

> I understand the behavior for the Master List, but in this test scenario I had several other non-master lists and subscribing to a local list via My Account resulted in being subscribed to all available lists.

I could not reproduce this. The only list I get automatically added to is the Master List

dkoo:
> Not so simple because if we did that we would block a legitimate use case. If the Local lists are mapped to Groups inside one Audience, you can still offer other Audiences as Lists. The conflict will only happen if you use an Audience and local lists with groups within the same audience.

Perhaps some help text in the sidebar when selecting whom to send to would help here. If you select a local list, it can say something like "This local list is a subset of <Audience name>. Subscribers to this group will be automatically subscribed to this audience as well."

> I could not reproduce this. The only list I get automatically added to is the Master List

After some more testing, it's not _every_ list, but I always get added to at least one other list besides the Master List. Here's my setup: 

Screen Shot 2023-02-21 at 10 23 21 AM

"Master List" is the master list for this site. "Weekly Update" and "Daily" are non-master lists.  "Fun Stuff in AC" is a local list. After following steps 11-13:

> 11. In the admin (as an admin) switch the ESP to Active Campaign
> 12. Back to your reader user, go to My Account > Newsletters again
> 13. All newsletters should be unchecked

I check only "Fun stuff in AC" and save, but after refreshing I'm also subscribed to "Weekly Update". Checking the AC dashboard I can confirm that the contact is subscribed to the Master List, Weekly Update, and the Newspack: Fun Stuff in AC tag.

leogermani:
Still can't reproduce the issue with AC.

Added a warning to the Mailchimp local lists:

Captura de tela de 2023-02-21 17-13-43

includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php
```diff
@@ -1219,11 +1221,33 @@ function( $tag ) {
...
+	 *
+	 * @param array $settings The List settings.
+	 * @return void
+	 */
+	public function lists_metabox_notice( $settings ) {
+		if ( $settings['tag_name'] ) {
+			?>
+			<p class="subscription-list-warning">
+				<?php
+				esc_html_e( 'Note for Mailchimp: The group is part of the Audience. When a reader subscribes to this List, they will also be subscribed to the Audience.', 'newspack-newsletters' );
```
dkoo, line 1247:
```suggestion
				esc_html_e( 'Note for Mailchimp: The group is a subset of the Audience selected above. When a reader subscribes to this List, they will also be subscribed to the selected Audience.', 'newspack-newsletters' );
```

dkoo (approved):
> Still can't reproduce the issue with AC.

Since this is so far only replicable after switching ESPs in the admin and then trying to subscribe to new tags/lists as an existing reader (which is an unlikely scenario in production), I'm okay with leaving this.

> Added a warning to the Mailchimp local lists

Looks good to me! Made some very minor non-blocking suggestions. I think this is ready, thanks for the revisions!

dkoo:
> I check only "Fun stuff in AC" and save, but after refreshing I'm also subscribed to "Weekly Update". Checking the AC dashboard I can confirm that the contact is subscribed to the Master List, Weekly Update, and the Newspack: Fun Stuff in AC tag.

Just to close the loop on this, it was happening this way because the local list was assigned to the "Weekly Update" list, not the master list. 🤦  So it's working as expected.

PR #1091, opened 2023-02-23
feat: add origin metadata to all contacts

leogermani:
### All Submissions:

Adds a new `origin` metadata to all contacts added via the plugin so we can easily identify and filter them in the ESP data.

1. Register to the site via different entry points (register block, newsletter subscriptio, making a donation with RAS enabled)
2. Confirm that the contact is addedd to the ESP with a new `origin = newspack` metadata

### Other information:

### DIFF
--- a/includes/class-newspack-newsletters-subscription.php
+++ b/includes/class-newspack-newsletters-subscription.php
@@ -384,7 +384,7 @@ class Newspack_Newsletters_Subscription {
 		if ( ! isset( $contact['metadata'] ) ) {
 			$contact['metadata'] = [];
 		}
-		$contact['metadata']['origin'] = 'newspack';
+		$contact['metadata']['origin_newspack'] = '1';
 
 		/**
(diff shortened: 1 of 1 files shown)


dkoo (approved):
Works as described! Is there any risk in the `origin` field conflicting with like-named fields from other integrations? Should we prefix it like `newspack_origin` and/or use the "Metadata field prefix" if RAS is active?

leogermani:
> Works as described! Is there any risk in the `origin` field conflicting with like-named fields from other integrations? Should we prefix it like `newspack_origin` and/or use the "Metadata field prefix" if RAS is active?

Yes, I think it's a good idea to choose a unique name and avoid conflicts. I changed it to `origin_newspack`and then left the value `1` (true). WDYT?

dkoo:
> I changed it to `origin_newspack` and then left the value `1` (true). WDYT?

That works for me!

PR #1096, opened 2023-02-27
fix: remove list from settings when deleted

leogermani:
### All Submissions:

When a Local list is deleted, it still shows up as an option in the Subscription Block. This PR fixes it

Confirm the bug

1. Create a couple of local lists
2. Go to Newsletters > Settings an enable them
3. On a new tab, create a page and insert the Newsletter Subscription Block
4. Check that in the block options you see the Local lists as options
5. Go to Newsletters > Lists and delete/trash one of the lists
6. Go back to the page, add the Subscription block again and see that the deleted list is still there

Checkout this PR and repeat the process. Confirm that the list is no longer there in step 6

### Other information:

### DIFF
--- a/includes/class-subscription-lists.php
+++ b/includes/class-subscription-lists.php
@@ -8,6 +8,7 @@
 namespace Newspack\Newsletters;
 
 use Newspack_Newsletters;
+use Newspack_Newsletters_Subscription;
 use WP_Post;
 
 defined( 'ABSPATH' ) || exit;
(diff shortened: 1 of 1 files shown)


includes/class-subscription-lists.php
```diff
@@ -446,4 +450,48 @@ function( $list ) {
...
+		$new_list_config = [];
+		// generate a new list config without the deleted list.
+		foreach ( $list_config as $list_id => $list ) {
+			$new_list_config[] = [
+				'id'          => $list_id,
+				'active'      => $list['active'],
+				'title'       => $list['title'],
+				'description' => $list['description'],
+			];
+		}
```
miguelpeixe, line 493: What's the purpose of regenerating the list after `unset()`?

miguelpeixe (approved):
Working as advertised. NIT comment inline:

includes/class-subscription-lists.php
miguelpeixe: Now realizing, is it transforming from an array mapped by list IDs to a numbered array?

miguelpeixe: NEVERMIND. That's how the `update_lists()` method receives params 😅

leogermani: yes... `get_lists_config` give me the list in one format but `update_lists` expect them in another format... Not very beatiful. We should avoid these loose parameters `$lists` which we don't really know what they are.

PR #1110, opened 2023-03-08
Fix and improve error messages

leogermani:
### All Submissions:

Very often, error messages are not displayed when trying to set Newsletter sender or sending Test emails.

This is a little hard to test because I've seen this happening intermittently.

My best guess was that it was a race condition that sometimes would not display the error notice because of a uncaught error.

The error was being thrown because all consumers of the `apiFetchWithErrorHandling` call it without a `.catch()` callback. When it rejects the error, the error is not caught and we see it on the console.

I have removed the `reject` call from it. Since I understand that the expectation is that this method will handle the errors, I thought this was acceptable. Otherwise, we would have to add a `.catch()` callback to all the places where this method is invoked.

I'm not sure it fixes the issue 100% though. I think I saw it happening once, right after I changed the ESP, but then could never reproduce the problem again...

Test this with both Mailchimp and Active Campaign

1. Create a new Newsletter
2. Make sure the FROM fields are empty
3. Try to send a Test email
4. Make sure there's no `Uncaught (in promise) ...` error in the console
5. Make sure the error notice appears at the top of the editor

Captura de tela de 2023-03-08 12-02-42

6. Try to update the sender information with an invalid email
7. Try to update the sender email with a random domain (mailchimp will complain is not an authorized domain)
8. With everything set, try to send a test email. Active Campaign should display an error about limitations in their API

Captura de tela de 2023-03-08 12-00-27

### Other information:

### DIFF
--- a/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php
+++ b/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php
@@ -605,7 +605,7 @@ final class Newspack_Newsletters_Mailchimp extends \Newspack_Newsletters_Service
 		} catch ( Exception $e ) {
 			return new WP_Error(
 				'newspack_newsletters_mailchimp_error',
-				$e->getMessage()
+				$this->get_better_error_message( $e->getMessage() )
 			);
 		}
 	}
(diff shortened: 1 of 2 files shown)


miguelpeixe (approved):
> My best guess was that it was a race condition that sometimes would not display the error notice because of a uncaught error.

Ha! That could be it!

I couldn't figure out what was causing this intermittent generic errors either

PR #1123, opened 2023-03-20
fix(posts-inserter): preview styles for WP 6.2

miguelpeixe:
### All Submissions:

This PR fixes the styling of the Posts Inserter block preview for WP 6.2.

The use of flexbox is only applied when loading, so the spinner can be properly centered.

1. While on master and with WP 6.2, draft a new newsletter with Posts Inserter 
2. Confirm the cut-off and incorrectly aligned content
3. Check out this branch, refresh the editor
4. Confirm it's styled as expected and the spinner is also properly centered
5. Repeat the steps on the latest stable version (6.1.1) and confirm it also renders as expected

### Other information:

### DIFF
--- a/src/editor/blocks/posts-inserter/index.js
+++ b/src/editor/blocks/posts-inserter/index.js
@@ -283,7 +283,11 @@ const PostsInserterBlock = ( {
 				) }
 			</BlockControls>
 
-			<div className="newspack-posts-inserter">
+			<div
+				className={ `newspack-posts-inserter ${
+					! isReady ? 'newspack-posts-inserter--loading' : ''
+				}` }
+			>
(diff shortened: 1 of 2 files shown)


leogermani:
> The second image is the expected UI.
> 
> Was the first image tested in 6.2 or 6.1.1? I was not able to reproduce the cut-off seen in the first block.

Bot images are of the same post, just different blocks down below. I added it to show that some of them look good, and some don't.

And it was the same result on both WP versions

(yes, I've built the plugin)

miguelpeixe:
Looks like this was an issue prior to this PR (or 6.2).

I was able to reproduce this when rendering a post without a featured image and using the image left (or right) layouts. Adding some padding for such cases fixes the issue:

image

PR #1144, opened 2023-04-12
fix: restrict custom editor font sizes

miguelpeixe:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Restricts the custom font sizes to the newsletters' editor only, preventing this definition to mess with theme definitions in other places.

Some Gutenberg behavior is also not allowing the newsletters' editor to have a consistent font size experience. Small and Large font sizes are applied as expected due to an `!important` flag in their rules, which is not applied to normal, medium, and x large. Although not ideal, 27de7726c7be9c139f5ffaa1e6006199fd3a432f enforces the font sizes for the editor across all options.

Closes #755

### How to test the changes in this Pull Request:

1. Check out this branch, edit a post with different font size definitions and confirm the choices are consistent between the editor and page
2. Draft a new newsletter and confirm the font sizes are as defined in pixels (12, 16, 24 and 36)

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

PR #1185, opened 2023-05-30
fix: remove important flag from font size rules

miguelpeixe:
### All Submissions:

The use of `!important` in the font size rules is enforcing the fixed size of `20px` to be replaced with `font-size: var( --wp--preset--font-size--normal );`, which renders 16px.

This PR removes the flag from all the font-size rules. cbc25b074c4ddcc8d1166a7769a62f2b880efbd9 also fixes the reason the rules were added in #1144. `after_setup_theme` is executed too early and the global post is not available for the `self::is_editing_email()` check. The font sizes were never registered.

1. Check out this branch and draft a new newsletter
2. With 5 paragraphs, change their font sizes to S, M, L, XL, and a custom fixed of 20px 
3. Confirm they render 12px, 16px, 24px, 36px, and 20px respectively
4. Send the newsletter and confirm the sizes persist

### Other information:

### DIFF
--- a/includes/class-newspack-newsletters-editor.php
+++ b/includes/class-newspack-newsletters-editor.php
@@ -197,7 +197,9 @@ final class Newspack_Newsletters_Editor {
 	 */
 	public static function newspack_font_sizes() {
 		global $pagenow;
-		if ( 'post.php' !== $pagenow || ! isset( $_GET['post'] ) || ! self::is_editing_email( absint( $_GET['post'] ) ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+		$is_editing_email  = 'post.php' === $pagenow && isset( $_GET['post'] ) && self::is_editing_email( absint( $_GET['post'] ) ); // phpcs:ignore WordPress.Security.NonceVerification.Recommended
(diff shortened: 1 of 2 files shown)


laurelfulford (changes requested):
Oh, good catch @miguelpeixe!!

The font sizes look correct in the newsletter with this change, but the theme also goes back to using the smaller newsletter font sizes when this PR is applied 😕 

A post without the PR:
image

A post with the PR:
image

I'm not entirely sure how Gutenberg is taking these sizes and turning them into CSS, but I think whatever change was introduced in 5.8 isn't working with the post type check -- it likes it just goes ahead and uses the font size it finds for the CSS variables it creates. Now I'm wondering if the `!important` worked because the fonts weren't being registered for the Newsletter, but the CSS was just overriding the WordPress generated CSS? 

image

miguelpeixe:
Ha! That's because of the strategy to check whether we're editing a newsletter in order to use the font sizes. I'm just checking `post.php?post={id}` but should also check `post-new.php?post_type={post_type}`.

Should work now for new and existing newsletters after 2a033f6cbb88898685a1fc62f686e96445abafe0 and 187ee1eee7101b364723085c76e382b09d3d681c

Thanks again for catching another bug 😅

PR #1189, opened 2023-06-06
feat: pixel and click-tracking

miguelpeixe:
### All Submissions:

Implements pixel and click-tracking for sent newsletters.

The pixel is appended to the body of every sent email and increments a post meta as well as sends out a [data event](https://github.com/Automattic/newspack-plugin/blob/08e53ff639ffe3c58573b4f368f1db7285ff4055/includes/data-events/README.md) called `newsletter_opened` with the newsletter ID and email address.

The click tracker filters the existing `newspack_newsletters_process_link` to replace the URL with a proxied version. The redirection handler increments a post meta and sends a data event named `newsletter_clicked` with the newsletter ID, email address, and destination URL.

Both data events should be listed in the Connections -> Webhooks while editing/adding a webhook. (with `define( 'NEWSPACK_EXPERIMENTAL_WEBHOOKS', true );`)

Opens and clicks are also added as sortable columns in the dashboard's newsletters table:

image

**NOTE: This branch will be merged to `epic/revenue-optimizations` to be released along with other enhancements.**

Written tests should pass

1. Check out this branch and send a newsletter with a link to `https://newspack.com/learn-more/?v=2#post-3415` and another to any other destination
2. Open the newsletter in your email client (make sure to allow images to load), visit the dashboard, and confirm "Opened" column shows `1` for this newsletter
3. Click on a link from the newsletter and confirm you are redirected to the correct destination (with hash and parameters)
4. Refresh the dashboard and confirm the "Clicks" count is 1
5. Click on the other link and confirm it's now 2
6. Click again on the same link and confirm it's 3 (clicks are not uniquely identified)
7. Send another newsletter, open and click on links as well
8. Refresh the dashboard and confirm you are able to sort by opens and clicks

### Other information:

### DIFF
--- a/includes/class-newspack-newsletters-renderer.php
+++ b/includes/class-newspack-newsletters-renderer.php
@@ -273,10 +273,10 @@ final class Newspack_Newsletters_Renderer {
 	 * Append UTM param to links.
 	 *
 	 * @param string   $html Input HTML.
-	 * @param \WP_Post $post Post object.
+	 * @param \WP_Post $post Optional post object.
 	 * @return string HTML with processed links.
 	 */
-	public static function process_links( $html, $post ) {
+	public static function process_links( $html, $post = null ) {
 		preg_match_all( '/href="([^"]*)"/', $html, $matches );
 		$href_params = $matches[0];
 		$urls        = $matches[1];
--- a/includes/class-newspack-newsletters-renderer.php
+++ b/includes/class-newspack-newsletters-renderer.php
@@ -272,10 +272,11 @@ final class Newspack_Newsletters_Renderer {
 	/**
 	 * Append UTM param to links.
 	 *
-	 * @param string $html input HTML.
+	 * @param string   $html Input HTML.
+	 * @param \WP_Post $post Post object.
 	 * @return string HTML with processed links.
 	 */
-	public static function process_links( $html ) {
+	public static function process_links( $html, $post ) {
 		preg_match_all( '/href="([^"]*)"/', $html, $matches );
 		$href_params = $matches[0];
 		$urls        = $matches[1];
@@ -290,7 +291,8 @@ final class Newspack_Newsletters_Renderer {
 					[
 						'utm_medium' => 'email',
 					],
-					$url
+					$url,
+					$post
 				),
 				$url
 			);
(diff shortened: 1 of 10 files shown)


dkoo (changes requested):
This is working great for the most part, but I ran into some weird behavior on the opens tracking, where the first time I opened each email, the pixel would fire many times in quick succession. This caused the "opens" number to be incremented many more times than I actually opened the email. In this screenshot I've only opened each email once each, but note the 12x and 9x counts.

Screen Shot 2023-07-27 at 2 28 24 PM

miguelpeixe:
Hmmm, which email client are you using? Could it be pre-fetching the email contents, causing the pixel to load without that being an actual email open?

I'm not sure how we can protect the pixel from loads that are not opens, to some extent, it's out of our hands. 🤔

miguelpeixe:
035f07ee78a7e8a600b4c8f23da907b1ae3094f0 adds some additional logic to prevent pre-fetching and previews to trigger the pixel. The user agent information for Google's image pre-fetching comes from 2 sources:

- https://www.litmus.com/blog/gmail-prefetching-images
- https://sendgrid.com/blog/impact-of-google-prefetch/

dkoo:
> Hmmm, which email client are you using? Could it be pre-fetching the email contents, causing the pixel to load without that being an actual email open?

It's Gmail. I'll re-test with your latest changes.

dkoo:
@miguelpeixe that change doesn't fix the issue for me. I wonder if there's a way we could detect multiple requests coming from the same origin within ~1 minute and reject all but the first? It seems like the pixel gets hit multiple times within the span of a couple seconds when opening the email in Gmail. But another challenge is that the requests seem to come from different proxy domains.

Screen Shot 2023-07-28 at 4 34 14 PM

miguelpeixe:
Weird... I'm starting to think there might be multiple pixels rendering there. Can you try a new newsletter with b7bd4bc02f4fca0d15d91d6a89a5ad8b72a91346?

dkoo (approved):
I apologize about this, but I now understand why the pixel was firing multiple times. It's because I was actually receiving multiple copies of the email in my Gmail inbox (due to multiple contacts with the emailaddress+extra@gmail.com formula existing in my test Mailchimp audience). So the multiple opens were actually mostly accurate. 🤦 I see a couple of discrepancies in terms of the number of copies sent vs. simultaneous opens, but I think this is such an edge case that I don't think we need to worry about that scenario.

miguelpeixe:
Ah, I'm glad it's working!

This hunt pushed some handy enhancements, though 😄 

Thanks for the review

PR #1196, opened 2023-06-13
Newsletter Block: add a width to the email inputs to prevent overflow

laurelfulford:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

In very extreme cases -- like on smaller screens, or in narrower/super padded containers, the Newsletter Subscription block's input field can push out of the container.

This PR adds a width to the text container, which seems to fix it 🤞 

See: 1200550061930446-as-1203833168966390

### How to test the changes in this Pull Request:

1. Add a few Newsletter Subscription blocks to your content, with a mix of settings: with and without names fields, inside of groups, inside of columns. You can also copy-paste this starter content (though it'd be good to add at least one weird setting of your own):

<details>
<summary><strong>Newsletter Block text code</strong></summary>

```
<!-- wp:columns -->
<div class="wp-block-columns"><!-- wp:column -->
<div class="wp-block-column"><!-- wp:newspack-newsletters/subscribe {"displayNameField":true,"lists":["3363aba3d5"]} /--></div>
<!-- /wp:column -->

<!-- wp:column -->
<div class="wp-block-column"><!-- wp:newspack-newsletters/subscribe {"displayNameField":true,"displayLastNameField":true,"lists":["3363aba3d5"]} /--></div>
<!-- /wp:column -->

<!-- wp:column -->
<div class="wp-block-column"><!-- wp:newspack-newsletters/subscribe {"lists":["3363aba3d5"]} /--></div>
<!-- /wp:column --></div>
<!-- /wp:columns -->

<!-- wp:columns -->
<div class="wp-block-columns"><!-- wp:column -->
<div class="wp-block-column"><!-- wp:newspack-newsletters/subscribe {"displayNameField":true,"lists":["3363aba3d5"]} /--></div>
<!-- /wp:column -->

<!-- wp:column -->
<div class="wp-block-column"><!-- wp:newspack-newsletters/subscribe {"displayNameField":true,"displayLastNameField":true,"lists":["3363aba3d5"]} /--></div>
[...]

PR #1202, opened 2023-06-19
fix(subscribe-block): allow spaces in submit button

dkoo:
### All Submissions:

Fixes a minor but annoying bug in which the `RichText` component for the submit button in the editor for Subscription Form blocks doesn't allow spaces if it's nested inside a `button` element. This fix changes the `button` element to a `div` with the same styles, in the editor only. See also https://github.com/Automattic/newspack-blocks/pull/1479

Closes `1204751246381643/1204483195828260`.

1. On `master`, add a Subscription Form block to a post or page.
2. Observe that you can edit the submit button labels but can't add spaces in either case.
3. Check out this branch, repeat step 2 and confirm that you can now add spaces, and that both editor and front-end styles and behavior are otherwise unchanged.

### Other information:

### DIFF
--- a/src/blocks/subscribe/edit.js
+++ b/src/blocks/subscribe/edit.js
@@ -301,7 +301,7 @@ export default function SubscribeEdit( {
 										) }
 									</label>
 									<input type="email" placeholder={ placeholder } />
-									<div className="submit">
+									<div className="submit-button">
 										<RichText
 											onChange={ value => setAttributes( { label: value } ) }
 											placeholder={ __( 'Sign up', 'newspack' ) }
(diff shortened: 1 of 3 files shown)


leogermani (changes requested):
Looks like this broke the block.

In the editor, the button shows up in a new line, instead of at the right of the text input.

In the front-end, the block isn't displayed at all!

dkoo:
> In the editor, the button shows up in a new line, instead of at the right of the text input.

Looks like this got added in #1196—this is the intended behavior at mobile viewports, but not larger ones. This is fixed by a0e5c0b.

> In the front-end, the block isn't displayed at all!

I couldn't replicate this—could you test again with the latest change? Maybe it was being pushed into a hidden overflow area.

leogermani (approved):
Button position in the editor is fixed!

the block wasn't showing up because I forgot to enable Subscriptions Lists. My bad.

all good now!

PR #1229, opened 2023-08-08
feat: refactor how we store lists configuration

leogermani:
### All Submissions:

This PR refactors how we store the lists configuration.

Instead of saving the lists configuration as an option in the database, we now create a custom post for each list. We are using the same post type used to store "local" lists, but it has been extended so it now also serves to store "remote" lists.

"local" lists are those lists that are created via wp-admin. When we create a local list, the plugin creates a tag (or group) in the ESP
"remote" lists are the lists that are dynamically fetched from the ESP. Those lists are always managed in the ESP. Users can set an alternate title and description for them, but this information is used only in the site and is not sent to the ESP.

Everytime a user visits the admin page:
* we fetch all the lists from the ESP
* we get a post that represents that list in the local database (the custom post). If the post doesn't exist yet, we create it
* We store the ID the list has in the ESP as a metadata called "remote_id"
* We allow lists to be enabled or disabled, represented by the post status publish or draft

Test migration:
* On master, select a provider and set up a couple of lists
* Make sure the migration flag does not exist `wp option delete _newspack_newsletters_lists_migrated`
* Checkout this branch and confirm the lists configuration is still the same
* Repeat with a couple of providers and with some local lists in the mix

Local lists are still treated the same. The difference is that we now handle the activation/deactivation by changing the post status. Same thing we do with the remote lists.

We don't use the configuration options we used to have any more.

Test everything related to list management:

1. Set up the provider as Mailchimp
2. Enable a couple of lists and customize their titles and descriptions
3. Set up a subscription block and confirm it works
4. Check that both the Newsletters > Settings UI and the Newspack wizard UI work the same way
5. Create a local list and confirm it works as expected
6. Try activating and deactivating different lists
7. Change to Active Campaign
8. Do all the tests again, inlcuding local lists
9. Switch back to Mailchimp
10. Confirm that the title customizations and the local lists you had created before a back
11. Test Campaign Monitor and confirm it works (provider that don't support local lists)

### Other information:

### DIFF
--- a/includes/class-subscription-list.php
+++ b/includes/class-subscription-list.php
@@ -53,14 +53,14 @@ class Subscription_List {
 	const REMOTE_ID_META = '_remote_id';
 
 	/**
-	 * Checks if a string $id is in the format of a Subscription List Form ID
+	 * Checks if a string $id is in the format of a local Subscription List Form ID
 	 *
 	 * @see self::get_form_id
 	 * @param string $id The ID to be checked.
 	 * @return boolean
 	 */
-	public static function is_form_id( $id ) {
-		return (bool) self::get_id_from_form_id( $id );
+	public static function is_local_form_id( $id ) {
+		return (bool) self::get_id_from_local_form_id( $id );
 	}
 
 	/**
@@ -70,7 +70,7 @@ class Subscription_List {
 	 * @param string $form_id The Form id.
 	 * @return ?int The ID on success, NULL on failure
 	 */
-	public static function get_id_from_form_id( $form_id ) {
+	public static function get_id_from_local_form_id( $form_id ) {
 		if ( ! is_string( $form_id ) ) {
 			return;
 		}
@@ -92,8 +92,8 @@ class Subscription_List {
 	 */
 	public function __construct( $post_or_id ) {
 		if ( ! $post_or_id instanceof WP_Post ) {
-			if ( self::is_form_id( $post_or_id ) ) {
-				$post_or_id = self::get_id_from_form_id( $post_or_id );
+			if ( self::is_local_form_id( $post_or_id ) ) {
+				$post_or_id = self::get_id_from_local_form_id( $post_or_id );
 			}
 			$post_or_id = get_post( (int) $post_or_id );
 			if ( ! $post_or_id instanceof WP_Post ) {
@@ -234,7 +234,7 @@ class Subscription_List {
 	}
 
 	/**
-	 * Gets the link to edit this list
+	 * Gets the link to edit this list. Only local lists can be edited locally.
 	 *
 	 * In some rest requests, the post type is not registered, so we can't use get_edit_post_link
 	 *
--- a/includes/class-subscription-list.php
+++ b/includes/class-subscription-list.php
@@ -85,16 +85,57 @@ class Subscription_List {
 	}
 
 	/**
+	 * Gets a Subscription List object by its form_id
+	 *
+	 * @param string $form_id The list's form ID.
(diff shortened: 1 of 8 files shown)


dkoo:
Tested as described with no issues. I'll give the code a more thorough review tomorrow with fresher eyes

dkoo (changes requested):
Last time I didn't test actually subscribing to the local lists as a reader. Testing now, it looks like adding the contact fails with this PHP warning:

```
 PHP Notice:  Trying to access array offset on value of type null in /public/wp-content/plugins/newspack-newsletters/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php on line 1369
```

I see the corresponding Groups in Mailchimp, but no contacts are added when I sign up either via a block or My Account. Subscribing/unsubscribing from remote lists seems to work without issue.

leogermani:
> Last time I didn't test actually subscribing to the local lists as a reader. Testing now, it looks like adding the contact fails with this PHP warning:
> 
> ```
>  PHP Notice:  Trying to access array offset on value of type null in /public/wp-content/plugins/newspack-newsletters/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php on line 1369
> ```
> 
> I see the corresponding Groups in Mailchimp, but no contacts are added when I sign up either via a block or My Account. Subscribing/unsubscribing from remote lists seems to work without issue.

Yes, I saw that later and fixed it here -> https://github.com/Automattic/newspack-newsletters/pull/1238/commits/349e80f19712c6775b796c35035d450935498b77

I'm gonna cherry pick it into this branch

leogermani:
Done, thanks for catching that @dkoo

includes/class-subscription-list.php
```diff
@@ -66,17 +84,58 @@ public static function get_id_from_form_id( $form_id ) {
...
-	 * @param WP_Post|int|string $post_or_id The post object, post ID or Subscription List form ID.
+	 * @param WP_Post|int $post_or_id The post object or post ID.
 	 * @throws \InvalidArgumentException In case the post is not found.
 	 */
 	public function __construct( $post_or_id ) {
 		if ( ! $post_or_id instanceof WP_Post ) {
-			if ( self::is_form_id( $post_or_id ) ) {
-				$post_or_id = self::get_id_from_form_id( $post_or_id );
-			}
 			$post_or_id = get_post( (int) $post_or_id );
```
dkoo, line 139: I understand why this isn't working for local lists, now—it's because the post ID is getting passed to the constructor with the `newspack-` prefix. This fixes it for me and allows me to subscribe or unsubscribe for both local and remote lists. Not sure if you'll want to make it into a method for code readability.

```suggestion
			$post_or_id = str_replace( self::FORM_ID_PREFIX, '', $post_or_id );
			$post_or_id = get_post( (int) $post_or_id );
```

dkoo (changes requested):
Signing up for local lists still isn't working for me, but I dug in a bit to understand why. I think I found it—details below!

includes/class-subscription-list.php
leogermani: Thanks Derrick and sorry for not testing it all the way.

I've added the fix here -> https://github.com/Automattic/newspack-newsletters/pull/1229/commits/f4acf7f5c22c36100b2ef39e936cf873027176ee I missed those in the refactor.

I've tested and it seems to be working fine now

dkoo: It's working now!

dkoo (approved):
Working for both local and remote lists, now!

PR #1231, opened 2023-08-09
fix: allow multiple subscribe forms to be submitted on one page

dkoo:
### All Submissions:

Fixes a bug where if more than one Newsletter Subscription Form block exists on a single page, and signing up for one form results in a new reader account registration, submissions of other form instances will fail due to the initial nonce value being invalidated by the first form submission.

Also slipped in a couple of CSS changes for the error messaging, and added a proper error message response if nonce verification fails (so you get a more useful error message than the SyntaxError).

Closes `1200550061930446/1205129530451151`.

1. On `master`, using a RAS-enabled site, create a post or page that contains more than one Newsletter Subscription form, each one pointing to a different list or group in your ESP.
2. In a new session, sign up for a newsletter using a new email address. Then, without refreshing the page, sign up for another newsletter using a different form but the same email address and observe a JS console error and a hanging "disabled" state in the form:

```
Uncaught (in promise) SyntaxError: Unexpected token '<', "<!doctype "... is not valid JSON
```

3. Check out this branch and repeat step 2 with a new email address. Confirm that you're able to submit both forms without refreshing the page.
4. Repeat step 3 again in a new browser session but using the same email address (to test that it also works with pre-existing registered accounts).
5. Also confirm that the reader account is actually signed up for all of the newsletters (and only the ones) you signed up for via the different forms.

### Other information:

### DIFF
--- a/src/blocks/subscribe/index.php
+++ b/src/blocks/subscribe/index.php
@@ -322,8 +322,12 @@ function send_form_response( $data ) {
  * Process newsletter signup form.
  */
 function process_form() {
-	if ( ! isset( $_REQUEST[ FORM_ACTION ] ) || ! \wp_verify_nonce( \sanitize_text_field( $_REQUEST[ FORM_ACTION ] ), FORM_ACTION ) ) {
+	if ( ! isset( $_REQUEST[ FORM_ACTION ] ) ) {
 		return;
+	}
+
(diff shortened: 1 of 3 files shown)


miguelpeixe (approved):
Working as described! 🙌

PR #1237, opened 2023-08-14
feat: tracking settings

miguelpeixe:
### All Submissions:

1205078412187623-as-1205085335040733

Implements toggling the tracking features.

image

1. Check out this branch and visit Newsletters -> Tracking
2. Confirm you see the options toggled on by default
3. Create and send a newsletter with links
4. Confirm the links are proxied and the tracking pixel is also present
5. Visit the Tracking settings again, toggle both off and save
6. Visit the Newsletters table and confirm the "Opened" and "Clicks" columns don't render
7. Send another newsletter with links and confirm the URLs are not proxied and the pixel is not present

### Other information:


leogermani (changes requested):
Everything works as expected but I have two request/suggestions

1. The rewrite rule flush, for some reason, didn't work for me. The option was created and everything, but I had to manually run a rewrite flush to make links work.

Perhaps we could check if the rules are updated in the settings page and, if not, display a button to refresh the rules. (or automatically do it only on that page, but checking for the rules itself instead of an option?)

2. For readability and maintainability I think we could add a `is_tracking_pixel_enabled` and `is_tracking_click_enabled` methods to the admin class (or other class) and use this instead of having to call `get_option` with `true` as the default value everywhere. Not only we don't have to repeat the option name everywhere, but also we have the logic that considers `true` the default value in only one place, documented. wdyt?

leogermani:
Another idea I had here was to make the settings disable by default and flush rewrite rules when saving the settings for the first time

miguelpeixe:
Thank you for the review and suggestions!

- b5da53bf1f752c350079b56d4d5976f9833ac311: added the helper method
- e71df8aa34ca8e114c42773d90ea4a6165273791: implements rewrite rules flush on settings update
- d56e8404300a8c342bcf89da5f00165eac308216: tweaks the setting field to have a label element so it toggles the checkbox on the label click

> Another idea I had here was to make the settings disable by default and flush rewrite rules when saving the settings for the first time

That would change how we want the feature to work. I believe you experienced an edge case. If you switch to a branch without this feature and the rewrite rules are flushed, it won't flush again once you change back to this branch because the option was already stored in the db. This is not expected to happen live.

leogermani (approved):
Now I see that the rewrite rules were there already.

Can I just ask something? Use `add_option` instead of `update_option` to set the flags to avoid that annoying cache bug?

miguelpeixe:
Thanks! Good call, done in 46ec6dae89371b8fedf9149f2889266e3b38915c

PR #1238, opened 2023-08-15
Add/premium newsletters

leogermani:
### All Submissions:

Requires https://github.com/Automattic/newspack-newsletters/pull/1229 - let's wait for it to be reviewed

Implements the proof of concept to support Premium Lists tied to Woocommerce Membership plans

1. Checkout https://github.com/Automattic/newspack-plugin/pull/2611
2. Create a Subscription List via the wp-admin panel (that will create e Tag or Mailchimp group in the ESP)
3. Also enable a regular list
4. Activate Woocommerce Membership and create a Memebership plan
5. In the "Restrict content" tab, select "Subscription Lists" and add the List you just created as the restricted content, plus one other list
6. Add a Newsletters Subscription Form block to a page and select all available lists, including the premium ones
7. As an anonymous reader, confirm you don't see the premium lists in the block
8. Confirm that the registration block also does not display the premium lists
9. Perform the action required to be granted the membership (register or purchase a product)
10. Confirm that now you see the lists in the block
11. Confirm the in ESP dashboard that the user was automatically added to the lists when they were added to the plan
12. Verify your account and confirm you are able to manage your subscription to the premium lists in My Account > Newsletters
13. Cancel the membership and confirm the user was removed from the lists in the ESP dashboard
14. Try the same flow with different lists added to different membership plans at the same time
15. See pemhSX-ne-p2 for more scenarios. Think of other scenarios and see if they behave as you'd expect it

### Other information:


includes/plugins/class-woocommerce-memberships.php
```diff
@@ -0,0 +1,195 @@
...
+	 * {
+	 *     Array of User Membership arguments.
+	 * 
+	 *     @type int $user_id the user ID the membership is assigned to.
+	 *     @type int $user_membership_id the user membership ID being saved.
+	 *     @type bool $is_update whether this is a post update or a newly created membership.
+	 * }
+	 * @return void
+	 */
+	public static function add_user_to_list( $plan, $args ) {
```
miguelpeixe, line 162: This method did not work as expected for me.

I purchased the subscription which grants me the membership. The method here is reached but the user never gets subscribed to the newsletters.

While testing I also noticed that `wc_memberships_user_membership_created` is not called when a membership is manually granted through the "Members" admin dashboard.

miguelpeixe (changes requested):
Everything works well, except for the automatic newsletter subscription when joining the membership:

includes/plugins/class-woocommerce-memberships.php
leogermani: @miguelpeixe 

> I purchased the subscription which grants me the membership. The method here is reached but the user never gets subscribed to the newsletters.

Can you double-check this? Or detail the steps you took? I can't reproduce it. This has been tested multiple times with different ESPs, in different sites including some "real" sites and it works every time

> While testing I also noticed that wc_memberships_user_membership_created is not called when a membership is manually granted through the "Members" admin dashboard.

I'm gonna check this

leogermani: @miguelpeixe https://github.com/Automattic/newspack-newsletters/pull/1238/commits/69f825bf309907cf9ff4c61da233bd72419b988e makes it work when adding a membership via admin. It also removes the user if the membership is deleted (before this, it would only remove when the membership was cancelled or paused, but not deleted)

The `wc_memberships_user_membership_saved` I'm using now is a bit inconsistent and hard to rely on. It may be called multiple times during the same action, its parameters are not always what the documentation says they will be, so I kept it only for adding subscription, but I'm still relying on other hook for removing subscriptions.

Note that there's something we can improve but that's not covered by this first phase of the project:

Imagine someone buys a membership, then unsubscribe to a premium Newsletter. Later, their subscription expires and our system tries to remove them for the Premium list. Fine.

Then they renew their subscription. When this happens, they will be re-subscribed to the list, even though they have unsubscribed before.

We can keep a record and not do it in the future.

--

Please test all the cases
- membership behind registration and product
- acquiring membership via registration or buying a product
- manually adding members
- manually canceling
- manually deleting memberships
- manually changing statuses...
- etc...

miguelpeixe: I do find the idea of subscribing someone to a newsletter list automatically when getting a membership a bit non-intuitive from a technical perspective. We may find ourselves decoupling that logic in the future so it's not a default behavior.

leogermani: That is the current request from a business that fits into the MVP.

In the future I think we'll want to have an opt in in the forms and let the users choose

miguelpeixe (approved):
Working great now, including the manual route.

Not sure what was happening before, I also tried a number of times and it wasn't updating the reader subscription.

PR #1241, opened 2023-08-18
feat: preview html

miguelpeixe:
### All Submissions:

1205078412187623-as-1205078412187635

Implements HTML preview before sending the newsletter:

### With provider

image

### With "Manual" provider

image

The editor should also be able to change the ad settings or send a test email in this modal screen.

A new "Preview Email" button also renders the HTML in a modal, but without the sending functionality:

image

image

1. Make sure you have active newsletters ads configured
2. Set your newsletter provider as "Manual" and draft a new newsletter
3. Click "Preview email" and confirm the modal renders with injected ads
4. Close the modal and click to Send
5. Toggle the "Disable automatic insertion of ads" on and confirm it updates the HTML
6. Back to Newsletters settings, change your provider to a supported ESP
7. Repeat steps 3-5
8. Send a test newsletter and confirm you receive the expected content
9. Send the newsletter and confirm you receive the expected content

### Other information:

### DIFF
--- a/src/ads/newsletter-editor/index.js
+++ b/src/ads/newsletter-editor/index.js
@@ -11,13 +11,17 @@ import apiFetch from '@wordpress/api-fetch';
 
 import { NEWSLETTER_AD_CPT_SLUG } from '../../utils/consts';
 
-function NewslettersAdsSettings() {
-	const { disableAutoAds, date } = useSelect( select => {
-		const { getEditedPostAttribute } = select( 'core/editor' );
+export function DisableAutoAds( { saveOnToggle = false } ) {
+	const { disableAutoAds, date, isSaving } = useSelect( select => {
+		const { getEditedPostAttribute, isSavingPost } = select( 'core/editor' );
 		const meta = getEditedPostAttribute( 'meta' );
-		return { disableAutoAds: meta.disable_auto_ads, date: getEditedPostAttribute( 'date' ) };
+		return {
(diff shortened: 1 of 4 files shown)


leogermani (approved):
Works as described.

The one thing I noticed, that might be treated as an improvement, is the lack of user feedback.

When you open the preview modal, it takes a very long time to load the preview, because it's saving the draft and performing other requests, but there's no feedback.

When you send a test, the same thing happens, and also the Success message saying that the test was sent is hidden behind the modal.

Many times an auto-save would happen right after the preview loaded, or after I changed the ads settings, and then I would have to wait twice...

I haven't tested, but I suppose that error messages will also be hidden behind the modal

miguelpeixe:
Thanks for the review and suggestions! Some updates:

- d487fbc61a2559643bf35a07f416121b7b8c2dc1 prevents the spinner from taking over on autosaves.
- b8d2bf307f28c8f9412106b6f8acbe805d44dad7 implements the support of the API response being rendered inline instead of the editor notification

leogermani (approved):
Looks good, thanks for the improvements.

Not for now, but what do you think of adding the same sidebar that allows for sending test emails to the Preview email modal as well?

miguelpeixe:
Thanks!

> Not for now, but what do you think of adding the same sidebar that allows for sending test emails to the Preview email modal as well?

PR #1263, opened 2023-09-01
refactor: ad insertion

miguelpeixe:
### All Submissions:

Implements support for block count insertion of newsletter ads and refactors the insertion strategy to its own logic outside the renderer class.

The refactored insertion logic is borrowed from [Newspack Popups's inline prompt insertion](https://github.com/Automattic/newspack-popups/blob/f091a24c4cb7e78c114fcb5d473a47c7ef1210aa/includes/class-newspack-popups-inserter.php#L267-L431).

1. Make sure you have an ESP configured
2. Create a new newsletter ad and make its insertion strategy set to "Block count" and set the count to 3
3. Create another newsletter ad and make its insertion strategy set to "Percentage" and the value to 50%
4. Draft a new newsletter using any layout and click to send
5. Confirm the modal opens with a newsletter preview and the ads are inserted with proper positioning
6. Close the modal and add two "Newsletter Ad" blocks
7. Leave the blocks as "Automatic selection" and position them at the top of the newsletter
8. Save and click "Preview email"
9. Confirm both ads render with the correct order
10. Close the modal and change the ad block to a specific ad
11. Save and click "Preview email"
12. Confirm the expected ad renders

### Other information:


dkoo (approved):
Smoke tests worked well. Just noting for the record that ads auto-inserted into Newsletter Ad blocks are inserted by publish date (more recently created ads are inserted first) rather than by insertion position, because it's not possible to know whether ads positioned by block number should go before ads positioned by percentage, or vice versa.

This PR also is a breaking change for the Letterhead integration. The API integration is kept intact, but we don't attempt to fetch Letterhead ads instead of our own, since we're currently unable to verify that this is working as intended. We'll revisit in a future PR.

PR #1279, opened 2023-09-12
feat: add option to change subscription button block color

laurelfulford:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

This PR adds an option to change the button colour in the subscribe block; as part of this change, there's also a small theme PR to make sure the palette colours get picked up correctly: https://github.com/Automattic/newspack-theme/pull/2177

I mostly followed how this was handled for the Checkout button, with a couple bigger differences:

* I had to make some changes to get it working with the PHP output for the front-end
* In the above, I added a normally-unnecessary check for the defaults so they're not output as inline styles. This is strictly to keep existing Custom CSS working -- the styles that are already out there easily override the block's default CSS, but won't override the colours if they're inline.

I'd love feedback on any of the changes made in this PR but wanted to flag those two in particular (especially if there's a better way to go about either of them!). 

See 1200550061930446-as-1204989537763335

### How to test the changes in this Pull Request:

1. Apply this PR and https://github.com/Automattic/newspack-theme/pull/2177 to the theme and run `npm run build` for both.
2. Add the Subscription block to the editor and confirm that you now have some colour options under the Styles tab:

image

3. Try changing the colours to options in your editor palette, and confirm they work in the editor and on the front-end.
4. Try changing the colours to custom colours, and confirm they work in the editor and on the front-end. 
5. Try clearing the colours (... menu in the Color panel and click 'Reset all'), and confirm the block goes back to the default red and white.
6. To test the CSS issue I mentioned above, you can try adding the following to the Additional CSS panel in the Customizer, and confirm it overrides the default styles:

```
div.newspack-newsletters-subscribe input[type="submit"] {
	background: yellow;
	color: black;
}
```

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [ ] Have you successfully ran tests with your changes locally?

PR #1283, opened 2023-09-13
feat: add color picker for the subscribe block button

laurelfulford:
### All Submissions:

This PR adds an option to change the button colour in the subscribe block; as part of this change, there's also a small theme PR to make sure the palette colours get picked up correctly: https://github.com/Automattic/newspack-theme/pull/2177

A couple specific things to call out:

* My previous attempt (#1279) tried to copy how the Checkout button handles this, but then realized there wasn't an easy way to add text recommending using colours that stand out. So this is a bit of a franken-attempt, nesting a `PanelColorSettings` inside of a `PanelBody` 😅 and probably a couple other minor crimes. (The other option was two separate `ColorPickers`, but they ate up a lot of space!).
* I added a normally-unnecessary check for the default colours so they're not output as inline styles. This is strictly to keep existing Custom CSS working -- the styles that are already out there easily override the block's default CSS, but won't override the colours if they're inline.

I'd love feedback on any of the changes made in this PR but wanted to flag those two in particular (especially if there's a better way to go about either of them!).

See 1200550061930446-as-1204989537763335

1. Apply this PR and https://github.com/Automattic/newspack-theme/pull/2177 to the theme and run `npm run build` for both.
2. Add the Subscription block to the editor and confirm that you now have some colour options under the Styles panel.

image

3. To test the CSS issue I mentioned above, you can try publishing the post, and adding the following to the Additional CSS panel in the Customizer, and confirm it overrides the default styles:

```
div.newspack-newsletters-subscribe input[type="submit"] {
	background: yellow;
	color: black;
}
```

4. Try changing the colours to options in your editor palette, and confirm they work in the editor and on the front-end.
5. Try changing the colours to custom colours, and confirm they work in the editor and on the front-end.
6. Review the text message displayed with the colour picker, and let me know if you have any suggestions to improve it!

### Other information:

### DIFF
--- a/src/blocks/subscribe/edit.js
+++ b/src/blocks/subscribe/edit.js
@@ -21,7 +21,14 @@ import {
 	Spinner,
 	Button,
 } from '@wordpress/components';
-import { useBlockProps, InspectorControls, RichText } from '@wordpress/block-editor';
+import {
+	useBlockProps,
+	InspectorControls,
+	RichText,
+	PanelColorSettings,
+	getColorObjectByColorValue,
+} from '@wordpress/block-editor';
+import { select } from '@wordpress/data';
 
 /**
  * Internal dependencies
@@ -38,6 +45,18 @@ const editedStateOptions = [
 	{ label: __( 'Initial', 'newspack-newsletters' ), value: 'initial' },
 	{ label: __( 'Success', 'newspack-newsletters' ), value: 'success' },
 ];
+
+function getColorName( colorvalue ) {
+	let colorName = '';
+	if ( colorvalue ) {
+		const settings = select( 'core/editor' ).getEditorSettings();
+		const colorObject = getColorObjectByColorValue( settings.colors, colorvalue );
+		if ( colorObject ) {
+			colorName = colorObject.slug;
+		}
+	}
+	return colorName;
+}
 
 export default function SubscribeEdit( {
 	setAttributes,
@@ -56,6 +75,8 @@ export default function SubscribeEdit( {
 		lists,
 		displayDescription,
 		mailchimpDoubleOptIn,
+		textColor,
+		backgroundColor,
 	},
 } ) {
 	const blockProps = useBlockProps();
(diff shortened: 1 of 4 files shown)


src/blocks/subscribe/edit.js
```diff
@@ -125,6 +157,33 @@ export default function SubscribeEdit( {
 						/>
 					) }
 				</PanelBody>
+				<PanelBody
+					title={ __( 'Styles', 'newspack-newsletters' ) }
+					initialOpen={ false }
```
dkoo, line 162: Nit: This forces the panel to be closed when the editor refreshes. Omitting the attribute lets the open state persist from the last state (so if you open the panel then refresh the editor, it stays open, and vice versa). This is the default behavior for most Gutenberg sidebar panels, so it feels like we should follow the convention.

dkoo (changes requested):
Working well! Just one issue I've found that we should probably fix before merging.

In the event you have custom CSS targeting `div.newspack-newsletters-subscribe input[type="submit"]` as in the PR description, all buttons in blocks with default color attributes show on the front-end and in the customizer with the custom CSS colors. But in the editor, they're still shown with our default red background/white text. Could we make sure this is overwritten by any custom CSS?

I also wish there were a "Clear" button like we have in the regular Button block, but I did a small amount of digging and couldn't see any easy way to add this to the `PanelColorSettings` component, so this is definitely not a blocker.

src/blocks/subscribe/edit.js
laurelfulford: Oh good call! This has been updated in [d1b0f75](https://github.com/Automattic/newspack-newsletters/pull/1283/commits/d1b0f7528db4ff627a4b6a0796adc30de3dcc378).

laurelfulford:
Thanks for taking this for a spin, @dkoo! 

> ... all buttons in blocks with default color attributes show on the front-end and in the customizer with the custom CSS colors. But in the editor, they're still shown with our default red background/white text. Could we make sure this is overwritten by any custom CSS?

Unfortunately this is the case with all Custom CSS: it doesn't get pulled into the editor 😕 I think sometimes it's a minor inconvenience, but other times it creates a pretty subpar editing experience! I have a task for myself to look into this (1200550061930446-as-1205258025399045) but I'm not sure we can do it as a blanket thing -- I think we'd almost want to only pull in specific styles, which adds complexity. For now, though, this is working as expected.

> I also wish there were a "Clear" button like we have in the regular Button block, but I did a small amount of digging and couldn't see any easy way to add this to the PanelColorSettings component, so this is definitely not a blocker.

Ugh, me too! [That did show up in my first attempt](https://github.com/Automattic/newspack-newsletters/pull/1279), but we lost the ability to display any kind of message around making sure the colours pop with that approach.

dkoo (approved):
I guess as long as publishers are aware of the limitations of custom CSS, that's fine. Thanks for the `initialOpen` fix!

PR #1287, opened 2023-09-15
fix: newsletterData handling and editor improvements

miguelpeixe:
### All Submissions:

Fixes the editor stuck in dirty state even after saving due to two issues:

 - The unsynced handling of `newsletterData` meta in the front-end
 - Transforming the apiFetch original payload when saving the post

You may still get an unexpected "unsaved changes" alert with the Post Inserter block when the posts are not inserted. Its editing state regenerates the block attributes on component rerenders and fixing that issue requires a bigger refactor that is not covered in this PR.

The `newsletterData` is now handled in the backend and the apiFetch original payload is no longer modified, as it doesn't impact the middleware functionality.

Other improvements were made for the sake of simplicity and readability:

 - Refactored the `NewsletterEdit` and `LayoutPicker` components to use hooks instead of hocs
 - Changed the layout selection to use `resetEditorBlocks` instead of both `insertBlocks` and `replaceBlocks`
 - Simplified the layout sidebar panel logic

These changes should improve the editor performance, experience and reliability, as it reduces the amount of browser requests and removes logic that adds additional friction to the editor state and data handling.

**The following steps should be tested for each available ESP**

1. Draft a new newsletter and select an existing layout
2. If it includes a Post Inserter block, make sure to insert the posts and save the draft
3. Refresh and confirm it doesn't alert about unsaved changes
4. Modify the newsletter body, save as a new layout and confirm the layout is saved
5. Click on "Reset layout", select a different layout and confirm the selected layout replaces the content without issues
6. Fill the required newsletter fields, send a test email and confirm the email is sent without issues
7. Send the newsletter and confirm the newsletter is sent
8. Draft a new newsletter using the custom layout created and confirm the layout is selected and renders without issues

### Other information:

### DIFF
--- a/includes/class-newspack-newsletters.php
+++ b/includes/class-newspack-newsletters.php
@@ -156,21 +156,6 @@ final class Newspack_Newsletters {
 				],
 			],
 			[
-				'name'               => 'newsletterValidationErrors',
-				'register_meta_args' => [
-					'show_in_rest' => [
-						'schema' => [
-							'type'    => 'array',
-							'context' => [ 'edit' ],
-							'items'   => [
-								'type' => 'string',
-							],
-						],
-					],
-					'type'         => 'array',
-				],
-			],
-			[
 				'name'               => 'senderName',
 				'register_meta_args' => [
 					'show_in_rest' => [
--- a/includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign.php
+++ b/includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign.php
@@ -569,10 +569,12 @@ final class Newspack_Newsletters_Active_Campaign extends \Newspack_Newsletters_S
 	/**
 	 * Retrieve a campaign.
 	 *
-	 * @param integer $post_id Numeric ID of the Newsletter post.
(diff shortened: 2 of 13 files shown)


dkoo (approved):
This does greatly reduce the number of times I got the "unsaved changes" warning, which is totally worth it. I didn't run into any major issues sending tests or campaigns in any ESP I tested. I wasn't able to test Constant Contact because I wasn't able to get my API credentials to work, but the other ESPs looked good!

I didn't notice a significant improvement in editor performance (saving still takes a while), but this doesn't introduce any new bugs or slowness that I could se.

I did see a JS error one time when selecting a layout for a new newsletter, but I couldn't replicate it after that one time. The only consequence is that I had to select the layout again, so I don't think this is a blocker.

```
Warning: Cannot update a component (`Unknown`) while rendering a different component (`image_Image`). To locate the bad setState() call inside `image_Image`, follow the stack trace as described in https://reactjs.org/link/setstate-in-render
image_Image@https://dkoo.jurassic.tube/wp-includes/js/dist/block-library.js?ver=c7c6cdd87ab7c5489593:24003:21
```

PR #1292, opened 2023-09-20
Add newsletter_id param to newspack_newsletters_render_mjml_component

kevmarsden:
### All Submissions:

Adds the newsletter post_id to the `newspack_newsletters_render_mjml_component` filter, so it can be accessed when the filter is used.  This is helpful when you need to conditionally update a block's mjml based on the post details.  

One example where this will help CalMatters:  we have four newsletters and each one has a different image and links in the header, which is styled in a custom block.  With this update, we can use the post_id to check the newsletter category and customize the mjml for each newsletter.  

Example code to test the filter by modifying a core block:

```php
/**
 * Render a custom block.
 *
 * See https://documentation.mjml.io/ for MJML documentation.
 */
add_filter(
	'newspack_newsletters_render_mjml_component',
	function( $mjml, $block, $is_in_column, $is_in_group, $default_attrs, $is_in_list_or_quote, $newsletter_id ) {
		if ( 'core/paragraph' === $block['blockName'] ) {
		     $mjml = '<mj-text>' . $block['innerHTML'] . "Test" . '</mj-text>';
		}
		return $mjml;
	},
	10,
	7
);
```

Note:  since the filter now has seven parameters, anyone who is already using this filter will need to update the number of parameters.

### Other information:

### DIFF
--- a/includes/class-newspack-newsletters-renderer.php
+++ b/includes/class-newspack-newsletters-renderer.php
@@ -393,10 +393,11 @@ final class Newspack_Newsletters_Renderer {
 		 * @param bool        $is_in_group Whether the component is a child of a group component.
 		 * @param array       $default_attrs Default attributes for the component.
 		 * @param bool        $is_in_list_or_quote Whether the component is a child of a list or quote block.
+		 * @param int         $newsletter_id The newsletter post ID.
 		 *
 		 * @return string|null The markup to return. If null, the default markup will be generated.
 		 */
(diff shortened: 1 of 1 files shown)


miguelpeixe (approved):
Good idea, @kevmarsden! Thank you for your contribution!

> Note: since the filter now has seven parameters, anyone who is already using this filter will need to update the number of parameters.

This is only the case if you'd like to use the new `$newsletter_id` parameter. Existing filters using the first 6 parameters don't have to change.

PR #1300, opened 2023-09-25
fix: avoid duplicate firing of add_contact

dkoo:
### All Submissions:

Adds some checks and early returns to avoid extraneous firings of the `add_contact` method which will result in additional requests to the connected ESP. There is no need to sync the contact to the ESP upon reader registration if the reader registration is happening as a side effect of either subscribing to a newsletter via the Newsletter Subscription Form block, or adding the user account to a Woo Memberships plan that doesn't restrict any newsletter lists.

1. On `release`, make sure your site is connected to an ESP and you've enabled at least one Group or local list in Newspack > Engagement > Newsletters.
2. Add a Newsletter Subscription Form block to a post or page. Enable a single list for that block.
3. Add `define( 'NEWSPACK_LOG_LEVEL', 2 );` to your wp-config.php so we know when the `add_contact` method is fired.
4. On the front-end, subscribe via the block. In your debug.log, observe there are multiple instances of the `[NEWSPACK-NEWSLETTERS][Newspack_Newsletters_Subscription::add_contact]: Adding contact to list(s):` event, and that at least one is fired with no list IDs.
5. Check out this branch, repeat step 4, confirm that you only see one instance of the `Adding contact to list(s)` event with the exact list/group ID you signed up for in the block.

Note that this can also happen if you have WooCommerce Memberships set up with a membership plan that will be granted to anyone who registers for an account. In this case, we shouldn't subscribe contacts for the membership unless the membership is restricting specific newsletter lists, in which case we should subscribe them to those lists.

### Other information:

### DIFF
--- a/includes/class-newspack-newsletters-subscription.php
+++ b/includes/class-newspack-newsletters-subscription.php
@@ -481,6 +481,11 @@ class Newspack_Newsletters_Subscription {
 	 * @param array          $metadata      Metadata.
 	 */
 	public static function newspack_registered_reader( $email, $authenticate, $user_id, $existing_user, $metadata ) {
+		// Prevent double-syncing to audience if the registration method was through a Newsletter Subscription Form block.
+		if ( isset( $metadata['registration_method'] ) && 'newsletters-subscription' === $metadata['registration_method'] ) {
+			return;
+		}
+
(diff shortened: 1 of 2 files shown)


miguelpeixe (approved):
Tests well! 🙌

PR #1301, opened 2023-09-26
fix: remove honeypot field if using reCAPTCHA

dkoo:
### All Submissions:

We've seen reports from readers that the honeypot field sometimes gets autofilled by Chrome's built-in password manager, despite the `autocomplete="off"` attribute. (Probably because Chrome [intentionally ignores this attribute](https://adamsilver.io/blog/stopping-chrome-from-ignoring-autocomplete-off/)). There's not much we can do to prevent this frustrating behavior, but since the honeypot is kind of redundant/unnecessary if protecting forms with reCAPTCHA, this PR proposes we only render the honeypot if reCAPTCHA is not on, as a sort of fallback security measure. Since most Newspack sites enable reCAPTCHA to protect their form inputs, this should sidestep the autocomplete issue for most readers.

See also https://github.com/Automattic/newspack-plugin/pull/2667

1. Check out this branch.
2. With reCAPTCHA enabled in the Connections wizard, view the site on the front-end. Confirm that the Newsletter Subscription Form block doesn't contain a honeypot input field (with `name="email"` and `class="nphp"`). Confirm you can still register/login without error.
3. Disable reCAPTCHA in Connections and confirm that the block form does render the honeypot field.

### Other information:

### DIFF
--- a/src/blocks/subscribe/index.php
+++ b/src/blocks/subscribe/index.php
@@ -74,6 +74,27 @@ function enqueue_scripts() {
 function get_form_id() {
 	return \wp_unique_id( 'newspack-subscribe-' );
 }
+/**
+ * Render a honeypot field to guard against bot form submissions. Note that
+ * this field is named `email` to hopefully catch more bots who might be
+ * looking for such fields, where as the "real" field is named "npe".
+ * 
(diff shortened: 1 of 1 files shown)


adekbadek (approved):
Added a small whitespace tweak.

PR #1306, opened 2023-09-29
feat: add alt attr to mj-image elements

dkoo:
### All Submissions:

Populates an `alt` attribute on `mj-image` elements in the rendered email markup. Implements fallbacks to make sure something is always populated even if the expected field is empty. Given the following media library fields, the attribute is populated with the value of the first non-empty field:

- Alternative text
- Caption
- Description
- Title (all media library attachments get a title auto-generated from the filename, so this will always contain some value)

Screenshot 2023-09-29 at 4 47 46 PM

1. Add some values to a media library attachment in the fields described above.
2. Create a new newsletter. Add the attachment as an image block.
3. Send yourself a test email (or if using the "Other" ESP option, just click "Send" to render via MJML and get the HTML).
4. Confirm that the image block is always rendered with a populated `alt` attribute, according to the fallback logic described above.

### Other information:


miguelpeixe (approved):
Working as described!

PR #1317, opened 2023-10-06
feat(tracking): experimental pixel tracking through log files

miguelpeixe:
### All Submissions:

Implements an alternative strategy for pixel tracking, suggested by @claudiulodro, by storing temporary log files that are processed through cron instead of processing each on every pixel request.

The cron, executed every minute, looks for an existing log file and processes each stored request. It will, then, delete the processed file and generate a new one using [`tempnam()`](https://www.php.net/manual/en/function.tempnam.php) for a new unique file name.

When a new log file is created, the `np-newsletters-pixel.php` file is also dynamically replaced to account for the changed file name.

This PR also implements a filter to the tracking pixel URL, used by the new strategy to point to the new `np-newsletters-pixel.php`.

Cycling log files with a unique name is the security layer I'm going for since the logs are stored in the uploads directory. I am open to suggestions for alternatives or improvements to this strategy.

Due to its experimental nature, it's currently placed behind the `NEWSPACK_NEWSLETTERS_PIXEL_LOG_PROCESSING` PHP constant and the default/existing approach should continue to work without issues.

1. Without yet modifying the `NEWSPACK_NEWSLETTERS_PIXEL_LOG_PROCESSING`, send a new newsletter, open it, and confirm the pixel tracking behaves as expected (incremented "Opens" in the newsletters dashboard table)
2. Add `define( 'NEWSPACK_NEWSLETTERS_PIXEL_LOG_PROCESSING', true );` to your `wp-config.php` file
3. Also confirm you have `define( 'NEWSPACK_LOG_LEVEL', 2 );` so we can track the data event being dispatched
4. Send a new newsletter and open it
5. Navigate to your `uploads/2023/10` directory and confirm a `newspack_newsletters_pixel_log_{xxxxxx}` file exists and contains a base64 encoded line
6. Open the email again and confirm the file has now 2 lines
7. Navigate a bit on your site until the cron executes another log processing
8. Confirm the file is replaced and the server logged `[49160][NEWSPACK-DATA-EVENTS][Newspack\Data_Events::dispatch]: Dispatching action "newsletter_opened" via Action Scheduler`
9. Confirm the "Opens" column in the newsletters dashboard table also had its value increment for the sent newsletter

### Other information:

### DIFF
--- a/includes/tracking/class-pixel.php
+++ b/includes/tracking/class-pixel.php
@@ -290,7 +290,7 @@ final class Pixel {
 				if ( ! $newsletter_id || ! $tracking_id || ! $email_address ) {
 					continue;
 				}
-				self::track_seen( $newsletter_id, $newsletter_id, $email_address );
+				self::track_seen( $newsletter_id, $tracking_id, $email_address );
 			}
 
 			// Remove the log file after processing.
--- a/includes/tracking/class-pixel.php
+++ b/includes/tracking/class-pixel.php
@@ -245,15 +245,9 @@ final class Pixel {
 	public static function log_pixel_url( $url, $post_id, $tracking_id, $email_tag ) {
 		return \add_query_arg(
 			[
-				'data' => \base64_encode(
-					\wp_json_encode(
-						[
-							'id'  => $post_id,
-							'tid' => $tracking_id,
-						]
-					)
-				),
-				'em'   => $email_tag,
+				'id'  => $post_id,
+				'tid' => $tracking_id,
+				'em'  => $email_tag,
 			],
 			\content_url( '/np-newsletters-pixel.php' )
 		);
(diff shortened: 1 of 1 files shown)


includes/tracking/class-pixel.php
```diff
@@ -207,5 +222,125 @@ public static function render() {
...
+					continue;
+				}
+				// Values must be sanitized as they are stored in the logs without sanitization.
+				$newsletter_id = isset( $data['id'] ) ? intval( $data['id'] ) : 0;
+				$tracking_id   = isset( $data['tid'] ) ? \sanitize_text_field( $data['tid'] ) : 0;
+				$email_address = isset( $item[0] ) ? \sanitize_email( $item[0] ) : '';
+				if ( ! $newsletter_id || ! $tracking_id || ! $email_address ) {
+					continue;
+				}
+				self::track_seen( $newsletter_id, $newsletter_id, $email_address );
```
adekbadek, line 293: Should be:

```suggestion
				self::track_seen( $newsletter_id, $tracking_id, $email_address );
```

includes/tracking/class-pixel.php
```diff
@@ -207,5 +222,125 @@ public static function render() {
...
+	 * @param int    $post_id     ID of the newsletter.
+	 * @param string $tracking_id Tracking ID.
+	 * @param string $email_tag   Email address of the recipient.
+	 *
+	 * @return string
+	 */
+	public static function log_pixel_url( $url, $post_id, $tracking_id, $email_tag ) {
+		return \add_query_arg(
+			[
+				'data' => \base64_encode(
```
adekbadek, line 248: Why is this content base64-encoded?

adekbadek (changes requested):
Found some minor issues.

includes/tracking/class-pixel.php
miguelpeixe: That's actually not required at all. Updated in 2efd68c70f7218c2ccd77d6edbc5d61c12ceb082 and 669a35886aa61c04f4f35a6c287b29f7137b5d2b for simplicity.

miguelpeixe:
Moved to #1332

PR #1331, opened 2023-10-23
feat: mailchimp usage data report

adekbadek:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Adds a usage report class, for retrieving ESP data. A report is an array of arrays representing days of activity. Currently it's hard-coded to retrieve the last day only, but this can be amended later. 

Currently this is only implemented for Mailchimp, handling more ESPs to come. 

### How to test the changes in this Pull Request:

1. Observe that the tests are passing.
2. With Mailchimp set as the ESP, run `wp eval "echo print_r( (\Newspack_Newsletters::get_service_provider())->get_usage_report(1) , true );"` in the console – observe an appropriate array printed
3. Switch ESP to another, run the snippet again – observe that its value is a `newspack_newsletters_not_implemented` WP Error

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [x] Have you written new tests for your changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

PR #1338, opened 2023-10-31
feat: active-campaigns data report

adekbadek:
### All Submissions:

Usage data report (see #1331) for ActiveCampaign. 

1. With ActiveCampaign set as the ESP, run `wp eval "echo print_r( (\Newspack_Newsletters::get_service_provider())->get_usage_report(1) , true );"` in the console – observe an appropriate array printed. If there's little recent data, use a higher argument to `get_usage_report` (the argument is the number of days-ago for the report).   

### Other information:


leogermani (changes requested):
I left a few comments in the code.

One other thing I noticed is that the data types for `opens`, `clicks` etc are inconsistent. Sometimes they come as strings and sometimes as int. Also, we only get the days that have some information. I wonder if we should fill in the missing days in between with zeros for everything, just so we have a consistent output.

Speaking of consistent output, if we are going to have the same output across all ESPs, I suggest we create a simple `Report` class for the output, that way we can ensure all ESPs will follow the same structure and not just trust that the array will come in the same format. We can also handle type casting and filling in missing days in that class to keep things in the ESP cleaner.

I can draft the class after we merge this PR if you agree

includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign-usage-reports.php
```diff
@@ -0,0 +1,193 @@
...
+	}
+
+	/**
+	 * Get all contacts.
+	 *
+	 * @param int   $status Status of contacts to get.
+	 * @param date  $created_after Date after which contacts were created.
+	 * @param array $contacts Array of contacts.
+	 */
+	private static function get_all_contacts( $status = 1, $created_after = null, $contacts = [] ) {
```
leogermani, line 30: I just thought of a more efficient way to do this.

Instead of having to do an unpredictable number of requests and loop though all contacts, we could simply ask for one contact "created_after" (or "updated_after") and look at the total number in the response for today.

Then we do the same for 1 day before, and subtract the number of today to get the number of yesterday... and so on for evey day in the past.

This would ensure only one request per day and a small response ( we only need one contact ).

What do you think?

This is what I'm planning to do for Campaign Monitor (1205653098588469-as-1205653098588484)

includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign-usage-reports.php
```diff
@@ -0,0 +1,193 @@
...
+			}
+			// If the send date is before the cutoff date, skip.
+			$campaign_send_date = strtotime( $campaign['sdate'] );
+			if ( $campaign_send_date < $cutoff_datetime ) {
+				continue;
+			}
+			$report[ gmdate( 'Y-m-d', $campaign_send_date ) ] = [
+				'emails_sent' => $campaign['send_amt'],
+				'opens'       => $campaign['uniqueopens'],
+				'clicks'      => $campaign['uniquelinkclicks'],
```
leogermani, line 151: I think we are not looking at the right thing here.

You are looking at the day that the campaign was sent, and then getting the number of opens and clicks.

I think what we expect is the number of opens and clicks across all campaigns in a given day. (and it could be an open of a campaign that was sent before the cutoff date).

Let's discuss it

includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign-usage-reports.php
```diff
@@ -0,0 +1,193 @@
...
+			return $campaigns_result;
+		}
+		foreach ( $campaigns_result['campaigns'] as $campaign ) {
+			if ( ! isset( $campaign['sdate'] ) ) {
+				continue;
+			}
+			// If the send date is before the cutoff date, skip.
+			$campaign_send_date = strtotime( $campaign['sdate'] );
+			if ( $campaign_send_date < $cutoff_datetime ) {
+				continue;
```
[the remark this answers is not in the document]
adekbadek: Great suggestion!

includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign-usage-reports.php
adekbadek: You're correct, if more than one campaign was sent on a day, the stats would be overriden with the latest campaign. I'll update it.

includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign-usage-reports.php
```diff
@@ -0,0 +1,193 @@
...
+				'created_after' => gmdate( 'Y-m-d', $created_after ),
+			];
+		}
+		$active_contacts_result = $ac->api_v3_request( 'contacts', 'GET', $params );
+		if ( \is_wp_error( $active_contacts_result ) ) {
+			return $active_contacts_result;
+		}
+		$total = intval( $active_contacts_result['meta']['total'] );
+		Newspack_Newsletters_Logger::log( 'Fetching all contacts with status ' . $status . ' (' . count( $contacts ) . '/' . $total . ')' );
+		if ( count( $contacts ) === $total ) {
```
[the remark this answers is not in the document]
adekbadek: Oh, totally :D I'll update that.

includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign-usage-reports.php
```diff
@@ -0,0 +1,193 @@
...
+			return $contacts;
+		}
+		return self::get_all_contacts( $status, $created_after, array_merge( $contacts, $active_contacts_result['contacts'] ) );
```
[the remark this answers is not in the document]
adekbadek: Good suggestions, I'll update it so the contact has just `['cdate', 'udate', 'email', 'id']` keys.

includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign-usage-reports.php
adekbadek: Huh, that sounds like a smart optimisation! In the interest of time, I'd prefer to consider it only if we run into issues with the current implementation (let's not optimize prematurely).

adekbadek:
> data types for opens, clicks etc are inconsistent

Fixed in 7e4e9febd2664789e08ecfecb07408ae1cd1884e

> we only get the days that have some information. I wonder if we should fill in the missing days

This depends on what's to be done with the data. If whatever is processing the result is fine with missing days, there's no need to fill in. 

> I suggest we create a simple Report class for the output

I'm not entirely convinced on added clarity vs added complexity here. But I'm not against it.

includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign-usage-reports.php
leogermani: Not only that. We are not looking at the opens at that day, we are looking at opens of campaigns sent that day, which are different things

leogermani: Documenting what we discussed:

- Here we want to store the number of emails sent on that day.

On another check, we will loop through campaigns sent in the last 30 days and check for the number of opens and clicks
- https://developers.activecampaign.com/reference/retrieve-a-campaign
- We will try to, as close as midnight as possible, get the totals and store them - then do the same thing the next day to get the number of opens and clicks for that particular day

leogermani: ah, and we can also drop the support to get data from more than one day in the past

adekbadek: > We are not looking at the opens at that day, we are looking at opens of campaigns sent that day, which are different things

I can't see an API endpoint to retrieve per-day data about opens and clicks, so we'll have to assume that the opens & clicks happen on the date when the campaign is sent. There are such endpoints in the V1 API ([campaign_report_open_list](https://www.activecampaign.com/api/example.php?call=campaign_report_open_list), [campaign_report_link_list](https://www.activecampaign.com/api/example.php?call=campaign_report_link_list)), but I've found them unreliable, often returning no data. 
IMO the exact dates of the opens & clicks are not very relevant – conversion rate per campaign, based on per-campaign numbers, is the important metric to pay attention to when evaluating campaign success.

adekbadek:
@leogermani this is ready for a re-review now.

includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign-usage-reports.php
leogermani: But if we do that we'll have to keep checking for the past campaigns and update the data fro the previous days that are already in BQ. 

We'll need to check the campaigns details totals, store the value and subtract the number we had yesterday (https://href.li/?https://developers.activecampaign.com/reference/list-all-campaigns)

adekbadek: Hm, same goes for Mailchimp implementation, doesn't it? The activity response we're reading is for yesterday, and a campaign might still be opened tomorrow.

leogermani: No, we are precisely trying to get the same result we have with Mailchimp.

Mailchimp gives us the total number of opens/clicks for each day across all campaigns.

> The activity response we're reading is for yesterday, and a campaign might still be opened tomorrow.

No problem. Tomorrow we'll get them.

adekbadek: You're right. With Mailchimp there will be no missed data, while with the current approach to AC here, if a campaign is opened a day after data was fetched, this fact won't make it to the report. 

Taking into account that we can only get totals from the API, a way to get per-day values would be:
1. Fetch data and store it locally (e.g. campaign A had 10 total opens last time)
2. Compare with stored data and subtract it from new data (e.g. campaign A has 15 total opens now, so it must've gotten opened 5 times since the last report)

This way we would not have to update historical data in BQ. I've added some tests so it's all clearer.

adekbadek:
@leogermani – again ready for re-review 👌

leogermani:
I think we are pretty close! I left my last review in the form of a PR. Figured it would be easier to explain myself in code:

https://github.com/Automattic/newspack-newsletters/pull/1361

adekbadek:
Thanks! Merged your changes, along with some tweaks from me. Ready for another review @leogermani !

leogermani (approved):

PR #1344, opened 2023-11-06
feat(subscription): support async `add_contact()`

miguelpeixe:
### All Submissions:

When subscribing, each selected list generates one (or more) ESP API requests. When subscribing to multiple lists the form may take more than 7 seconds to resolve, which is not reasonable.

This PR implements an `async` parameter to the `Newspack_Newsletters_Subscription::add_contact` method. This parameter should be used by the Newsletter Subscription Form and Reader Registration blocks when subscribing readers.

#### Error handling

2191b5758a693994111df5d2ae9b671b7b1bff07 introduces a subscription intent system, which allows the backend to recover from server hiccups or other reasons for a subscription attempt to fail while performed asynchronously.

The subscription intent is stored and processed 3 times, once every minute, before giving up. If the subscribed email is an existing user, a user meta stores the last error so it can be displayed in the "My Account -> Newsletters" page:

image

1. Connect Newsletters to ActiveCampaign and make sure you have multiple newsletter lists (5+) enabled
2. Add the Newsletters Subscription Form and Reader Registration blocks with all the subscription lists enabled to a page
3. Make sure you are in the **master** branch
5. Visit the page and subscribe to all lists using the Newsletters Subscription Form
6. Confirm it takes 5+ seconds to resolve the form
7. Check out this branch, make sure you have `define( 'NEWSPACK_LOG_LEVEL', 2 );` and watch the logs
8. Repeat step 5 and confirm the form resolves faster and that you have `[57058][NEWSPACK-DATA-EVENTS][Newspack\Data_Events::dispatch]: Dispatching action "newsletter_subscribed"` logged after the form is resolved
9. Visit the page anonymously, register through the Reader Registration block, and confirm it also resolves in a reasonable time and the subscription executes and is logged afterward

#### Test error handling

1. Add the following to line `999` of `includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign.php`:
```php
		return new WP_Error( 'test_error', 'Some error message from the ESP' );
```

2. While authenticated as a reader, visit the page with the Newsletters Subscription Form block and subscribe to a list
3. Confirm block shows a success state
4. Visit "My Account -> Newsletters" and confirm you see the "Some error message from the ESP" notification
5. Remove the added line and wait 1 minute
6. Refresh the My Account page, confirm the error is gone and you are subscribed to the list

### Other information:

### DIFF
--- a/includes/class-newspack-newsletters-subscription.php
+++ b/includes/class-newspack-newsletters-subscription.php
@@ -53,6 +53,7 @@ class Newspack_Newsletters_Subscription {
 		/** Async `add_contact` handling */
 		add_action( 'wp_ajax_' . self::ASYNC_ACTION, [ __CLASS__, 'handle_async_subscribe' ] );
 		add_action( 'wp_ajax_nopriv_' . self::ASYNC_ACTION, [ __CLASS__, 'handle_async_subscribe' ] );
+		add_action( 'init', [ __CLASS__, 'schedule_subscription_intents_processing' ] );
 	}
 
 	/**
@@ -416,6 +417,7 @@ class Newspack_Newsletters_Subscription {
 		if ( true !== $async ) {
 			return self::add_contact_to_provider( $contact, $lists, $is_updating );
 		} else {
+			self::add_subscription_intent( $contact, $lists, $is_updating );
 			$nonce = wp_create_nonce( self::ASYNC_ACTION );
 			$url   = admin_url( 'admin-ajax.php?action=' . self::ASYNC_ACTION . '&nonce=' . $nonce );
 			$args  = [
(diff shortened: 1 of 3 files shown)


includes/class-newspack-newsletters-subscription.php
```diff
@@ -402,8 +414,187 @@ public static function add_contact( $contact, $lists = false ) {
...
+	 * Schedule subscription intents processing.
+	 */
+	public static function schedule_subscription_intents_processing() {
+		if (
+			! empty( get_option( 'newspack_newsletters_subscription_intents', [] ) ) &&
+			! \wp_next_scheduled( 'newspack_newsletters_process_subscription_intents' )
+		) {
+			\wp_schedule_single_event( time() + 60, 'newspack_newsletters_process_subscription_intents' );
+		}
+		add_action( 'newspack_newsletters_process_subscription_intents', [ __CLASS__, 'process_subscription_intents' ] );
```
leogermani, line 559: When you do this, aren't you open the possibility that this callback will run in any random request. So this could delay a random user from seeing the site if they are unlucky enough. I know this should be an edge case because we will only fall here if there were errors in the async request. Thoughts?

(just curious, any difference between using single_event and re-scheduling it to 60 seconds in the future and using recurring events?)

includes/class-newspack-newsletters-subscription.php
```diff
@@ -402,8 +414,187 @@ public static function add_contact( $contact, $lists = false ) {
...
+	 * @param string $email Optional email address to process. It'll process all
+	 *                      intents if email is not provided.
+	 *
+	 * @return void
+	 */
+	public static function process_subscription_intents( $email = null ) {
+		if ( $email ) {
+			$intents = [ self::get_subscription_intent( $email ) ];
+		} else {
+			$intents = get_option( 'newspack_newsletters_subscription_intents', [] );
```
leogermani, line 505: I'm always afraid of storing things like this is an option entry.

But not a blocker in this case.. I don't think the subscription cadence is too big

includes/class-newspack-newsletters-subscription.php
leogermani: Just thinking that this has a potential to hurt a site in case there's an issue in the ESP.

If the ESP API has an issue for a moment, error might start to pile up.

My personal preference would be not to try again, and just send the user an email telling them about the error and asking them to try again (maybe with a handy link that would do it right away).

But it's not a blocker for me, I can see how you built some guard rails here... I'm gonna test everything

leogermani (approved):
Everything works as described! I also tested removing the error code and navigating the site to see the scheduled action being triggered and the error being removed.

I'm not a fan of storing so much information in an option, with the potential to become huge in case of ESP outage combined with a site with a large audience and new subscribers. The option could become so big that could cause OOMs and slow down the site because of the cron. But I do see this is an unlikely situation, even though far from impossible.

I also think we could create a Design task to surface this error message better... Maybe something in the "My account" button showing that there's an error.. (this is an improvement for the future, not for this PR, of course)

includes/class-newspack-newsletters-subscription.php
miguelpeixe: > When you do this, aren't you open the possibility that this callback will run in any random request. So this could delay a random user from seeing the site if they are unlucky enough.

Depends on how the system leverages WP cron. With the default WP way, yes, but with appropriate caching in place, the edge case of an errored async intent should most likely be processed in an admin execution but could also hit a logged-in reader... Not sure there's any way around this, that's scheduling tasks in WP.

> (just curious, any difference between using single_event and re-scheduling it to 60 seconds in the future and using recurring events?)

There's no need to schedule the task if there are no intents in the db.

> My personal preference would be not to try again, and just send the user an email telling them about the error and asking them to try again (maybe with a handy link that would do it right away).

That's a good point, but sending emails to readers also comes with risks. RAS welcome emails were disabled because readers would often resolve subscription campaign prompts with junk email, which would impact the email server score negatively. I wouldn't rule this option out, though. I see the current implementation as a simple lower-risk approach that we can iterate over.

includes/class-newspack-newsletters-subscription.php
miguelpeixe: There could be an option serving as the intent index and have the intent data spread out in different entries. WDYT?

leogermani: WDYT of simply dispatching another async request and trying again immediately instead of having to store stuff in the DB and doing cron jobs?

If we were to spread the data out we should probably use a CPT... but it feels like too much.

If you want to stick with this implementation, maybe just put a cap of a total amount of intents we will ever store in the option as a fail safe.

includes/class-newspack-newsletters-subscription.php
leogermani: > There's no need to schedule the task if there are no intents in the db.

Gotcha! Nice!

includes/class-newspack-newsletters-subscription.php
miguelpeixe: Storing the intent gives us more room for recovery strategies. If we never store it, there is nothing we can do other than try again immediately which would not cover a short ESP outage.

Not sure about caping the intents. By doing so we are assuming the risk of having a subscription submission going nowhere.

Well... a CPT would add ~10 lines of code to the existing approach. I think it's fine, no?

leogermani: > Storing the intent gives us more room for recovery strategies. If we never store it, there is nothing we can do other than try again immediately which would not cover a short ESP outage.

But if we try three times over the next 3 minutes is not a big difference. Maybe increase the interval then?

> Not sure about caping the intents. By doing so we are assuming the risk of having a subscription submission going nowhere.

If we are confident enough that this will never grow to a point that will OOM the site, we should be confident that this will never grow to a point that a submission will be lost ;) 

> Well... a CPT would add ~10 lines of code to the existing approach. I think it's fine, no?

yeah, I think so

leogermani:
Working fine after the refactor!

PR #1349, opened 2023-11-13
feat: change usage report signature

leogermani:
### All Submissions:

A small refactor to the mailchimp usage report to

* Add the total number of contacts
* Add a Report class that will be shared by all ESPs to wirte their reports
* Drop the feature to look for reports more than one day old

See pamTN9-6YO-p2#comment-9637

With Mailchimp set as the ESP, run:
* `wp shell`
* `$rep = Newspack_Newsletters::get_service_provider())->get_usage_report();`
* `$rep->to_array()` and confirm you see the same results you see before this PR is applied but only with the data from yesterday. (see instructions in https://github.com/Automattic/newspack-newsletters/pull/1331) with the addition of `total_contacts`. 

### Other information:


includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp-usage-reports.php
```diff
@@ -25,63 +25,38 @@ private static function get_mc_instance() {
...
+				$report->emails_sent  += $activity_response['activity'][1]['emails_sent'];
+				$report->opens        += $activity_response['activity'][1]['unique_opens'];
+				$report->clicks       += $activity_response['activity'][1]['recipient_clicks'];
+				$report->subscribes   += $activity_response['activity'][1]['subs'];
+				$report->unsubscribes += $activity_response['activity'][1]['unsubs'];
+			}
 		}
 
-		return self::process_data_to_report( $lists['lists'] );
+		return $report;
```
adekbadek, line 59: Shouldn't this method should return the formatted report?

```suggestion
		return $report->to_array();
```

includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp-usage-reports.php
```diff
@@ -25,63 +25,38 @@ private static function get_mc_instance() {
...
-							'emails_sent' => $existing_day['emails_sent'] + $day['emails_sent'],
-							'opens'       => $existing_day['opens'] + $day['unique_opens'],
-							'clicks'      => $existing_day['clicks'] + $day['recipient_clicks'],
-							'subs'        => $existing_day['subs'] + $day['subs'],
-							'unsubs'      => $existing_day['unsubs'] + $day['unsubs'],
-						]
-					);
-				} else {
-					$report[ $date ] = [
-						'date'        => $date,
```
adekbadek, line 75: The report needs a `date` on the row, it's missing the the new implementation.

adekbadek (changes requested):

includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp-usage-reports.php
leogermani: I was just assuming this will be always yesterday. but it doesn't hurt to add it

includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp-usage-reports.php
leogermani: I think it could return the object. Then we modify the consumer to handle the object instead of the array. But again, either way work fine

includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp-usage-reports.php
leogermani: Done in https://github.com/Automattic/newspack-newsletters/pull/1349/commits/e25b76e5eeee1552ca4f15025347f6875b433764

includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp-usage-reports.php
leogermani: Actually, my preference is to return the object. The whole point is to normalize the output of the report in all ESPs, and the signature enforces ESPs to return the object. If we let this method return an array, an ESP is not enforced to use that class.

adekbadek (approved):

PR #1361, opened 2023-11-22
Feat/usage data ac leo

leogermani:
Here is how I see it working.

We consider sents, opens and clicks for campaigns that were sent in the last 30 days (see pamTN9-6YO-p2#comment-9637), otherwise we would always have to loop through all campaigns that were sent in history. So we might miss a few opens and clicks, but that should not be relevant.

We store the value from the campaigns today so we can subtract them tomorrow. However, we need to store their IDs because we don't want to subtract the opens and clicks from a campaign that was sent more than 30 days ago and is not part of the current response


adekbadek:
Got it, thanks! So the idea is that the initial report should be empty, because there is no prior data to compare the values with – makes sense. I've added some tweaks in 00e44cbdea165aad1e00b4eb2aecc3e199a4a878.

PR #1355, opened 2023-11-16
fix(ras): contact syncing to mailchimp

dkoo:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

When syncing metadata fields to Mailchimp merge fields, we hit the [merge-fields endpoint](https://mailchimp.com/developer/marketing/api/list-merges/list-merge-fields/). By default, this endpoint only returns a max of 10 results. So if there are more than 10 merge fields in the audience (likely there will be), we won't see all of the available fields and will attempt to create duplicates of fields that have already been created.

### How to test the changes in this Pull Request:

1. On `master` in this plugin and Newsletters, set up RAS, connect Mailchimp as the ESP and set an audience in RAS advanced settings.
4. As a new reader, purchase a donation or non-donation product without signing up for any newsletters and observe that in Mailchimp, only a small amount of reader data is synced to the contact in Mailchimp, and few or none of the [metadata fields](https://github.com/Automattic/newspack-plugin/blob/master/includes/plugins/class-newspack-newsletters.php#L26).
5. (optional) If your Mailchimp account has 30 merge fields, observe that there may be duplicate merge fields in the audience, and that not all of the metadata fields have been created as merge fields.
6. Check out this branch.
7. In a new session, log into the account that you created in step 3. Confirm that full metadata is now synced for the contact in Mailchimp.

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

PR #1365, opened 2023-11-29
fix(ras): fall back to RAS audience ID if not passed to add_contact method

dkoo:
### All Submissions:

Plugs another hole in the RAS Mailchimp reader data sync. #1355 almost did it but I misunderstood and removed the line that fixed things completely in the current implementation.

For future reference, the longer-term fix should be to refactor the WP hook callbacks that are currently handling syncing upon reader login into Data Events handlers in the Mailchimp and AC Data Connector classes.

1. On `master` or `release`, set up RAS, connect Mailchimp as the ESP and set an audience in RAS advanced settings.
2. As a new reader, purchase a new non-donation subscription without signing up for any newsletter lists.
3. In Mailchimp, find the contact and observe that the [metadata fields](https://github.com/Automattic/newspack-plugin/blob/master/includes/plugins/class-newspack-newsletters.php#L26) are missing data about the subscription (start date, amount, etc.)
4. Test the login sync with error logging in a `wp shell` session. Use the following:

```bash
wp> $customer = new WC_Customer( 111 ); // use the ID of the user account created after purchasing the subscription
wp> $order = $customer->get_last_order();
wp> $contact = Newspack\WooCommerce_Connection::get_contact_from_order( $order );
wp> \Newspack_Newsletters_Subscription::add_contact( $contact );
```

6. This will log the response from the contact sync. Observe that it's an error message:

```
=> object(WP_Error)#11243 (3) {
  ["errors"]=>
  array(1) {
    ["newspack_newsletters_mailchimp_list_id"]=>
    array(1) {
      [0]=>
      string(16) "Missing list id."
    }
  }
  ["error_data"]=>
  array(0) {
  }
  ["additional_data":protected]=>
  array(0) {
  }
}
```

7. Check out this branch, repeat step 4. Confirm that it now succeeds and that the contact in Mailchimp now contains metadata about the subscription, including the following fields:
  - `NP_Total Paid`
  - `NP_Last Payment Amount`
  - `NP_Last Payment Date`
  - `NP_Next Payment Date`
  - `NP_Billing Cycle`
  - `NP_Recurring Payment`
  - `NP_Product Name`

8. Repeat steps 2-3 with a new reader/purchase and confirm that the metadata fields are now synced upon purchase to begin with.

### Other information:

### DIFF
--- a/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php
+++ b/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php
@@ -1054,6 +1054,10 @@ final class Newspack_Newsletters_Mailchimp extends \Newspack_Newsletters_Service
 	 * @return array|WP_Error Contact data if it was added, or error otherwise.
 	 */
 	public function add_contact( $contact, $list_id = false ) {
+		// If RAS is available and a default audience is set.
+		if ( false === $list_id && class_exists( 'Newspack\Reader_Activation' ) && method_exists( 'Newspack\Data_Events\Connectors\Mailchimp', 'get_audience_id' ) ) {
+			$list_id = \Newspack\Data_Events\Connectors\Mailchimp::get_audience_id();
+		}
 		if ( false === $list_id ) {
 			return new WP_Error( 'newspack_newsletters_mailchimp_list_id', __( 'Missing list id.' ) );
 		}


includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php
```diff
@@ -1054,6 +1054,10 @@ public function process_link( $processed, $original ) {
 	 * @return array|WP_Error Contact data if it was added, or error otherwise.
 	 */
 	public function add_contact( $contact, $list_id = false ) {
+		// If RAS is available and a default audience is set.
+		if ( false === $list_id && class_exists( 'Newspack\Reader_Activation' ) && method_exists( 'Newspack\Data_Events\Connectors\Mailchimp', 'get_audience_id' ) ) {
+			$list_id = \Newspack\Data_Events\Connectors\Mailchimp::get_audience_id();
+		}
```
miguelpeixe, line 1060: Back to the original point of removing the line from #1355, this would make every contact subscribed to an audience and receive emails they didn't sign up for. This can result in a pretty messy database to deal with later on.

The [sync made through Mailchimp connector](https://github.com/Automattic/newspack-plugin/blob/fcad059abbe0314ba8a6f3eaaf92d5c96350e248/includes/data-events/connectors/class-mailchimp.php#L135-L162) creates the contact as transactional if it's a new or unsubscribed contact, which is the way Mailchimp allows a contact to exist without being subscribed.

miguelpeixe (changes requested):

includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php
dkoo: Thanks for clarifying the difference between this and the `put` method in the Mailchimp connector. As discussed offline, a quick fix (f176abb) would be to simply use this `put` method if we need to add the contact as a transactional contact. The better long-term solution would be to move all of this to the connector.

miguelpeixe: Feels this method is doing a lot more than it should. Have you considered staying in Newspack Plugin and replace `	\Newspack_Newsletters_Subscription::add_contact( $contact );` with `\Newspack\Data_Events\Connectors\Mailchimp::put();` [here](https://github.com/Automattic/newspack-plugin/blob/fcad059abbe0314ba8a6f3eaaf92d5c96350e248/includes/reader-revenue/woocommerce/class-woocommerce-connection.php#L360)? This is remaining method that rely on `add_contact` and is used by the hooks on `woocommerce_payment_complete` and `woocommerce_subscription_status_updated`.

dkoo: Hmm, that `sync_reader_from_order` method is ESP-agnostic, so we would introduce a condition to call `put` if the ESP is MC and continue with the regular `add_contact` call for any other ESP. It does seem like it will require less code, but what's your opinion on that?

miguelpeixe: Hm... I completely missed that...

I'm not a fan of how it's looking right now because it adds more back and forth to this logic, even though it's meant to be temporary. It goes from the plugin (hook), to newsletters (add_contact), then back to the plugin (Mailchimp::put).

Given that, I believe the cleaner temporary approach is still to leave Newspack Newsletters out of it and do a conditional check on the main plugin to use Mailchimp::put. WDYT?

dkoo: Sounds good—I'll roll it into the [other hotfix](https://github.com/Automattic/newspack-plugin/pull/2780)

PR #1370, opened 2023-12-01
fix: detect and ignore duplicate merge fields

dkoo:
### All Submissions:

Duplicate merge fields were an issue for our data sync. This will detect merge fields with the same name and delete the ones that were created more recently (with a higher ID).

1. Ensure that your site is connected to Mailchimp + RAS, and `define( 'NEWSPACK_LOG_LEVEL', 2 );` is defined in wp-config.php
2. In your Mailchimp account > **Audience > All contacts > Settings > Audience fields and MERGE tags** settings, manually create a merge field with the same name as an existing one that we sync for RAS (see https://github.com/Automattic/newspack-plugin/blob/master/includes/plugins/class-newspack-newsletters.php#L26)
  - Make sure you create a field with a duplicate name (left-most column). So for example, if you already have a merge field with the name `NP_Registration Method`, create another one with the same name and make sure to click "Save" after editing the name to persist the changes.
  - The tag ( `*|MERGE16|*`) will not match as MC enforces uniqueness for this property, but since we're mapping meta fields to merge fields by name, which can repeat, this is what creates the duplication issue.
3. Check out this branch and https://github.com/Automattic/newspack-plugin/pull/2780, then subscribe to a newletter list
4. Confirm that the duplicate field you created in step 2 is logged in your debug log: ` Duplicate merge field <field name> found with tag <field tag>.` Duplicate fields will be ignored for the time being—we will investigate a fix for these and their data in a future PR.

### Other information:

### DIFF
--- a/tests/test-renderer.php
+++ b/tests/test-renderer.php
@@ -97,20 +97,22 @@ class Newsletters_Renderer_Test extends WP_UnitTestCase {
 		);
 
 		// Rich embed as HTML.
-		$inner_html = '<figure><div>https://twitter.com/automattic/status/1395447061336711181</div></figure>';
+		$inner_html       = '<figure><div>https://twitter.com/automattic/status/1395447061336711181</div></figure>';
+		$formatted_string = '<mj-section url="https://twitter.com/automattic/status/1395447061336711181" padding="0"><mj-column padding="12px" width="100%"><mj-text padding="0" line-height="1.5" font-size="16px" ><blockquote>We&#039;re Hiring! We are continuing to grow and have some exciting open positions available, including in Engineering, Product, Marketing, Business Development, HR, Customer Support, and more. Work with us, from anywhere. <a href="https://t.co/EZST4WBsy2">https://t.co/EZST4WBsy2</a> <a href="https://t.co/z8bKfCgn14">pic.twitter.com/z8bKfCgn14</a>&mdash; Automattic (@automattic) <a href="https://twitter.com/automattic/status/1395447061336711181?ref_src=twsrc%5Etfw">May 20, 2021</a></blockquote></mj-text></mj-column></mj-section>';
+		$rendered_string  = Newspack_Newsletters_Renderer::render_mjml_component(
+			[
+				'blockName'    => 'core/embed',
+				'attrs'        => [
+					'url' => 'https://twitter.com/automattic/status/1395447061336711181',
+				],
+				'innerBlocks'  => [],
+				'innerContent' => [],
+				'innerHTML'    => $inner_html,
+			]
+		);
 		$this->assertEquals(
-			Newspack_Newsletters_Renderer::render_mjml_component(
-				[
-					'blockName'    => 'core/embed',
-					'attrs'        => [
-						'url' => 'https://twitter.com/automattic/status/1395447061336711181',
-					],
-					'innerBlocks'  => [],
-					'innerContent' => [],
-					'innerHTML'    => $inner_html,
-				]
-			),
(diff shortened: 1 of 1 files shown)


includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php
```diff
@@ -1078,30 +1078,36 @@ public function add_contact( $contact, $list_id = false ) {
...
+					function( $a, $b ) {
+						return $a['merge_id'] - $b['merge_id'];
+					}
+				);
+
+				$list_merge_fields = [];
+
+				// Handle duplicate fields.
+				foreach ( $existing_merge_fields as $key => $field ) {
+					if ( ! isset( $list_merge_fields[ $field['name'] ] ) ) {
```
adekbadek, line 1097: I could not reproduce based on the instructions provided. Isn't this here a constant condition? `$list_merge_fields` is defined above as an empty array, so `! isset( $list_merge_fields[ $field['name'] ] )` will always be true.

adekbadek (changes requested):

includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php
dkoo: That's intended. Here's the breakdown of the logic:

1. We [fetch](https://github.com/Automattic/newspack-newsletters/pull/1370/files/f25ec1e0a1e60bafff5140e28ab55c2b046ce425#diff-961258e170b9bfe6e81621bab727a9a65947cd67cc016bf5ce1926dd8c78d77cR1085) all merge fields from the ESP.
2. We [sort](https://github.com/Automattic/newspack-newsletters/pull/1370/files/f25ec1e0a1e60bafff5140e28ab55c2b046ce425#diff-961258e170b9bfe6e81621bab727a9a65947cd67cc016bf5ce1926dd8c78d77cR1086) the fields in order of ascending `merge_id`. MC increments the ID so it's a decent approximation of when fields were created relative to each other.
3. We [add](https://github.com/Automattic/newspack-newsletters/pull/1370/files/f25ec1e0a1e60bafff5140e28ab55c2b046ce425#diff-961258e170b9bfe6e81621bab727a9a65947cd67cc016bf5ce1926dd8c78d77cR1097) the fields to `$list_merge_fields` (which is essentially an array of the fields we actually want to sync data to), keyed by field name.
4. Because of the `! isset( $list_merge_fields[ $field['name'] ] )`, if we encounter another field with the exact same name (the issue we're trying to correct for), we do not add this field to `$list_merge_fields`. Instead, since it's a duplicate field, we [delete](https://github.com/Automattic/newspack-newsletters/pull/1370/files/f25ec1e0a1e60bafff5140e28ab55c2b046ce425#diff-961258e170b9bfe6e81621bab727a9a65947cd67cc016bf5ce1926dd8c78d77cR1100) the field in the ESP.

Maybe the testing instructions aren't as detailed as they need to be. I've amended them with a bit more detail so hopefully it's easier to replicate the duplicate field handling.

miguelpeixe (approved):
The merge fields are now being fetched correctly in order to be parsed through the incoming metadata (without needing a name field).

I can also see the duplicate warnings in my log after manually adding a duplicate merge field to my Mailchimp audience.

dkoo:
Thanks for the review! Holding for deployment until Monday.

includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php
adekbadek: Thank you for the thorough response, this is much clearer now. My comment about the constant condition was misguided, I failed to notice that the array will be updated during the iteration 🙈

PR #1377, opened 2023-12-06
fix: store and recover lists on membership reactivation

leogermani:
### All Submissions:

Currently, every time a user is granted a membership, they get add to all lists behind that membership.

If a membership is paused and then restored, the user will be re-added to all lists, even if they opt out of some of the lists at some point.

We recently found out that every membership that is tied to a Subscription, will be put on pause and active again when the subscription is being renewed. This could amplify this problem.

What this PR does is:

* When a membership is deactivated, we store which of the lists behind the membership the user is actually subscribed to
* Whenever a membership is reactivated, we will only resubscribe the user to the lists they were before (which could be none)

1.Create a membership plan
2. Configure Newsletters with 2 or more lists
3. In the membership plan, configure Restricted Content, and select 2 "Subscription Lists"
4. Set the membership to be connected to a Subscription product
5. As a reader, buy the subscription
6. Confirm the reader is granted the membership
7. Confirm in the ESP that the reader was added to both lists
8. As the reader, go to My Account and unsubscribe from one of the lists
9. Confirm it worked
10. Now, process the subscription renewal (one good way of doing this, that simulates the real automatic renewal, is to find the scheduled action under Tools > Scheduled Actions and Run it - search by the subscription ID)
11. After the renewal is complete, confirm in the membership activity that it was set to paused and active again
12. Confirm that the user is still subscribed only to the one list they were subscribed before
13. Ideally, repeat the test leaving all lists selected and no lists selected

### Other information:

---
  - https://app.asana.com/0/0/1206093655507511

### DIFF
--- a/includes/plugins/class-woocommerce-memberships.php
+++ b/includes/plugins/class-woocommerce-memberships.php
@@ -164,11 +164,13 @@ class Woocommerce_Memberships {
 		 * in which case we want to resubscribe them to the lists they were in.
 		 */
 		$current_user_lists = \Newspack_Newsletters_Subscription::get_contact_lists( $user_email );
+		$existing_lists     = [];
 
 		if ( is_array( $current_user_lists ) ) {
 			$existing_lists = array_values( array_intersect( $current_user_lists, $lists_to_remove ) );
-			self::update_user_lists_on_deactivation( $user->ID, $user_membership->get_id(), $existing_lists );
 		}
+
+		self::update_user_lists_on_deactivation( $user->ID, $user_membership->get_id(), $existing_lists );
 
 		if ( ! empty( $provider ) ) {
 			$provider->update_contact_lists_handling_local( $user_email, [], $lists_to_remove );
--- a/includes/plugins/class-woocommerce-memberships.php
+++ b/includes/plugins/class-woocommerce-memberships.php
@@ -20,6 +20,15 @@ defined( 'ABSPATH' ) || exit;
 class Woocommerce_Memberships {
 
 	/**
+	 * The user meta key that stores the list IDs that the user is subscribed to when their memberships are deactivated.
+	 *
+	 * This is used to restore the user's subscriptions when their membership is reactivated.
+	 *
+	 * @var string
+	 */
+	const SUBSCRIBED_ON_DEACTIVATION_META_KEY = 'np_newsletters_lists_on_membership_deactivation';
+
+	/**
 	 * Holds the ID of the user in the scope of the current request.
 	 *
 	 * When a user is granted membership to a plan that requires a registration, the user is not logged in yet,
@@ -29,14 +38,14 @@ class Woocommerce_Memberships {
 	 */
 	protected static $user_id_in_scope;
 
-	/** 
+	/**
 	 * Initialize the class
 	 */
 	public static function init() {
 		add_action( 'plugins_loaded', [ __CLASS__, 'init_hooks' ] );
 	}
-	
-	/** 
+
+	/**
 	 * Initialize the hooks after all plugins are loaded
 	 */
 	public static function init_hooks() {
@@ -81,7 +90,7 @@ class Woocommerce_Memberships {
 
 				return wc_memberships_user_can( $user_id, 'view', [ 'post' => $list_object->get_id() ] );
 
-			} 
+			}
 		);
 		return array_values( $lists );
 	}
@@ -124,7 +133,7 @@ class Woocommerce_Memberships {
 		if ( in_array( $new_status, $status_considered_active ) ) {
 			return;
 		}
-		
+
 		$lists_to_remove = [];
 		$user            = $user_membership->get_user();
 		if ( ! $user ) {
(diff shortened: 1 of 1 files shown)


includes/plugins/class-woocommerce-memberships.php
```diff
@@ -207,6 +235,14 @@ public static function add_user_to_list( $plan, $args ) {
 			}
 		}
 
+		/**
+		 * In case the user was previously removed from the lists, we want to resubscribe them only to the lists they were in.
+		 */
+		$pre_existing_lists = self::get_user_lists_on_deactivation( $user_id, $user_membership->get_id() );
+		if ( ! empty( $pre_existing_lists ) ) {
+			$lists_to_add = array_intersect( $lists_to_add, $pre_existing_lists );
```
miguelpeixe, line 243: Non-blocking suggestion, as there's no need to keep that after the subscription has been restored:

```suggestion
			$lists_to_add = array_intersect( $lists_to_add, $pre_existing_lists );
			delete_user_meta( $user_id, self::SUBSCRIBED_ON_DEACTIVATION_META_KEY );
```

Any time this happens again the meta should have the most recent subscription information, so it's also safer to not keep it in the database.

miguelpeixe (approved):

includes/plugins/class-woocommerce-memberships.php
leogermani: If a membership switch status from a status considered active to another status considered active, we would end up subscribing them to all lists. For example, if you switch from `free_trial` to `active`. Or even `pending` is considered active. See `WC_Memberships_User_Memberships::get_active_access_membership_statuses()`

leogermani: But we need to make sure to update it when the membership is set to inactive. Looking at the code now, it seems tere could be some situations where we would keep outdated values there, let me have a closer look

miguelpeixe: So if they unsubscribe and switch to another "active" status, they'll get re-subscribed to the list they've previously opted out?

leogermani: No... that's what we are trying to avoid

miguelpeixe: > Looking at the code now, it seems tere could be some situations where we would keep outdated values there, let me have a closer look

This is what I meant. If they unsubscribe and the subscription status changes from "active" to "active", they'll get resubscribed.

leogermani: what I'm saying is I think we need to remove this check _> https://github.com/Automattic/newspack-newsletters/pull/1377/files#diff-257035a86886b24eaffe64a9bbaf2b138a59b755e71ebf379317a588e7a28b5fR170

And store an empty list in case they are not subscribed to any list.

This is to cover the case in which they:
* are granted a membership
* the membership is set inactive (stores the current lists -> say they have one)
* the membership is set to active again -> resubscribe to one list
* They unsubscribe from that one list
* the membership is set o inactive again (with the current check, we would keep the old value there, with one list)
* the membership is set to active again (if the old meta is there, they would get subscribed to that one list again)

miguelpeixe: > if the old meta is there, they would get subscribed to that one list again

this is what I'm trying to say 😅

leogermani: https://github.com/Automattic/newspack-newsletters/pull/1377/commits/bae7c857e22a63e1face44a797a0e7f607000736

It's basically what you were suggesting, but doing that when the membership is set to inactive. Always update the info.

Actually. Also need this -> https://github.com/Automattic/newspack-newsletters/pull/1377/commits/e149a74b2065715b6149790188c4d6df90e27770

miguelpeixe: Are there no cases in which the membership goes from "active" to "active"? If the meta is only updated when they are removed from a list, the "active" to "active" switch would also trigger a potentially outdated value.

leogermani: hmm you're right. Thinking...

miguelpeixe: Can't the method bail if the previous status is "active"?

leogermani: We don't have the previous status in that context. And it took me a long while to figure out what hooks to use, I didn't want to explore changing it to another hook

miguelpeixe: An additional user meta can store the last known membership status so you can check against that.

leogermani: status transitions run before `save_post` which is where the other hook is triggered. This seems to work on my tests:

https://github.com/Automattic/newspack-newsletters/pull/1377/commits/3b3d96c2a1c5dff44ecb25fd705eda1891b23a01

Additional testing instructions

* add user to membership
* as the user, unsubscribe from a list
* change the membership from an active state to another active state
* see that nothing is done

miguelpeixe: I have not tested it yet, but how does the `$previous_statuses` is populated when switching from "active" to "active" if it's happening inside the `remove_user_from_list()` method, which is not called in that context?

miguelpeixe: Nevermind, it is called but bails [here](https://github.com/Automattic/newspack-newsletters/blob/3b3d96c2a1c5dff44ecb25fd705eda1891b23a01/includes/plugins/class-woocommerce-memberships.php#L142-L144)

includes/plugins/class-woocommerce-memberships.php
```diff
@@ -117,6 +133,10 @@ function ( $list_id ) use ( $list_ids ) {
 	 */
 	public static function remove_user_from_list( $user_membership, $old_status, $new_status ) {
```
miguelpeixe, line 134: Given that this method doesn't strictly remove a user from a list, I think it'd be good to change its name to something like `handle_membership_status_change`.

miguelpeixe (changes requested):

includes/plugins/class-woocommerce-memberships.php
leogermani: Yes, maybe that method deserved a different name now... ?

includes/plugins/class-woocommerce-memberships.php
leogermani: Looking good: https://github.com/Automattic/newspack-newsletters/pull/1377/commits/1082a5a6cb27696bec2301523c5d6c5d2774d0cc

miguelpeixe (approved):
Working flawlessly! Thank you for the revisions!

leogermani:
> Working flawlessly! Thank you for the revisions!

Thank you for triggering this discussion

PR #1380, opened 2023-12-12
fix(renderer): ensure group blocks use conditional tags

miguelpeixe:
### All Submissions:

1200550061930446-as-1205996773681761

Group blocks placed in the root of the body are processed outside the usual `render_mjml_component()` method so they didn't have the conditional tags properly applied.

This PR ensures the conditional tags are rendered for such blocks.

1. Make sure you have your ESP setup to Mailchimp
2. While in the master branch, draft a new newsletter
3. Add a group block with some content and set the conditional tags to any rule, like `*|IF:FNAME|*` and `*|END:IF|*`
4. Add a paragraph block below that one and apply the same rule
5. Save the draft and click to preview
6. Confirm only the paragraph is contained in the conditional tags
7. Check out this branch, save the draft and preview again
8. Confirm both blocks are contained in the conditional tags

### Other information:

### DIFF
--- a/includes/class-newspack-newsletters-renderer.php
+++ b/includes/class-newspack-newsletters-renderer.php
@@ -104,9 +104,9 @@ final class Newspack_Newsletters_Renderer {
 
 	/**
 	 * Get a value for an image alt attribute.
-	 * 
+	 *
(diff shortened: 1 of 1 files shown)


leogermani (approved):
Works well!

PR #1381, opened 2023-12-14
fix(tracking): do not strip UTM params from proxied redirect URL

dkoo:
### All Submissions:

Enabling click tracking passes destination URLs through a proxied URL in order to capture the click. In this process, any params attached to the the destination URLs get stripped. At the very least, we should allow all UTM params to pass through, but perhaps we can safely allow any arbitrary params to pass through? In case that's a security concern, I've opted in this PR to allow only UTM params plus any params intentionally whitelisted via a new `newspack_newsletters_allowed_url_params` filter.

1. On `master` or `release`, enable tracking in Newsletters settings. Create a newsletter containing a link to your site with some UTM params and arbitrary test params appended, e.g. `https://site-url.com/post-name/?utm_medium=email&utm_campaign=Newspack-Campaign-2023&utm_term=0_-0000000-%5BLIST_EMAIL_ID%5D&arbitrary_param=test-value`
2. Send the newsletter, then click on the link from the email. Observe that only `utm_medium=email` is passed through to the destination URL, and all other params are stripped.
3. Check out this branch and click the link again. This time confirm that all `utm_*` params are passed through to the destination, but any arbitrary params are still stripped.
4. Temporarily add a filter callback somewhere to whitelist the test param, e.g.

```php
\add_filter(
    'newspack_newsletters_allowed_url_params',
    function( $allowed_params ) {
        $allowed_params[] = 'arbitrary_param';
        return $allowed_params;
    }
);
```

5. Click the link again and confirm that the whitelisted params are passed through to the destination.

### Other information:

---
  - https://app.asana.com/0/0/1206176822900759

### DIFF
--- a/includes/tracking/class-click.php
+++ b/includes/tracking/class-click.php
@@ -138,20 +138,33 @@ final class Click {
 			return;
 		}
 
-		// phpcs:disable WordPress.Security.NonceVerification.Recommended
-		$newsletter_id = \intval( $_GET['id'] ?? 0 );
-		$email_address = \sanitize_email( $_GET['em'] ?? '' );
-		$url           = \sanitize_text_field( \wp_unslash( $_GET['url'] ?? '' ) );
-		// phpcs:enable
+		$params        = filter_input_array( INPUT_GET, FILTER_SANITIZE_FULL_SPECIAL_CHARS );
+		$newsletter_id = intval( $params['id'] ?? 0 );
+		$email_address = \sanitize_email( $params['em'] ?? '' );
+		$redirect_url  = \sanitize_text_field( \wp_unslash( $params['url'] ?? '' ) );
 
-		if ( ! $url || ! \wp_http_validate_url( $url ) ) {
+		/**
+		 * Allow additional URL params to be passed through to the redirect URL.
+		 * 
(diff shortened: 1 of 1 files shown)


miguelpeixe (changes requested):
The `$params` in this case are not the destination URL query arguments, they are the [redirection configuration parameters](https://github.com/Automattic/newspack-newsletters/blob/5dbe02a078e9c1817660d8ed36fa633fc22f26a4/includes/tracking/class-click.php#L75-L79).

The `arbitrary_param` in the example (or any other) will not be part of the params you are proposing to filter. This is how the variable looks:

```php
$params = [
	np_newsletters_click => 1, // Query var that determines it's a newsletter click.
	id => 123, // Newsletter ID.
	url => 'https://site-url.com/post-name/?utm_medium=email&utm_campaign=Newspack-Campaign-2023&utm_term=0_-0000000-%5BLIST_EMAIL_ID%5D&arbitrary_param=test-value', // Destination URL.
	em => '*|EMAIL|*', // To be transformed by the ESP
];
```

The click tracking never strips or modifies the query arguments of the original URL. If they are not going through, there might be another issue. Perhaps related to encoding?

dkoo:
Hmm, this may be a red herring—will investigate further

PR #1421, opened 2024-02-05
fix: activate newsletter subscribers if needed

claudiulodro:
### All Submissions:

Closes an issue recently encountered on a client site. When someone gets added to a newsletter list (premium or otherwise), they should have the Subscribed status in MailChimp. This is because MC [categorizes newsletters as a form of marketing email](https://mailchimp.com/help/about-non-subscribed-contacts/). So if someone subscribes to a list but they don't have the Subscribed status, MC won't send them the newsletter because they technically didn't opt-in to marketing emails.

The only place we currently set the status [is here](https://github.com/Automattic/newspack-newsletters/blob/trunk/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php#L1159), and only for new contacts. So if a contact has already been added somehow, they will never get Subscribed status and will never receive a newsletter they signed up for.

I'm not super familiar with the MC API, but my understanding is `status_if_new` is the field we want to use here. Let me know if that's not accurate!

1. Create a "Non-subscribed" contact by going into MC > All Contacts > Add Contacts > Import > Paste and paste an email address into the import. When imported if you did it right you should see they have the "Non-subscribed" status:

Screenshot 2024-02-05 at 2 43 05 PM

2. Before applying this patch, give them a membership or log in as them and change your newsletter preferences. Observe that while they do get added to the correct groups, the status stays "Non-subscribed".
3. Apply this patch. change the newsletter preferences again (or do anything else to trigger a contact sync). Observe the status is now "Subscribed"
Screenshot 2024-02-05 at 3 03 02 PM

### Other information:

### DIFF
--- a/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php
+++ b/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php
@@ -1156,7 +1156,8 @@ final class Newspack_Newsletters_Mailchimp extends \Newspack_Newsletters_Service
 			// If we're subscribing the contact to a newsletter, they should have some status 
 			// because 'non-subscriber' status can't receive newsletters.
 			if ( ! empty( $group_id ) || ! empty( $list_id ) ) {
-				$update_payload['status_if_new'] = $new_contact_status;
+				$update_payload['status_if_new'] = $new_contact_status ?? 'subscribed';
+				$update_payload['status']        = $new_contact_status ?? 'subscribed';
 			}
 
 			// Create or update a list member.
--- a/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php
+++ b/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php
@@ -1153,6 +1153,12 @@ final class Newspack_Newsletters_Mailchimp extends \Newspack_Newsletters_Service
 				];
 			}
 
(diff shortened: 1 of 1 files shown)


adekbadek (approved):

leogermani:
> The only place we currently set the status [is here](https://github.com/Automattic/newspack-newsletters/blob/trunk/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php#L1159), and only for new contacts. So if a contact has already been added somehow, they will never get Subscribed status and will never receive a newsletter they signed up for.
> I'm not super familiar with the MC API, but my understanding is status_if_new is the field we want to use here. Let me know if that's not accurate!

I haven't tested, but doesn't it sound that `status_if_new` will have the same effect that our current code has? Only adding the status if it's  a new contact but not changing the status if it's an existing one?

claudiulodro:
@leogermani the docs are super unclear:

> status_if_new: Subscriber's status. This value is required only if the email address is not already present on the list.

So it _should_ set the subscriber's status in theory. Let me do some testing/research

laurelfulford:
Switching to 'On Hold' for a bit of additional research!

claudiulodro:
I added a little bit to this to make it more "bulletproof", there were a couple edge cases that were still possible before. I'll tag it "Needs Review" again, but probably doesn't need as thorough a review as before :)

leogermani (approved):
Makes sense and tests well!

PR #1425, opened 2024-02-08
fix: only sync once when sending newsletters

chickenn00dle:
### All Submissions:

It looks like we are syncing multiple times whenever we send a newsletter. This is because [we trigger a sync `save_post_{cpt}` hook](https://github.com/Automattic/newspack-newsletters/blob/b5cfee339d660dfe42fc6da8e52ff598b11dfeaf/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php#L50), but, like `save_post`, this seems to be triggered multiple times on a single save. 

In addition to this, we also [trigger a sync on status transition](https://github.com/Automattic/newspack-newsletters/blob/8bbd6f29ba3a60c31c6a3cf37d5665c64d74ac47/includes/service-providers/class-newspack-newsletters-service-provider.php#L346) when [the publisher clicks send](https://github.com/Automattic/newspack-newsletters/blob/b5cfee339d660dfe42fc6da8e52ff598b11dfeaf/src/components/send-button/index.js#L203).

So the result is a lot of syncing whenever a newsletter is saved.

1.
2.
3.

### Other information:

### DIFF
--- a/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php
+++ b/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php
@@ -756,13 +756,19 @@ final class Newspack_Newsletters_Mailchimp extends \Newspack_Newsletters_Service
 	 * @param boolean $update Whether this is an existing post being updated or not.
 	 */
 	public function save( $post_id, $post, $update ) {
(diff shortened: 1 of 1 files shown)


chickenn00dle:
Gonna put this one on hold since it's not the crux of the memory problem we are trying to solve.

PR #1434, opened 2024-02-15
feat: notify site admins on campaign send failure

adekbadek:
### All Submissions:

Notify site admin if a campaign fails to send.

See 1200550061930446-as-1205525012341463/f

1. Set Mailchimp as the ESP
1. Throw a wrench in the works by changing the MC API URL [here](https://github.com/Automattic/newspack-newsletters/blob/dbc9ac1d9b57882b6eae0a91aa4bc3e5a633ba84/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php#L820)
2. Try to send a campaign
3. Observe an email about the failure is sent to the site administrator's email

### Other information:

### DIFF
--- a/includes/service-providers/class-newspack-newsletters-service-provider.php
+++ b/includes/service-providers/class-newspack-newsletters-service-provider.php
@@ -337,24 +337,6 @@ abstract class Newspack_Newsletters_Service_Provider implements Newspack_Newslet
 	}
 
 	/**
-	 * Notify the site administrators.
-	 *
-	 * @param string $subject The email subject.
-	 * @param string $message The email message.
-	 */
-	private static function notify_site_admins( $subject, $message ) {
-		$site_administrators = get_users(
-			[
-				'role'  => 'administrator',
-				'limit' => 10, // Notifying 10 people should be enough.
-			]
-		);
-		foreach ( $site_administrators as $user ) {
-			$sent = \wp_mail( $user->user_email, $subject, $message ); // phpcs:ignore WordPressVIPMinimum.Functions.RestrictedFunctions.wp_mail_wp_mail
-		}
-	}
-
-	/**
 	 * Send a newsletter.
 	 *
 	 * @param WP_Post $post The newsletter post.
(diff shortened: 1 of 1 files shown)


includes/service-providers/class-newspack-newsletters-service-provider.php
```diff
@@ -336,6 +336,24 @@ public function insert_post_data( $data, $postarr ) {
 		return $data;
 	}
 
+	/**
+	 * Notify the site administrators.
+	 *
+	 * @param string $subject The email subject.
+	 * @param string $message The email message.
+	 */
+	private static function notify_site_admins( $subject, $message ) {
```
chickenn00dle, line 345: I think sending these emails as well as selecting how many and which accounts are notified should be tied to some option or hook to allow publishers to opt-in or modify this. Forcing every publisher to start receiving failure emails seems a bit extreme.

includes/service-providers/class-newspack-newsletters-service-provider.php
```diff
@@ -336,6 +336,24 @@ public function insert_post_data( $data, $postarr ) {
...
+		$site_administrators = get_users(
+			[
+				'role'  => 'administrator',
+				'limit' => 10, // Notifying 10 people should be enough.
+			]
+		);
+		foreach ( $site_administrators as $user ) {
+			$sent = \wp_mail( $user->user_email, $subject, $message ); // phpcs:ignore WordPressVIPMinimum.Functions.RestrictedFunctions.wp_mail_wp_mail
```
chickenn00dle, line 353: This is the email I receive when my newsletter fails:

Screenshot 2024-02-27 at 17 03 37

It feels a bit bare and doesn't really provide me with too much information. We might want to include some additional details like:

- Some details about which newsletter failed
- related to the first point, maybe a link to their newsletter or newsletter settings in wp-admin
- Maybe even a support link if they can't figure out how to address this on their end

chickenn00dle (commented):
Hey @adekbadek,

This is working as expected, but I had some concerns around allowing users to opt-in or customize as well as with the email content. Left some comments.

includes/service-providers/class-newspack-newsletters-service-provider.php
adekbadek: Updated in cfa11e6255df7be8427ea4af6e44b11a3bde6ab0 to use the email listed as the `admin_email` option.

includes/service-providers/class-newspack-newsletters-service-provider.php
adekbadek: Updated in cfa11e6255df7be8427ea4af6e44b11a3bde6ab0 to include campaign title and edit link.

chickenn00dle (approved):

PR #1437, opened 2024-02-19
fix: handle inactive woocommerce

adekbadek:
### All Submissions:

Add a check for WC function existence. 

Should be self-explanatory, but if testing is needed – I've encountered this after deactivating `woocommerce` and editing a post. PHP Errors are logged in this situation on `trunk`. 

### Other information:

### DIFF
--- a/includes/plugins/class-woocommerce-memberships.php
+++ b/includes/plugins/class-woocommerce-memberships.php
@@ -79,6 +79,9 @@ class Woocommerce_Memberships {
 		if ( ! is_array( $lists ) || empty( $lists ) ) {
 			return $lists;
 		}
+		if ( ! function_exists( 'wc_memberships_is_post_content_restricted' ) ) {
+			return $lists;
+		}
 		$lists = array_filter(
 			$lists,
 			function ( $list ) {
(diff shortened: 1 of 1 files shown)


includes/plugins/class-woocommerce-memberships.php
```diff
@@ -79,6 +79,9 @@ public static function filter_lists( $lists ) {
 		if ( ! is_array( $lists ) || empty( $lists ) ) {
 			return $lists;
 		}
+		if ( ! function_exists( 'wc_memberships_is_post_content_restricted' ) ) {
```
leogermani, line 82: Hmm so apparently this happens if woo memberships is active but woocommerce is not.

I think it's best to fix the check on the `init` method, that checks `class_exists( 'WC_Memberships_Loader' )`. If Woo and/or Woo memberships is not active, the hooks should not even be initalized

leogermani (changes requested):

includes/plugins/class-woocommerce-memberships.php
adekbadek: Good idea, added in acacf58eba286b3325a5a0b5f3b07a7e4345d5d4

leogermani (approved):

PR #1450, opened 2024-03-05
fix(pixel-tracking): better logfile handling

dkoo:
### All Submissions:

Improves file handling for pixel tracking logs to avoid out-of-memory errors when the logfile grows too large to convert its entire contents to an array via `implode()` or `file()`. The streaming approach only attempts to process up to 100 lines at a time as an array, and then removes the processed lines from the file before reading the next chunk of 100 lines.

Related to (but may not close) `1206709674365921/1206756862605997`.

No functional changes, but should be able to handle much larger pixel logfiles. Enable [pixel tracking](https://help.newspack.com/engagement/newspack-newsletters/#newsletter-tracking) for the Newsletters plugin, and ensure that the logfile system works even with very large files (> 200MB). To test this, you can download a very large logfile (DM me for a sample) and then change [this line](https://github.com/Automattic/newspack-newsletters/pull/1450/files#diff-a5e070b5e885b8924c56d10fcf3aab9296219f8d27dd4dffa795a31cf6e8f358R295) to point to the absolute path in your test environment. Then run `wp shell` and `Newspack_Newsletters\Tracking\Pixel::process_logs()` to process the logfile.

### Other information:

---
  - https://app.asana.com/0/0/1206756862605997

### DIFF
--- a/includes/tracking/class-pixel.php
+++ b/includes/tracking/class-pixel.php
@@ -295,7 +295,7 @@ final class Pixel {
 		$current_log_file = \get_option( 'newspack_newsletters_tracking_pixel_log_file' );
 
 		if ( $current_log_file && file_exists( $current_log_file ) ) {
-			// Read the tracking data from the log file. Process in batches from EOF to avoid memory issues.
+			// Read the tracking data from the log file. Process in batches to avoid memory issues.
 			$handle    = fopen( $current_log_file, 'r' ); // phpcs:ignore WordPress.WP.AlternativeFunctions.file_system_operations_fopen
 			$file_end  = false;
 			$pos       = 0;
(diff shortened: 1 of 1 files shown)


includes/tracking/class-pixel.php
```diff
@@ -260,37 +265,91 @@ public static function log_pixel_url( $url, $post_id, $tracking_id, $email_tag )
...
 				}
-				$item = explode( '|', $item );
-				if ( 3 !== count( $item ) ) {
-					continue;
+
+				// Get the current position in the file after processing the chunk.
+				$pos = ftell( $handle );
+
+				// If we've reached the end of the file, stop the loop.
+				if ( $pos >= $file_size ) {
```
adekbadek: In my testing, this condition is never met, and the `$file_size` is never updated from its initial value. 

Testing steps:
1. Hardcode the value of `$current_log_file` to `<file-name>`
2. Generate a 150-lines logfile: `printf '1|2|hey@email.com\n%.0s' {1..150} > <file-name>`
3. Trigger the logfile processing: `wp eval "\Newspack_Newsletters\Tracking\Pixel::process_logs();"` 

The eval function never exits, and then the `$file_size` is logged, it's never updated. The file is cleared, though. If the `process_logs` execution is halted, then the method called again, the file is removed since the file size is initially `0`.

adekbadek (changes requested):
Encountered an issue with an infinite loop. Might be due to the way I'm testing – let me know if there are any PHP-file-handling caveats I'm missing here.

includes/tracking/class-pixel.php
dkoo: Whoops, I must have lost a `break;` statement in there somewhere. ac4f352 fixes this

adekbadek (approved):
Looking good now 👌

PR #1453, opened 2024-03-08
feat: flag to delete all local lists

dkoo:
### All Submissions:

Small cleanup feature to delete all local lists for a connected ESP. Default behavior only deletes data for lists that are tied to a list in the connected ESP. Passing a truthy value for the new arg will delete all `newspack_nl_list` CPTs.

1. Check out this branch.
2. Connect an ESP in Newspack > Engagement > Newsletters. Enable some subscription lists from the ESP, update the titles and descriptions, and add some new (local) ones.
3. Run in CLI: `wp eval "\Newspack\Newsletters\Subscription_Lists::garbage_collector( [], $esp, true );"` where `$esp` is the slug for your connected ESP (e.g. `mailchimp` or `active_campaign`)
4. Confirm that all of the lists you edited and created in step 2 have been deleted. The remote ones will be recreated automatically but their names should be reset to match the source in the connected ESP.

### Other information:

### DIFF
--- a/includes/class-subscription-lists.php
+++ b/includes/class-subscription-lists.php
@@ -582,16 +582,17 @@ class Subscription_Lists {
 	 *
 	 * @param array  $existing_ids The list of IDs that exist in the ESP. All other remote lists will be deleted.
 	 * @param string $provider_slug The provider slug to clean up lists for. Default is the current configured provider.
+	 * @param bool   $delete_local If true, delete all local lists as well.
 	 * @return void
 	 */
-	public static function garbage_collector( $existing_ids, $provider_slug = null ) {
+	public static function garbage_collector( $existing_ids, $provider_slug = null, $delete_local = false ) {
 		if ( is_null( $provider_slug ) ) {
 			$provider      = Newspack_Newsletters::get_service_provider();
 			$provider_slug = $provider->service;
 		}
 		$all_lists = self::get_all();
 		foreach ( $all_lists as $list ) {
-			if ( ! $list->is_local() && $provider_slug === $list->get_provider() && ! in_array( $list->get_id(), $existing_ids ) ) {
+			if ( ( $delete_local || ! $list->is_local() ) && $provider_slug === $list->get_provider() && ! in_array( $list->get_id(), $existing_ids ) ) {
 				self::delete_list( $list );
 			}
 		}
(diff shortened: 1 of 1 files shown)


chickenn00dle (commented):
Hmm, this doesn't appear to be deleting my added lists. This is what I see before running the cli command:
Screenshot 2024-03-14 at 10 10 09

And this is what I see after running `wp eval "\Newspack\Newsletters\Subscription_Lists::garbage_collector( [], 'mailchimp', true );"`:
Screenshot 2024-03-14 at 10 10 25

dkoo:
@chickenn00dle are those lists coming back because they also exist in the connected MC account? The command will delete all local data, but if the site is still connected to an ESP any lists/audiences/groups, etc. that exist in the ESP will get recreated on your site. (Which is what we want when setting up new staging sites)

chickenn00dle:
> are those lists coming back because they also exist in the connected MC account? The command will delete all local data, but if the site is still connected to an ESP any lists/audiences/groups, etc. that exist in the ESP will get recreated on your site.

Hmm, I created these lists within wp-admin via the `Add New` button in Newspack > Engagement > Newsletters > Subscription Lists. Does this automatically create the list/group in the ESP as well? When I check the ESP (in my case MC) after creating a list via wp-admin newsletters wizard, I don't see a reference to the newly created list. In any case, how can I create a local list that isn't synced with the ESP?

dkoo:
> Hmm, I created these lists within wp-admin via the Add New button in Newspack > Engagement > Newsletters > Subscription Lists. Does this automatically create the list/group in the ESP as well? When I check the ESP (in my case MC) after creating a list via wp-admin newsletters wizard, I don't see a reference to the newly created list. In any case, how can I create a local list that isn't synced with the ESP?

All local lists are synced as Groups to Mailchimp—you should be able to find them under All Contacts > Manage Contacts > Groups > Newspack Newsletters. Try deleting any groups you see there and re-running the CLI command to confirm whether the local lists get cleared out.

chickenn00dle (commented):
> Try deleting any groups you see there and re-running the CLI command to confirm whether the local lists get cleared out.

Thanks! I tried deleting all groups in Mailchimp first, but unfortunately am seeing the same issue. Here is what I see in MC/WPadmin after deleting groups and before running the command:

**Mailchimp**
Screenshot 2024-03-14 at 14 24 08

**WPAdmin**
Screenshot 2024-03-14 at 14 24 14

And after in WPAdmin:
Screenshot 2024-03-14 at 14 24 48

I also tried first disabling the local subscription list and running the command, but it is still not being removed:
Screenshot 2024-03-14 at 14 28 47

chickenn00dle (approved):

PR #1456, opened 2024-03-14
fix: optimistic newsletter form submissions

chickenn00dle:
### All Submissions:

Closes [0/1206709674365921/1206756862605997](https://app.asana.com/0/1206709674365921/1206756862605997/f).

When Reader Activation is set, the newsletter subscription form takes a long time to process. This is because the form is doing alot:

1. Sending a recaptcha request
2. Creating a WP user via reader activation
3. Creating a new contact via the ESP

Altogether, this takes anywhere from 3-7 seconds to complete before the form renders an error or success. This PR attempts to make this feel more immediate for readers by optimistically succeeding, and reverting to error message whenever one is returned from the backend:

https://github.com/Automattic/newspack-newsletters/assets/17905991/1359ecdc-153e-442e-83da-e857edeba2d0

Note that it's fine to default to success here since even if the reader navigates away from the page before the form finishes processing, the user account and contact should still be created on the backend provided there is not an error from ESP.

1. Verify reader activation is all set up on your test site
5. Add a newsletter subscription block to any page
6. Go to the relevant page as a logged out user
7. Input an invalid email address (such as `test@invalid.com`) and submit
8. On `trunk` the form will take anywhere from 3-7 seconds to process, before returning an error message. On this branch, the form will succeed immediately, then render an error after a few seconds.
9. Log out and clear local storage/cookies then repeat the above steps, this time with a valid email address. 
10. Verify you only see a success message.

### Other information:

### DIFF
--- a/src/blocks/subscribe/index.php
+++ b/src/blocks/subscribe/index.php
@@ -502,14 +502,8 @@ function process_form() {
 			'email'    => $email,
 			'metadata' => $metadata,
 		],
-		$lists,
-		true // Async.
+		$lists
 	);
-
-	// The async subscription strategy returns true.
-	if ( true === $result ) {
-		$result = [];
-	}
 
 	if ( \is_wp_error( $result ) ) {
 		return send_form_response( $result );
--- a/src/blocks/subscribe/view.js
+++ b/src/blocks/subscribe/view.js
@@ -65,33 +65,43 @@ domReady( function () {
 		if ( ! form ) {
 			return;
 		}
-		const responseContainer = container.querySelector(
-			'.newspack-newsletters-subscribe__response'
-		);
 		const messageContainer = container.querySelector( '.newspack-newsletters-subscribe__message' );
 		const emailInput = container.querySelector( 'input[type="email"]' );
 		const submit = container.querySelector( 'input[type="submit"]' );
-		form.endFlow = ( message, status = 500, wasSubscribed = false ) => {
+
+		form.onSubmit = ( status = 200 ) => {
 			container.setAttribute( 'data-status', status );
+			// Add success message.
 			const messageNode = document.createElement( 'p' );
+			messageNode.innerHTML = container.getAttribute( 'data-success-message' );
+			messageNode.className = `message status-${ status }`;
+			messageContainer.appendChild( messageNode );
+			// Hide the form.
+			form.style.display = 'none';
+		};
+
+		form.onError = ( message, status = 400 ) => {
+			container.setAttribute( 'data-status', status );
+			// Clear any previous messages.
+			messageContainer.innerHTML = '';
+			// Add error message.
+			const messageNode = document.createElement( 'p' );
+			messageNode.innerHTML = message;
+			messageNode.className = `message status-${ status }`;
+			messageContainer.appendChild( messageNode );
+			// Re-enable the form.
 			emailInput.removeAttribute( 'disabled' );
 			submit.removeAttribute( 'disabled' );
-			messageNode.innerHTML = wasSubscribed
-				? container.getAttribute( 'data-success-message' )
-				: message;
-			messageContainer.appendChild( messageNode );
-			messageNode.className = `message status-${ status }`;
-			if ( status === 200 ) {
-				container.replaceChild( responseContainer, form );
-			}
+			form.style.display = 'initial';
 		};
+
 		form.addEventListener( 'submit', ev => {
 			ev.preventDefault();
 			messageContainer.innerHTML = '';
 			submit.disabled = true;
 
 			if ( ! form.npe?.value ) {
-				return form.endFlow( newspack_newsletters_subscribe_block.invalid_email, 400 );
+				return form.onError( newspack_newsletters_subscribe_block.invalid_email );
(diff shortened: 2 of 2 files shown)


src/blocks/subscribe/index.php
```diff
@@ -502,15 +502,9 @@ function process_form() {
...
 	);
 
-	// The async subscription strategy returns true.
-	if ( true === $result ) {
-		$result = [];
-	}
-
 	if ( \is_wp_error( $result ) ) {
 		return send_form_response( $result );
 	}
```
chickenn00dle, line 510: I have a general question about the behavior of this process_form function here. In it, we first create wp user, and then attempt to create a contact in the ESP. In the event the contact fails to be added, what happens is we have a wp account that has no corresponding ESP contact data. 

Is this the expected behavior? If not, should we flip this around and attempt to create the contact first, only creating a wp user if that is successful. This would prevent us from adding wp accounts for invalid email addresses that aren't accepted by the ESP (test@invalid.com for example)

adekbadek: Good idea! I agree that if adding the contact to the ESP fails, the WP users should not be added. The order of events was intentionally switched in https://github.com/Automattic/newspack-newsletters/pull/1250, but the context there is segmentation. @dkoo – can you weigh in here? Can the segmentation fix from #1250 be achieved while also implementing the change suggested above?

adekbadek (changes requested):
I'm not on board with the UX change here. With the invalid email, the immediate success message suggests that the user was signed up, while in fact they were not. What would be better is a "your submissions is being processed" message, or something along these lines – to let the user know that they can safely navigate away unless they are waiting for the submission result.

chickenn00dle:
> With the invalid email, the immediate success message suggests that the user was signed up, while in fact they were not. What would be better is a "your submissions is being processed" message, or something along these lines – to let the user know that they can safely navigate away unless they are waiting for the submission result.

That's fair. Although I think the long loading time is the reason this issue was opened in the first place and so I'm not sure if adding a loading message is really addressing the problem here. With this approach, I think the majority of users signing up will not fail and so will perceive a quick response time. There definitely is room for confusion for the minority that fail if they navigate away from the page before the error has time to appear though, although I'm not sure that experience is much worse than the existing one in which the form can appear to stall with no messaging at all before a user navigates away.

In any case, I'm happy to switch this up if you feel strongly against the optimistic success!

src/blocks/subscribe/index.php
dkoo: I think this might be a subjective thing, but IMO creating the WP account is more important than the contact data sync. Even if the initial data sync to the ESP fails, if the reader performs another action that would trigger a contact sync (like completing a new transaction, having a subscription renewal or status update, signing up for another newsletter list, unsubscribing from a newsletter list, logging into another browser session etc.) the contact data will be created at the first successful sync and will reflect the reader's most recent state. In other words, the WP reader account should be the source of truth, not the ESP data, since we can always correct ESP data after the fact, but if the WP account isn't created in the first place, the reader's activity won't be persisted in future sessions.

adekbadek:
>  There definitely is room for confusion for the minority that fail if they navigate away from the page before the error has time to appear though, although I'm not sure that experience is much worse than the existing one in which the form can appear to stall with no messaging at all before a user navigates away.

Precisely! That's why I'm suggesting implementing a loading message that would frankly inform the reader about the state of affairs: the form submission is being processed, but we don't have the result yet.

dkoo:
> Precisely! That's why I'm suggesting implementing a loading message that would frankly inform the reader about the state of affairs: the form submission is being processes, but we don't have the result yet.

I think maybe we want to loop in @kevinzweerink or @thomasguillot on this, as it dips into the realm of UX. The purpose of an "optimistic" UI update is to assume that the request will succeed without errors and update the UI immediately to reflect that—only showing an error state when it actually happens. The effect is that if we're confident the request will succeed most of the time, the signup will appear to happen instantaneously to most readers. I think @chickenn00dle's original changes as shown in the video accomplish that. But then yes, what do we do about the readers who navigate away before the error can be shown?

I remember at some point Miguel implemented a feature that cached the error in a transient (or maybe as user meta) so that it could be displayed in a warning modal or other element at the top of the page in future pageloads. The transient would then be cleared when the user either saw it or acknowledged seeing it via a dismissal. Would something like that be a good compromise for this scenario?

adekbadek:
> a feature that cached the error […] so that it could be displayed in a warning modal or other element at the top of the page in future pageloads.

That could work. My concern is a UI that lies by displaying a success message while there's no success. But I'll defer to @Automattic/newspack-design

thomasguillot:
I think it's totally ok to wait a few seconds when submitting any kind of form, as long as the user has some sort of visual feedback.

Subscribe form

thomasguillot:
> My concern is a UI that lies by displaying a success message while there's not success.

I 100% agree with Adam.

dkoo:
Thanks for weighing in, @thomasguillot!

chickenn00dle:
Given the discussion in Asana, I'm gonna go ahead and close this one out.

PR #1460, opened 2024-03-20
hotfix: adding contact to multiple groups within a list

adekbadek:
### All Submissions:

Fixes an issue with adding a contact to multiple interest groups in one go.

This is a bit rough – ideally the `add_contact` would handle the groups without the need for the `add_contact_with_groups` wrapper. 

1. Configure Mailchimp as the ESP and select at least two interest groups (Audience Dashboard -> Manage Audience -> Settings -> Groups). Select these in the Engagement wizard, so they are available for the Newsletter Signup Block
2. Visit a page with the Newsletter Signup Block
3. On `trunk`, check all "lists" and subscribe
4. Observe that an error* is being logged in the server logs, and the contact is added** only to one of the interest groups (or not at all)
5. Switch to this branch, again subscribe to all lists, observe the contact is added to all interest groups in MC

\* "Use PUT to insert or update list members"
\** In my testing MC instance, double-opt-in is required for all contacts, so a real email address will be needed if that's the case on your test instance

### Other information:

### DIFF
--- a/includes/class-newspack-newsletters-subscription.php
+++ b/includes/class-newspack-newsletters-subscription.php
@@ -629,7 +629,7 @@ class Newspack_Newsletters_Subscription {
 			$errors->add( $result->get_error_code(), $result->get_error_message() );
 		}
 
-		// Perform side-effects of lists adding.
+		// Handle local lists feature.
 		foreach ( $lists as $list_id ) {
 			try {
 				$provider->add_contact_handling_local_list( $contact, $list_id );
--- a/includes/class-newspack-newsletters-subscription.php
+++ b/includes/class-newspack-newsletters-subscription.php
@@ -614,29 +614,30 @@ class Newspack_Newsletters_Subscription {
 	 *
 	 * @return array|WP_Error Contact data if it was added, or error otherwise.
 	 */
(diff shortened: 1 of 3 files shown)


dkoo (changes requested):
Fixes the issue as described, but also seems to introduce another regression. If signing up for a local list only, the contact never gets added to the corresponding group. (If signing up for both a local list and a remote group, then it works as expected.)

adekbadek:
Fixed in 6bf25994acbbfc3da4a5f60a0bcaffa90b9cce6b

adekbadek:
I've changed base to `release`, it should have been a hotfix from the start.

dkoo (approved):
Sorry for the delay! This is working—thanks for catching. Is it a regression from recent changes?

adekbadek:
> Is it a regression from recent changes?

The [list-by-list processing](https://github.com/Automattic/newspack-newsletters/commit/815a63049b8abae75d86864d9def37db18a53fd0#diff-dce6057acc32bc5e49a68179cfc17e3e12a57c347768c34779d25668bd8f16beR274-R276) was how the feature was set up in the first place, but initially the plugin did not support interest groups, and it was just extrapolated from there. I think the regression must've started with interest groups support.

PR #1471, opened 2024-04-03
feat(mailchimp): support tags as subscription lists

dkoo:
### All Submissions:

Mailchimp has several different data structures that can be used to divide an audience into what I'll collectively call **sublists**: Groups, Segments, and Tags. Currently, we support only entire audiences or Groups as configurable subscription lists in the Newspack admin.

This PR enables the use of [Mailchimp Tags](https://mailchimp.com/help/getting-started-tags/) as subscription lists. These can be enabled and configured just like Groups and Audiences in the **Newspack > Engagement > Newsletters** page, and Newsletter posts can be sent to them just like with Groups, Audiences, and Segments. The advantage of using Tags over Groups is that unlike Groups, Tags are not visible to anyone outside of the MC admin dashboard, so they're better-suited for creating premium sublists within an audience for which publishers want to control access and visibility.

Also makes some slight improvements to the Newsletter editor experience when using Mailchimp:

-  In current `trunk`, the `newsletterData` meta value which contains Mailchimp audience/group/segment/tag data is generated once when a new newsletter post is created, and then never again for that post. So this means that if any of those items change in the MC account, they're not updated in any preexisting newsletter post drafts. This PR updates the `newsletterData` meta value from the server whenever the post is first loaded in the editor so at least we fetch updated list data on editor load. This shouldn't have any major performance effects because the list data from Mailchimp is already [heavily cached](https://github.com/Automattic/newspack-newsletters/pull/1063) at the server level.
- Allows newsletter posts to be sent directly to segments. On current `trunk` these are fetched with the audience and group data but not shown in the "Send to" dropdown options. The code suggests they should be, even though they're not available to manage as subscription lists. In a future PR, they could be supported as subscription lists just like Tags (as in this PR). It would be a light lift because Tags are a sub-class of Segments, so the data structure is almost the same.
- Tweaks the dropdown list for selecting a Group/Tag/Segment to make use of an existing `SelectControlWithOptGroup` component for better readability. Because HTML only supports one level of nesting with `optgroup` elements, Groups are treated slightly differently from Segments and Tags, but I think it makes sense because only Groups can be sorted into separate categories, and each Group can have exactly one category. So the list is organized like this:
  - **Group Category 1** (also called "Interest Category" in MC)
    - **Group** (also called "Interest" in MC)
  - **Group Category 2**
    - **Group**
  - **Segments**
    - **Segment**
  - **Tags**
    - **Tag**

Screenshot 2024-04-05 at 10 32 18 AM

- If a Group or Tag has been edited locally with a different name than in MC, the dropdown will now display the local name in [brackets] so it's easier to see which sublist you're sending to.
- Segments and Tags now show subscriber counts, similar to how Groups already do.

Also makes a small but significant change to how Premium Newsletters features handle auto-subscriptions to premium lists. On `trunk`, when a user gains membership with a plan that's associated with any subscription lists, they're automatically subscribed to all lists. Now, when the post-checkout newsletter signup UI is enabled via the `NEWSPACK_ENABLE_POST_CHECKOUT_NEWSLETTER_SIGNUP` constant, the reader will now not be auto-subscribed to any lists as they can choose the one(s) they want in this post-checkout UI or in My Account. The exception is for when subscriptions get renewed—in this case, their membership is paused while payment is processing, and then reactivated when payment completes, so we want to seamlessly resubscribe the reader to any lists they were already subscribed to at the time the membership was paused.

1. On a site already connected to Mailchimp on `trunk`, make sure the Mailchimp account has some groups, tags, and segments created and that at least one group is enabled in **Newspack > Engagement > Newsletters**. Edit the name and description to be different from the group's name in Mailchimp.
2. Create a new newsletter and configure it to send to one of the enabled groups. Do not send it, just save it as a draft. (This is so we can confirm that existing drafts won't be affected negatively by the changes in this PR.)
3. Check out this PR and https://github.com/Automattic/newspack-plugin/pull/3035. Either wait for your Mailchimp cache to refresh after 10 minutes, or run `wp cron event run newspack_nl_mailchimp_refresh_cache` at the command line to force a refresh.
4. Visit **Newspack > Engagement > Newsletters** and confirm that you now see your MC account's tags as sublists of your connected audience:

Screenshot 2024-04-05 at 10 41 11 AM

5. Enable the tags, give them new/different names and descriptions and save.
6. Add a Registration and a Newsletter Subscription Form block to a post or page and confirm that the tags you just edited are available for the block. Enable them and publish.

Screenshot 2024-04-05 at 10 42 48 AM

7. On the front-end, visit the post with the Registration and Subscription blocks. Use both to sign up for different lists, including the tags. Confirm that the signups are reflected accurately in:
  - My Account for the reader, after verifying:

Screenshot 2024-04-05 at 10 44 26 AM

  - In the Mailchimp contacts list (the contact should have the corresponding groups and/or tags applied):
Screenshot 2024-04-05 at 10 45 33 AM

8. Back in WP admin, edit the draft newsletter you created all the way back in step 2. Confirm that the dropdown to select a sublist has been updated to include the tags as well as the new organization as described above. Also confirm that the group you had selected before applying the changes in this PR is still selected.
9. Send this newsletter and confirm it gets sent to the proper contacts within that group.
10. Create a second newsletter and send it to a tag. Confirm that it gets sent to the proper contacts with that tag. (Note that if the tag is newly created and didn't have any subscribers until you subscribed in step 7, you may have to force a Mailchimp cache refresh again with `wp cron event run newspack_nl_mailchimp_refresh_cache` before it's selectable to send to)
11. Create a third newsletter and send it to a segment. Confirm that it gets sent to the proper contacts in that segment.

#### Optional: testing premium newsletter and other experimental features

Tags should also work for premium newsletter and post-checkout signup features, just like Groups.

1. Enable post-checkout signup by adding `define( 'NEWSPACK_ENABLE_POST_CHECKOUT_NEWSLETTER_SIGNUP', true );` to wp-config.php.
2. Enable premium newsletters by activation the Woo Memberships plugin and creating a Membership Plan. Set the plan to grant access upon **product(s) purchase** and choose a subscription product on your site, then add a rule under Restrict Content and select "Subscription Lists" and one of your tags here.

Screenshot 2024-04-05 at 10 53 13 AM
Screenshot 2024-04-05 at 10 53 25 AM

3. As a new reader, visit the site and confirm that the Registration and Newsletter Subscription Form blocks no longer show the restricted tag as an option to sign up.
4. Sign up for other non-restricted lists and visit My Account/verify. Confirm that the restricted tag isn't shown here, either.
5. Use the Checkout Button block + modal checkout flow to purchase the product you selected in the Membership Plan in step 2. Confirm that the post-checkout modal screen shows all your newsletters as an option to sign up, including the restricted tag.

Screenshot 2024-04-05 at 10 59 52 AM

6. Keep all selected (or at least the restricted tag) and visit My Account. Confirm that the restricted tag is now shown for this user, now that they meet the membership requirements.
7. As an admin, create a new newsletter and send it to the restricted tag. Confirm that it's sent to the reader who signed up for it with the subscription purchase, as well as any other contacts who have that tag in MC.
8. In WP admin, manually trigger a renewal of the reader's subscription. Confirm that the tag temporarily gets removed from the contact in MC while payment is processing and the subscription is in "On hold" status, but then gets automatically re-added when payment completes and the subscription is reactivated.
9. In WP admin, manually change the reader's subscription to "Cancelled" or "Expired" status. Confirm that the restricted tag is immediately removed from the contact in MC, and that the restricted tag is no longer available to the reader in My Account or in Registration/Newsletter Signup blocks unless their subscription is reactivated to "Active" status.

### Other information:

---
  - https://app.asana.com/0/0/1206759769105465

### DIFF
--- a/includes/class-subscription-lists.php
+++ b/includes/class-subscription-lists.php
@@ -462,13 +462,21 @@ class Subscription_Lists {
 			throw new \Exception( 'Invalid list' );
 		}
 
-		$saved = Subscription_List::from_form_id( $list['id'] );
+		$subscriber_count = ! empty( $list['subscriber_count'] ) ? (int) $list['subscriber_count'] : 0;
+		$subscriber_count = ! empty( $list['member_count'] ) ? (int) $list['member_count'] : 0; // Tags have member_count instead of subscriber_count.
+		$saved_list       = Subscription_List::from_form_id( $list['id'] );
+		if ( $saved_list ) {
+			// Update remote name, in case it's changed in the ESP.
+			$saved_list->set_remote_name( $list['title'] );
 
-		if ( $saved ) {
-			return $saved;
+			// Update subscriber count, if available.
+			if ( 0 > $subscriber_count ) {
+				$saved_list->set_subscriber_count( $subscriber_count );
+			}
+			return $saved_list;
 		}
 
-		return self::create_remote_list( $list['id'], $list['title'] );
+		return self::create_remote_list( $list['id'], $list['title'], null, $subscriber_count );
 	}
 
 	/**
@@ -477,9 +485,10 @@ class Subscription_Lists {
 	 * @param string $remote_id The ID of the list in the ESP.
 	 * @param string $name The name of the list.
 	 * @param string $provider_slug The provider slug to create the list for. Default is the current configured provider.
+	 * @param int    $subscriber_count The number of subscribers in the list, if available.
 	 * @return Subscription_List|WP_Error
 	 */
-	public static function create_remote_list( $remote_id, $name, $provider_slug = null ) {
+	public static function create_remote_list( $remote_id, $name, $provider_slug = null, $subscriber_count = 0 ) {
 		$post_id = wp_insert_post(
 			[
 				'post_type'   => self::CPT,
@@ -494,6 +503,7 @@ class Subscription_Lists {
 
 		$list = new Subscription_List( $post_id );
 		$list->set_remote_id( $remote_id );
+		$list->set_remote_name( $name );
 		$list->set_type( 'remote' );
 		if ( is_null( $provider_slug ) ) {
 			$provider = Newspack_Newsletters::get_service_provider();
@@ -502,6 +512,9 @@ class Subscription_Lists {
 		}
 		if ( ! empty( $provider ) ) {
 			$list->set_provider( $provider->service );
+		}
+		if ( 0 > $subscriber_count ) {
+			$list->set_subscriber_count( $subscriber_count );
 		}
 		return $list;
 	}
--- a/includes/class-subscription-list.php
+++ b/includes/class-subscription-list.php
@@ -53,6 +53,16 @@ class Subscription_List {
 	const REMOTE_ID_META = '_remote_id';
 
 	/**
+	 * The name of the meta key used to store the list's remote name, only present for remote lists
+	 */
+	const REMOTE_NAME_META = '_remote_name';
+
+	/**
+	 * The post meta key used to store the list's subscriber count, only present for Group-based lists.
+	 */
+	const SUBSCRIBER_COUNT_META = '_subscriber_count';
+
+	/**
 	 * Checks if a string $id is in the format of a local Subscription List Form ID
 	 *
 	 * @see self::get_form_id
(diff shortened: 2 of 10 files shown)


laurelfulford (approved):
I ran through the test steps and the optional test steps while watching the logs, and everything looks good to me! I also tried connecting the same site to AC, and didn't see any issues in the To/Segment UI; the latter dropdown has a bunch of label-less options, but that happens on trunk and looks to be something weird with our AC account (probably too much messy test data!).

I really like the updates to the Group/Segment/Tag dropdown -- the optGroup format makes it easier to see the groupings, and it's handy to have the custom names listed there with the MC ones.

dkoo:
@laurelfulford thanks for the review! I'll look into that AC "To" UI issue before merging.

UPDATE: Looks like those blank options are actually unnamed segments in ActiveCampaign. So I think you're right that it's due to our messy test environment.

includes/class-subscription-lists.php
```diff
@@ -462,13 +462,21 @@ public static function get_or_create_remote_list( $list ) {
 			throw new \Exception( 'Invalid list' );
 		}
 
-		$saved = Subscription_List::from_form_id( $list['id'] );
-
-		if ( $saved ) {
-			return $saved;
+		$subscriber_count = ! empty( $list['subscriber_count'] ) ? (int) $list['subscriber_count'] : 0;
+		$subscriber_count = ! empty( $list['member_count'] ) ? (int) $list['member_count'] : 0; // Tags have member_count instead of subscriber_count.
```
leogermani, line 466: @dkoo Isn't this setting the value of `$subscriber_count` even when only `subscriber_count` is present?

Shouldn't it be

```
$subscriber_count = ! empty( $list['member_count'] ) ? (int) $list['member_count'] : $subscriber_count;
```

?

dkoo: Ah yes, you're right—funny how no one has caught this since April!

dkoo: Strangely, it doesn't seem to affect the counts shown in my editor

Screenshot 2024-06-24 at 5 37 56 PM

PR #1502, opened 2024-05-09
feat(ras-acc): add helper method to identify premium lists

chickenn00dle:
### All Submissions:

Closes https://app.asana.com/0/1206943664367857/1205992141743582/f

Related to https://github.com/Automattic/newspack-plugin/pull/3105 and https://github.com/Automattic/newspack-blocks/pull/1727

This PR adds a helper method to identify whether a given list is associated with a membership plan.

There should be no changes in functionality. 

- Smoke tests around newsletter lists tied to membership plans.

### Other information:

### DIFF
--- a/includes/plugins/class-woocommerce-memberships.php
+++ b/includes/plugins/class-woocommerce-memberships.php
@@ -388,6 +388,41 @@ class Woocommerce_Memberships {
 		}
 		return $value[ $membership_id ];
 	}
+
+	/**
+	 * Determines whether a given list id is associated with a membership plan.
+	 *
+	 * @param string $list_id The list ID.
+	 *
+	 * @return bool
+	 */
+	public static function is_membership_list( $list_id ) {
+		if ( ! function_exists( 'wc_memberships_get_membership_plans' ) ) {
+			return false;
+		}
+
+		$plans = wc_memberships_get_membership_plans();
+
+		if ( empty( $plans ) ) {
+			return false;
+		}
+
+		foreach ( $plans as $plan ) {
+			$rules = $plan->get_content_restriction_rules();
+
+			foreach ( $rules as $rule ) {
+				if ( Subscription_Lists::CPT !== $rule->get_content_type_name() ) {
+					continue;
+				}
+				$object_ids = $rule->get_object_ids();
+				if ( in_array( $list_id, $object_ids, true ) ) {
+					return true;
+				}
+			}
+		}
+
+		return false;
+	}
 }
 
 Woocommerce_Memberships::init();
--- a/includes/plugins/class-woocommerce-memberships.php
+++ b/includes/plugins/class-woocommerce-memberships.php
@@ -414,6 +414,7 @@ class Woocommerce_Memberships {
 				if ( Subscription_Lists::CPT !== $rule->get_content_type_name() ) {
 					continue;
 				}
+
 				$object_ids = $rule->get_object_ids();
 				if ( in_array( $list_id, $object_ids, true ) ) {
 					return true;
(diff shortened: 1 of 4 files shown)


tests/mocks/wc-memberships.php
```diff
@@ -0,0 +1,45 @@
+<?php // phpcs:ignoreFile
```
chickenn00dle, line 1: Added a mock for wc-memberships plugin testing. Not sure if this is the best way to approach this though, so let me know if there is a cleaner approach!

includes/plugins/class-woocommerce-memberships.php
```diff
@@ -388,6 +388,42 @@ private static function get_user_lists_on_deactivation( $user_id, $membership_id
...
 	}
+
+	/**
+	 * Determines whether a given list id is associated with a membership plan.
+	 *
+	 * @param string $list_id The list ID.
+	 *
+	 * @return bool
+	 */
+	public static function is_membership_list( $list_id ) {
```
adekbadek: I don't find this method name very helpful. Starting with "is membership" suggests that the subject of this method is a membership, while in fact it's a subscription list. I'd suggest renaming to something more descriptive, like `is_subscription_list_tied_to_plan`.

tests/plugins/test-woocommerce-memberships.php
```diff
@@ -0,0 +1,18 @@
...
+
+namespace Newspack_Newsletters\Plugins;
+
+/**
+ * Test Woocommerce Memeberships.
+ */
+class Woocommerce_Memberships_Test extends \WP_UnitTestCase {
+	public function test_is_membership_list() { // phpcs:ignore Squiz.Commenting.FunctionComment.Missing
+		$this->assertTrue( Woocommerce_Memberships::is_membership_list( 1 ) );
+		$this->assertFalse( Woocommerce_Memberships::is_membership_list( 0 ) );
```
adekbadek: These tests don't make the use of the method clear. After digging into the code, I see that `0` is an id of a non-existing list, so the method returns false. `1` is apparently an id of a list tied to a membership plan, but it's not evident anywhere in the test. 
A test should be able to be read as a component spec – _given these conditions, code does that, etc_. In this test, the subscription list tied to a membership plan should first be set up, then it's id used in the method test.

adekbadek (changes requested):
Glad to see the tests! Left some remarks. IMO mocking is the right way to go, unless the mock becomes too large.

includes/plugins/class-woocommerce-memberships.php
chickenn00dle: Sure thing! I've updated in https://github.com/Automattic/newspack-newsletters/pull/1502/commits/33fae942c8c09f32f60df78e6a49b4976bebd96e

tests/plugins/test-woocommerce-memberships.php
chickenn00dle: I see what you mean. I had a hard time moving the setup of the membership plan into the test since getting the memberships relies on the global `wc_memberships_get_membership_plans` method and as far as I can tell, there isn't a simple way to mock this from within the test without adding some more complexity to the wc-memberships mock.

That said, I did move things around and move the setup into a static helper method to try and make things more clear in https://github.com/Automattic/newspack-newsletters/pull/1502/commits/33fae942c8c09f32f60df78e6a49b4976bebd96e

Let me know if this is good enough or if you think there is another strategy I should take here.

chickenn00dle:
Thanks for the review @adekbadek! Sorry for the delay in addressing the feedback. For some reason I missed the notification on this one.

I tried to address your feedback. Let me know if I need to make further changes. I'm a bit rusty when it comes to PHPUnit so am probably missing some functionality that might make the tests simpler.

adekbadek (approved):

tests/plugins/test-woocommerce-memberships.php
adekbadek: Much better!

PR #1503, opened 2024-05-14
feat: save a backup of subscription attempts in a custom DB table

adekbadek:
### All Submissions:

Saves subscription attempts in a custom DB table. If for any reason a subscription attempt does not reach ESP, and we find out only later, this table will serve as a backup to restore the subscriptions. 

The table will be cleaned weekly of any attempts older than half a year.  

1. Use the Newsletter Subscribe block to subscribe to a newsleter
2. Observe a new DB table called `newspack_newsletters_subscription_attempts` has been created, and that it recorded the subscription attempt

### Other information:

---
  - https://app.asana.com/0/0/1207267351460369

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@
 .DS_Store
 release
 .phpunit.result.cache
+codecov
--- a/includes/class-newspack-newsletters-subscription-attempts.php
+++ b/includes/class-newspack-newsletters-subscription-attempts.php
@@ -30,6 +30,7 @@ class Newspack_Newsletters_Subscription_Attempts {
 		add_action( self::CRON_HOOK, [ __CLASS__, 'cleanup' ] );
 
 		add_action( 'newspack_newsletters_pre_add_contact', [ __CLASS__, 'save_attempt' ], 10, 2 );
+		add_action( 'newspack_newsletters_update_contact_lists', [ __CLASS__, 'save_update' ], 10, 5 );
 	}
 
 	/**
@@ -103,6 +104,22 @@ class Newspack_Newsletters_Subscription_Attempts {
 	}
 
 	/**
+	 * Get a row by email.
+	 *
+	 * @param string $email Email address.
+	 */
+	public static function get_by_email( $email ) {
+		global $wpdb;
+		return $wpdb->get_row( // phpcs:ignore WordPress.DB.DirectDatabaseQuery.DirectQuery, WordPress.DB.DirectDatabaseQuery.NoCaching
+			$wpdb->prepare(
+				'SELECT email,list_ids FROM %i WHERE email = %s',
+				self::get_table_name(),
+				$email
+			)
+		);
+	}
+
+	/**
 	 * Set a value in the database.
 	 *
 	 * @param string $email The reader's unique ID.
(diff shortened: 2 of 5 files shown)


dkoo (changes requested):
EDIT: Many of these points are addressed in internal discussion. Leaving here for posterity

This is working mostly as described, but I wanted to discuss this a bit in the interest of trying not too add more complexity to an increasingly complex plugin codebase.

We already have a concept of "subscription intents" which are created and used as part of the optimistic async subscription request feature (implemented in #1344). Would it be feasible to use the `np_nl_sub_intent` CPT to log all subscription attempts, even the non-async ones, and show these in a WP admin post table as a backup? We could add some post meta as needed to filter these posts by async/successful/failed, etc. attempts. WDYT?

Aside from that, the custom table as implemented has a couple of holes:

- If a contact that already exists in the ESP subscribes to a new list, this is handled by the [`update_contact_lists` method](https://github.com/Automattic/newspack-newsletters/blob/feat/record-signup-attempts/includes/class-newspack-newsletters-subscription.php#L795) which isn't firing the action that logs a new row in the table.
- (Mailchimp only) If a contact is added to a Mailchimp audience but not subscribed to any newsletter lists, this is handled by a [`put` method in the main plugin](https://github.com/Automattic/newspack-plugin/blob/trunk/includes/data-events/connectors/class-mailchimp.php#L201) which doesn't fire the action to log the contact add.
- This method only logs _signup_ attempts. Is there any value in logging _unsubscribe_ attempts as well?
- The `add_contact_to_provider` method fires a [`newspack_newsletters_add_contact` action](https://github.com/Automattic/newspack-newsletters/blob/feat/record-signup-attempts/includes/class-newspack-newsletters-subscription.php#L680) which would let us see whether the signup attempt was actually successful or not. Rather than simply keeping all rows around for 6 months, we could use this action to delete any rows related to positively verified successful signup attempts.

adekbadek:
Thanks for the thorough review! As noted, the intents vs attempts discussion was addressed internally. 

> If a contact that already exists in the ESP subscribes to a new list […]

Done in 2fc5bdafb7330efa757365dc86e9dbbed59ff6c4.

> If a contact is added to a Mailchimp audience but not subscribed to any newsletter lists

That's out of scope of this PR. The use of MC API in the Newspack plugin is different – these are not subscription attempts, but using MC as a CRM. The aim of this PR is to enable recovery of newsletter signup attempts, not of any and all communication with MC API.  

> This method only logs signup attempts. Is there any value in logging unsubscribe attempts as well?

Interesting suggestion, but again – that'd be out of scope of this plugin's functionality. This plugin does not offer means of unsubscribing. 

> Rather than simply keeping all rows around for 6 months, we could use this action to delete any rows related to positively verified successful signup attempts.

I'd rather opt for simplicity here. Implementing this suggestion would optimize at the cost of added complexity – what if the deletion happens by accident? If there's only one place of deletion (the `cleanup` method), it's simpler and less error-prone.

dkoo (approved):
Approving since this seems sufficient for the purposes of retaining a backup resource. Another lightweight alternative in the future could be to use the `Logger` class to log subscription attempts to someplace like a flat file on the server, but I don't really have an opinion on the method of storing the data.

miguelpeixe:
Something to discuss about this approach is that the custom table does not have any indexes, so operations like this one:

```php
$wpdb->query( $wpdb->prepare( 'DELETE FROM %i WHERE created_at < now() - interval 6 MONTH LIMIT 1000', self::get_table_name() ) ); 
```

can get quite heavy to execute.

Simpler operations like the one for `get_by_email()`, called on every new subscription, can also be expensive on a large set.

PR #1513, opened 2024-05-21
feat: add new modern style variation to block

thomasguillot:
### All Submissions:

This PR introduces a new style variation (SV) to the Newsletter Subscription Form block.

This SV is based on our Newspack Components and is designed to match the rest of our reader-facing UI.

subscription-block-eg

1. Switch to this branch
2. Add the block to a page and change the SV to "Modern"
3. Play with the settings (and preview front-end)

### Other information:

### DIFF
--- a/src/blocks/subscribe/block.json
+++ b/src/blocks/subscribe/block.json
@@ -97,12 +97,7 @@
 	"styles": [
 		{
 			"name": "modern",
-			"label": "Modern",
-			"isDefault": true
-		},
-		{
-			"name": "default",
-			"label": "Classic"
+			"label": "Modern"
 		}
 	]
 }
--- a/src/blocks/subscribe/block.json
+++ b/src/blocks/subscribe/block.json
@@ -93,5 +93,16 @@
 	"supports": {
 		"html": false,
 		"align": true
-	}
+	},
+	"styles": [
+		{
+			"name": "modern",
+			"label": "Modern",
+			"isDefault": true
+		},
+		{
+			"name": "default",
+			"label": "Classic"
+		}
+	]
 }
--- a/src/blocks/subscribe/editor.scss
+++ b/src/blocks/subscribe/editor.scss
@@ -1,11 +1,11 @@
 @use '~@wordpress/base-styles/colors' as wp-colors;
 
-.editor-styles-wrapper .newspack-newsletters-subscribe {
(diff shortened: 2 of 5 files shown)


chickenn00dle (changes requested):
The style changes look good, but I am running into an issue where the block is showing modern as the selected style but actually defaulting to classic styles initially until I toggle between the styles one time. Here is a screen recording:

https://github.com/Automattic/newspack-newsletters/assets/17905991/191e3344-6324-4cfa-8bd5-2b83ec65e152

thomasguillot:
> I am running into an issue where the block is showing modern as the selected style but actually defaulting to classic styles initially until I toggle between the styles one time.

@chickenn00dle this is super weird. I'm not able to replicate this bug. Any idea?

chickenn00dle:
>  I'm not able to replicate this bug. Any idea?

Hmm. Another step I missed was this happens when I only have one list selected at first when the block is added to the page, and then I toggle on a second list. This bit is also shown in the video. So for clarity, here are my steps:

1. Add the newsletter subscription block to a page
2. In block settings, select a second list from the Subscriptions Lists menu:
Screenshot 2024-05-22 at 17 38 12
3. the newsletter options appear with classic styles in the block despite modern being the styles selection
4. select the classic style from the block settings
5. select the modern style again from block settings
6. the newsletter options now appear with modern styles as expected

This happens on the latest version of Firefox for me.

thomasguillot:
@chickenn00dle thank you very much. I was able to replicate this issue (I didn't notice it at first because my site was using system font so with one list, there was 0 difference between the 2 SV).

As a "fix" I simply removed `isDefault` from the Modern SV. I'm not quite sure why it's not being respected.

https://github.com/Automattic/newspack-newsletters/pull/1513/commits/b9b6b3475cd3e5560cab9a96fa759540f7b7441e

chickenn00dle (approved):
> As a "fix" I simply removed isDefault from the Modern SV. I'm not quite sure why it's not being respected.

This works!

Although I would suggest opening a new Asana task for digging into changing the default if the intention is to have modern be the new default.

In any case, nice work! Giving this a thumbs up 👍

thomasguillot:
> Although I would suggest opening a new Asana task for digging into changing the default if the intention is to have modern be the new default.

Good idea. We will need to do the same for the Donate block anyway.

PR #1517, opened 2024-05-27
fix(email-rendering): add side margin to list blocks

adekbadek:
### All Submissions:

Adds a bit of side margin to list blocks, to prevent misaligned content:

image

1. Insert a list block in a newsletter, with paragraphs above and below for reference
2. Preview email or send a test email
3. Observe the list is aligned with the paragraps

### Other information:

---
  - https://app.asana.com/0/0/1206910890455747

### DIFF
--- a/includes/email-template-mjml.css
+++ b/includes/email-template-mjml.css
@@ -71,7 +71,7 @@ blockquote,
 .wp-block-quote {
 	background: transparent;
 	border-left: 0.25em solid;
-	margin: 0;
+	margin: 0 0 0 12px;
 	padding-left: 24px;
 }
 blockquote cite,
--- a/includes/email-template-mjml.css
+++ b/includes/email-template-mjml.css
@@ -61,7 +61,7 @@ h6 {
 
 /* List */
 ul, ol {
-	margin: 12px 0;
+	margin: 12px 15px;
 	padding: 0;
 	list-style-position: inside;
 }
--- a/src/editor/style.scss
+++ b/src/editor/style.scss
@@ -210,8 +210,13 @@
 		.wp-block-quote {
 			margin: 0 auto;
 
+			&::before {
+				left: 0 !important;
+			}
+
 			p {
-				margin: 0 0 12px;
+				margin: 0;
+				padding: 8px 0;
 			}
 
 			&__citation {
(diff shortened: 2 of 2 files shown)


chickenn00dle (changes requested):
Nice! Looks much better but I ran into two issues:

1. There's some inconsistencies between what the preview shows and what the test email + editor show:

The editor and test email render the list block with more padding than other types of blocks:

Screenshot 2024-05-28 at 04 14 30

The test email renders the list block with padding that is the same as other blocks (although it seems just slightly off):

Screenshot 2024-05-28 at 04 14 22

2. Quotes have the same issue as list blocks. I'm not sure if you'd like to address this in this PR or create a follow up task to address in another PR:

Screenshot 2024-05-28 at 04 18 12

adekbadek:
Thanks for catching the blockquote issue! Fixed all in 8eee36ec82c164212e1f127291c24642609cecb3

chickenn00dle (changes requested):
There are still some inconsistencies with alignment between the editor, preview, and actual email. Can we get the list and quote blocks to have the same alignments across all three?

Editor:
Screenshot 2024-05-29 at 16 35 44

Preview:
Screenshot 2024-05-29 at 16 36 54

Email:
Screenshot 2024-05-29 at 16 38 20

adekbadek:
Here's what I see:

| Editor | Preview | Mailchimp Email In Gmail
|--------------|-----------|--|
| image | image | image |

Can you confirm if the editor styles were recompiled on your end?

I don't think we can do much about the last case. The ESP and then the email client will always add some styling changes. The main point here is that the content does not get clipped from the side.

chickenn00dle (approved):
> Can you confirm if the editor styles were recompiled on your end?

Ah. Looks like this was the problem. After a clean install and build, I see the changes on my end.

> I don't think we can do much about the last case. The ESP and then the email client will always add some styling changes. The main point here is that the content does not get clipped from the side.

That's fair. Although I would argue if we can't modify the actual email output, we should at least try to get what our editor/preview renders as close to the email as possible since it's a bit misleading to publishers. Either that, or add a general note somewhere (even if its just in docs) that there may be minor differences between the preview and what is actually rendered in emails.

In any case, I'll go ahead and approve since this addresses the problem of the side clipping.

PR #1526, opened 2024-06-07
fix: refresh email MJML + HTML after save, and refresh for preview without saving

dkoo:
### All Submissions:

Refactors how we refresh the MJML + email-compliant HTML for a newsletter upon save:

#### Current

We hook into the `apiFetch` method itself with a custom middleware. The intention is for the middleware to intercept save requests and do the following: If the request method is `POST` or `PUT` and the request body contains `content` and `id` params, we:

- Trigger another save request to persist edited post meta before the "actual" save
- Generate MJML from post data on the server via REST API
- Generate email-compliant HTML from the regenerated MJML via `mjml2html` API
- Save the generated HTML to post meta via another REST API request
- Finally, proceed with the original request

Weaknesses:

- With the only conditions being the request method + required `content` and `id` params, this middleware is running much more often than intended. Particularly, it runs on autosave as well as manual saves, and could potentially run on any `POST` or `PUT` request that happen to have the required params in the request body.
- Since it runs on autosave, it creates a sort of race condition if multiple instances of the same newsletter are open in the post editor at the same time (either in the same browser or even different devices). A long-dormant tab with the editor open can continue to generate autosave requests that might end up regenerating MJML+HTML using old post content and metadata, which could overwrite newer HTML that was intentionally saved.
- Triggers three extra REST API requests + one async `mjml2html` request for every save and autosave request!
- If any of these extra middleware API requests fail, there's currently no error handling so the save request itself will ultimately fail as well.

#### Proposed

Instead of hooking into `apiFetch`, we instead watch for save and autosave state changes from the WP core editor's data store. If the editor transitions from an `isSavingPost` state to a non-`isSavingPost` state, while not also autosaving, AND `didPostSaveRequestSucceed()` returns true, that indicates that the editor has successfully completed a manual save as a result of either hitting the "Save" button or dispatching a `savePost()` request. If these conditions are met, we then regenerate MJML, then HTML, then save the HTML to post meta.

Improvements:

- MJML and HTML are refreshed and persisted only after an intentional save based on the WP editor's state, never on autosave or any other type of request.
- Much less risk of multiple editor instances overwriting data—an open editor instance won't regenerate MJML+HTML unless someone actually clicks "Save" in that instance.
- Reduces the number of extra requests to 2 REST API requests + 1 `mjml2html` request! Not ideal still, but 25% better than before.
- The extra requests are run after the save request instead of during, so even if they fail they won't affect the post's data integrity.
- The newsletter data is now synced to the ESP whenever the `newspack_email_html` meta value is updated—which we can be 100% certain happens only as a result of the above flow.
- Bonus: we've moved the requests to regenerate MJML+HTML to a standalone util function. Now that we've untethered those requests from the save request itself, we can now regenerate email-compliant HTML for previewing without persisting it—so the "Preview" button is now a true preview that doesn't automatically dispatch a `savePost()` request, AND it's much faster as a result.

1. On `trunk`, create a new newsletter and edit the post content. Save.
2. Open the same newsletter in a different tab (Tab B) and don't touch it.
3. In the first tab (Tab A), edit the post content a bunch more, save the post, then close Tab A while leaving Tab B open.
4. Go take a break and get a coffee for ~10 minutes or however long it takes for the editor in Tab B to autosave.
6. Use WP CLI to inspect the saved HTML: `wp post meta get <post ID> newspack_email_html`
7. Observe that it matches the content from Tab B despite Tab A having saved content more recently. (Alternatively, refresh the editor and click "Preview"—the rendered preview may or may not match the actual saved post content.)
8. Edit the post a bunch more (content and other sidebar options) and click "Preview" without saving. Using WP CLI, observe that the unsaved changes were actually saved when you clicked "Preview".
9. Check out this branch and repeat all steps. This time, confirm that the `newspack_email_html` meta value and the rendered preview content always match the post at the moment of the last time you hit "Save" in any tab.
10. Also confirm in your connected ESP that the synced campaign always matches the post at the moment you last hit "Save".
11. For good measure, send a test email and then a live send of the newsletter, and confirm that the sent email content is up-to-date.

### Other information:


dkoo:
I can't think of a way to add unit test coverage to this feature as it stands right now, because it relies so much on handoffs between JS and PHP APIs. Any ideas on that, @adekbadek or @claudiulodro?

src/newsletter-editor/editor/index.js
```diff
@@ -49,153 +56,234 @@ const Editor = compose( [
...
+		postTitle,
+		postType,
+		sent,
+		successNote,
+		updateMetaValue,
+	} ) => {
+		const [ publishEl ] = useState( document.createElement( 'div' ) );
+
+		// TODO: on first editor load, refresh HTML and set as edited post meta.
+		// const response = mjml2html( mjml, { keepComments: false, minify: true } );
```
adekbadek: Is that to be resolved as a part of this PR?

adekbadek (approved):
Works as expected! It's a great change, and the preview generation is much faster now. Pre-approving with one comment – the commented out code should be removed or used.

src/newsletter-editor/editor/index.js
dkoo: Whoops, thanks for catching this! I removed in e0f5a92. It ended up not being needed due to the strategy I ended up implementing (MJML + HTML are regenerated and persisted after save, so on first editor load, the HTML meta should always be up-to-date).

adekbadek:
> I can't think of a way to add unit test coverage to this feature as it stands right now

I don't know if an E2E test would even be possible with phpunit.

PR #1529, opened 2024-06-12
feat: add spinner to subscribe block submit button

chickenn00dle:
### All Submissions:

Closes https://app.asana.com/0/1207385073694408/1206756862605997/f

The subscribe block can take a few seconds to complete and the current loading state is not entirely clear for all publishers. This PR adds a loading spinner to the subscribe form submit button when clicked to make it more clear the form is loading:

https://github.com/Automattic/newspack-newsletters/assets/17905991/c18eda2d-89c5-4355-9c83-b8951dd48e9e

1. Verify reader activation is all set up on your test site
2. Add a newsletter subscription block to any page
3. Go to the relevant page as a reader
4. Input an email address and submit. Confirm a loading spinner appears within the submit button as shown above.

### Other information:

### DIFF
--- a/src/blocks/subscribe/index.php
+++ b/src/blocks/subscribe/index.php
@@ -274,7 +274,9 @@ function render_block( $attrs ) {
 						<input type="hidden" name="double_optin" value="1" />
 					<?php endif; ?>
 
-					<input class="<?php echo \esc_attr( get_block_button_classes( $attrs ) ); ?>"type="submit" value="<?php echo \esc_attr( $attrs['label'] ); ?>" style="<?php echo \esc_attr( get_block_button_styles( $attrs ) ); ?>" />
+					<button class="<?php echo \esc_attr( get_block_button_classes( $attrs ) ); ?>"type="submit" style="<?php echo \esc_attr( get_block_button_styles( $attrs ) ); ?>">
+						<?php echo \esc_html( $attrs['label'] ); ?>
+					</button>
 				</div>
 			</form>
 		<?php endif; ?>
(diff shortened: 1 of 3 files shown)


laurelfulford (changes requested):
Ah, this is a nice addition to the block, @chickenn00dle! 🙌 

The only issue I ran into is when I change the button colour -- the styles "expect" the button to always be white text on red, so when it's changed, the text doesn't disappear, and the spinner doesn't match the color:

image

The Newspack UI styles have an example of a spinner button that's not 100% functional (there's a style applied to a CSS class, but there isn't an example of the CSS class being applied/removed), but I think a couple things there could be used here for the colours issue:

* It has the button text [wrapped in a `span` and the span is given the style `visibility: hidden`](https://github.com/Automattic/newspack-plugin/blob/epic/ras-acc/assets/newspack-ui/scss/elements/forms/_buttons.scss#L200-L202) while loading, so it can't be seen, but it still maintains the size/width of the button.
* It uses [`currentcolor` and `transparency` for the spinner border colours](https://github.com/Automattic/newspack-plugin/blob/epic/ras-acc/assets/newspack-ui/scss/elements/forms/_buttons.scss#L207), so it picks up the text colour for the visible part, and is transparent for the invisible part.

Just let me know if you have any questions at all!

chickenn00dle:
Thank you @laurelfulford!

I updated the changes to use the same newspack-ui spinner strategy in https://github.com/Automattic/newspack-newsletters/pull/1529/commits/90906bb3a233182bb697b6f29054fc2f7f00f12f

I also updated a related PR in the main plugin (for registration form) to use the same strategy if you wanna tackle review for that one as well 🙏 : https://github.com/Automattic/newspack-plugin/pull/3180

laurelfulford (approved):
Thanks @chickenn00dle! That took care of it -- I tried with a few different colour combinations, and didn't see any further issues 🚢 

Edited to add: I'll try to get to the registration one today, too!

PR #1547, opened 2024-06-27
fix: disable autosaves for newsletters

dkoo:
### All Submissions:

After testing #1526 we found that for `newspack_nl_cpt` posts, for currently unknown reasons, autosaving behaves differently compared to other post types.

- In other post types: Autosave requests create a revision which can be reviewed and manually restored to the current edited post state. Autosave revisions will never automatically replace data in the post object.
- In `newspack_nl_cpt` posts: Autosave requests do not create any revision. Instead, they behave exactly like a regular save request, so that any unsaved changes in the post editor are persisted to the post object, just as if you clicked "Save draft".

This PR disables autosaving entirely for `newspack_nl_cpt` posts because this behavior is undesirable and may cause unintentional updates. Since there's no core-sanctioned hook or function for disabling autosaves in a specific post type, this PR does so somewhat redundantly in two different ways:

- By hooking into the `block_editor_settings_all` filter on `newspack_nl_cpt` posts and setting the `autosaveInterval` option to a very high interval (`999999` seconds or ~11 days—this means that a newsletter editor continually open for longer than this might still autosave, but we found that setting this any higher resulted in _far more_ autosave requests rather than fewer).
- By dispatching `lockPostAutosaving()` when the newsletter editor first loads, and whenever `isAutosaveLocked` changes to `false`, so that autosave requests regardless of origin are effectively blocked by the editor.

1. On `alpha`, open a newsletter and make some edits, but do not save.
2. Wait few minutes and refresh the editor, and observe that the unsaved edits were in fact saved.
3. Check out this branch and repeat, and this time confirm that any unsaved edits are lost unless you manually save, send a test, or send the live campaign.

### Other information:

### DIFF
--- a/includes/class-newspack-newsletters-editor.php
+++ b/includes/class-newspack-newsletters-editor.php
@@ -45,6 +45,7 @@ final class Newspack_Newsletters_Editor {
 	 */
 	public function __construct() {
 		add_action( 'init', [ __CLASS__, 'register_meta' ] );
+		add_action( 'block_editor_settings_all', [ __CLASS__, 'disable_autosave' ], 10, 2 );
 		add_action( 'the_post', [ __CLASS__, 'strip_editor_modifications' ] );
 		add_action( 'after_setup_theme', [ __CLASS__, 'newspack_font_sizes' ], 11 );
 		add_action( 'enqueue_block_editor_assets', [ __CLASS__, 'enqueue_block_editor_assets' ] );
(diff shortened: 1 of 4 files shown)


includes/class-newspack-newsletters-editor.php
```diff
@@ -45,6 +45,7 @@ public static function instance() {
 	 */
 	public function __construct() {
 		add_action( 'init', [ __CLASS__, 'register_meta' ] );
+		add_action( 'block_editor_settings_all', [ __CLASS__, 'disable_autosave' ], 10, 2 );
```
claudiulodro: [Nitpick but you'd technically want add_filter](https://developer.wordpress.org/reference/hooks/block_editor_settings_all/)

claudiulodro (approved):
Works well in my testing, but, full disclosure, I did not leave an editor tab open for 11 days to fully verify the autosave timing. :)

I left one small nitpick, doesn't need another review after I think.

includes/class-newspack-newsletters-editor.php
dkoo: Whoops, good catch! Fixed in ba2c652.

I suspect the JS-based `lockPostAutosaving()` is actually what's doing most of the heavy lifting here, but that also seems like a pretty janky way to accomplish this, so I figured some redundancy can't hurt.

PR #844, opened 2022-06-01
feat: active campaign segments support

miguelpeixe:
### All Submissions:

* [ ] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [ ] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [ ] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Implements support for sending newsletters to Active Campaign's segments.

Active Campaign currently supports APIs v1 and v3 and to implement this feature we must use both versions. v1 does not have a method for fetching existing segments whereas v3 does not have a method to create campaigns. To handle that in an organized manner, `api_v1_request()` and `api_v3_request()` methods were implemented.

This ESP has campaigns and messages, a campaign can hold multiple messages and choose which message to send on the scheduling/sending action of the campaign. The current sync action creates a message and a campaign, attaching the message to the campaign. This is problematic because neither API (v1 or v3) has an endpoint for editing a campaign. Once a sync occurs with a specific target (sender and list) it cannot be edited through us.

This PR also changes this behavior to only sync the message while editing, the campaign is only created on sending. This fixes the issue by not having to edit a campaign prior to sending.

Closes #812.

### How to test the changes in this Pull Request:

1. Connect to an existing Active Campaign account that has segments
2. Create a new newsletter and select a list and a segment to send to
3. Send and confirm on the ESP dashboard that it was sent to the selected list and segment

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

PR #1556, opened 2024-07-03
fix: sync newsletter campaigns on post meta update for all ESPs

dkoo:
### All Submissions:

#1526 changed the hook used to sync newsletter posts to the connected ESP to `updated_post_meta`. However, this was erroneously done only for Mailchimp and not the other ESPs.

1. On `release`, switch your ESP to ~a non-Mailchimp ESP~ Constant Contact.
2. Create a new newsletter and observe a message that never resolves in the sidebar: 

cleanshot_2024-07-03_at_10 51 57_2x

3. Also observe that in the connected Constant Contact account, a synced campaign is never created.
4. Switch your ESP to Active Campaign.
5. Create a new newsletter and edit the content, then save. Observe that in the synced campaign in Active Campaign, the newsletter content is not updated after the save.
6. Check out this branch, switch back to CC, create another new newsletter, and confirm that the campaign syncs correctly for new newsletters. (You can also refresh and save the original newsletter created in step 1 and confirm that it gets synced upon save).
7. Switch to AC, then edit and save the newsletter you created in step 5. Confirm that the synced campaign in AC is updated after saving.

### Other information:

### DIFF
--- a/includes/class-newspack-newsletters-editor.php
+++ b/includes/class-newspack-newsletters-editor.php
@@ -318,6 +318,14 @@ final class Newspack_Newsletters_Editor {
 			],
 			'supported_social_icon_services' => Newspack_Newsletters_Renderer::get_supported_social_icons_services(),
 		];
+
+		// Fetch async error messages to display on editor load.
+		$transient_name = $provider->get_transient_name( get_the_ID() );
+		$error_message  = get_transient( $transient_name );
+		if ( $error_message ) {
+			$email_editor_data['error_message'] = $error_message;
+		}
+
 		if ( self::is_editing_email() ) {
 			wp_register_style(
 				'newspack-newsletters',
--- a/includes/class-newspack-newsletters-renderer.php
+++ b/includes/class-newspack-newsletters-renderer.php
@@ -110,6 +110,10 @@ final class Newspack_Newsletters_Renderer {
 	 * @return string A value for the alt attribute.
 	 */
 	private static function get_image_alt( $attachment_id ) {
+		if ( ! $attachment_id ) {
+			return '';
+		}
+
 		$alt        = get_post_meta( $attachment_id, '_wp_attachment_image_alt', true );
 		$attachment = get_post( $attachment_id );
 
@@ -595,7 +599,7 @@ final class Newspack_Newsletters_Renderer {
 				$dom->loadHTML( mb_convert_encoding( $inner_html, 'HTML-ENTITIES', get_bloginfo( 'charset' ) ), LIBXML_HTML_NOIMPLIED | LIBXML_HTML_NODEFDTD );
 				$img        = $dom->getElementsByTagName( 'img' )->item( 0 );
 				$img_src    = $img->getAttribute( 'src' );
-				$img_alt    = self::get_image_alt( $attrs['id'] );
+				$img_alt    = self::get_image_alt( $attrs['id'] ?? '' );
 				$figcaption = $dom->getElementsByTagName( 'figcaption' )->item( 0 );
 
 				$img_attrs = array(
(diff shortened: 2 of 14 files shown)


laurelfulford:
I tried testing this but I run into issues that I'm not sure are "real" or just me -- adding below as a comment instead of as a review since I'm not confident in them. 

**Constant Contact:**
* The "Retrieving data..." message in the sidebar is gone. 
* When I create a new newsletter, it syncs the initial draft to Constant Contact, but subsequent draft saves don't appear to update it. I'm not sure if this is an issue or not, except...
* When I try to send to a small email list I created, I get an error:

image

Unfortunately it doesn't look like we can send test emails with our CC account, so I can't confirm whether this actually is updating right. 

**MailChimp:**
* Does not show any "Retrieving data..." message.
* When I create a new newsletter, it synced to MC, and subsequent draft saves update it.
* Sending an email to a small group works fine.

**ActiveCampaign:**
* Does not show any "Retrieving data..." message. I noticed I briefly got the spinner in the same area, but it quickly resolved. That issue doesn't appear to happen for AC on release either, though.
* When I create a new newsletter it doesn't seem to sync to AC at all -- it doesn't show up in the Campaigns list, but I'm not sure this is expected.
* Sending an email to a small group works fine. It included my latest changes after making edits and saving the draft a few times, so I suspect the previous bullet point doesn't matter. 

If someone else tests this and doesn't run into the same CC issues, I'm going to chalk that up to a bad day on my part, but the site I tested on can be shared if that would help!

dkoo:
@laurelfulford thanks for testing! I did confirm that our Constant Contact `sync` method was missing the step to rename the synced campaign based on the "Campaign Name" meta value, so I added that in 024e6ed. 

I also was able to confirm that the campaign subject/preview text/body/send to options are updated correctly upon save, and that when sending a test email or a live campaign, the changes I make persist to the sent email.

> When I create a new newsletter it doesn't seem to sync to AC at all -- it doesn't show up in the Campaigns list, but I'm not sure this is expected.

I think this is expected. ActiveCampaign doesn't support updating a campaign via their API once it's been created. So we don't create a campaign in AC until we have to send it (either as a test, or as a live campaign). More details here: #844

laurelfulford:
Thanks for double-checking that, @dkoo! 

I'm still having issues with Constant Contact sending, and updating the email drafts past the newsletter creation 😕 I'm going to try recreating the connection and see if I have more luck!

Edited to add: it's still not working for me. I'm not sure what I'm doing wrong, but if you're able to send emails through Constant Contact, it's most likely something I'm doing wrong.

dkoo:
@laurelfulford  Turns out the `sync` request from your test site was failing with an error message because of a setup issue with the connected CC account, but due to the fact that #1526 triggers the sync as a side effect of a save (and not during the save itself) we're not seeing the error message in the editor. Once I corrected the issue in the CC account, the syncs  started working again.

I think this approach will need to be refined in a future PR, but 224d40d adds some basic error handling to capture error messages from the `sync` requests in a transient which is then loaded and displayed when the editor next loads. This still means the error message won't be displayed until the editor is next refreshed or reloaded, but at least we'll show it somewhere. The transient is cleared the next time a `sync` request is attempted.

To test this, in your test CC account visit Account Settings and edit your business details. Set the ZIP code to something invalid like `11-11` and save, then attempt to save a newsletter from the WP newsletter editor. After 224d40d, when you next refresh the editor you should see an error message like `Constant Contact: Physical Address Postal Code is not in the valid format, please include a valid address postal code (zip code) with this request`

leogermani (approved):
All works fine. The Campaign only gets created in AC after I send a test email. I believe that's the expected behavior for AC

PR #1562, opened 2024-07-09
feat: refactor add_contact and put async sign up behind a flag

leogermani:
### All Submissions:

Taking baby steps to centralize our data flows by deprecating some methods and moving them to a new class.

At this point there's no relevant changes made, it's mostly moving code around. In a next PR we are going to start changing things, and also making changes to the places where this deprecated methods are called.

The only change we make is to put the Async strategy behind a constant.

1. Test signing up for a newsletter using the Newsletters Subscription block
2. Confirm it works as expected (check the ESP dashboard to make sure the contact was added to the correct lists)
3. Confirm you see `PHP Deprecated:  Function Newspack_Newsletters_Subscription::add_contact is <strong>deprecated</strong> since version 2.21! Use Newspack_Newsletters_Contacts::upsert instead. in /var/www/html/wp-includes/functions.php on line 6085` on your logs
4. Go to my account, verify your email, and test updating your selection of newsletters lists. Confirm it works
5. Confirm you see `PHP Deprecated:  Function Newspack_Newsletters_Subscription::add_contact is <strong>deprecated</strong> since version 2.21! Use Newspack_Newsletters_Contacts::upsert instead. in /var/www/html/wp-includes/functions.php on line 6085`
6. As an admin, delete the user
7. Confirm the user was deleted from the ESP
8. Add `define( 'NEWSPACK_NEWSLETTERS_ASYNC_SUBSCRIPTION_ENABLED', true );` to `wp-config.php` and repeat the tests, making sure the async strategy is still working

### Other information:

---
  - https://app.asana.com/0/0/1207717004877598

### DIFF
--- a/includes/class-newspack-newsletters-contacts.php
+++ b/includes/class-newspack-newsletters-contacts.php
@@ -1,0 +1,257 @@
+<?php
+/**
+ * Newspack Newsletters Contacts class.
+ *
+ * This class holds the methods for managing contacts. It's meant to be used by this plugin and by external integrations.
+ *
+ * @package Newspack
+ */
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Class Newspack_Newsletters_Contacts
+ */
+class Newspack_Newsletters_Contacts {
+
+	/**
+	 * Upserts a contact to lists.
+	 *
+	 * A contact can be added asynchronously, which means the request will return
+	 * immediately and the contact will be added in the background. In this case
+	 * the response will be `true` and the caller must handle it optimistically.
+	 * NEWSPACK_NEWSLETTERS_ASYNC_SUBSCRIPTION_ENABLED must be defined as true for this feature to be available.
+	 *
+	 * @param array          $contact {
+	 *          Contact information.
+	 *
+	 *    @type string   $email    Contact email address.
+	 *    @type string   $name     Contact name. Optional.
+	 *    @type string[] $metadata Contact additional metadata. Optional.
+	 * }
+	 * @param string[]|false $lists   Array of list IDs to subscribe the contact to. If empty or false, contact will be created but not subscribed to any lists.
+	 * @param bool           $async   Whether to add the contact asynchronously. Default is false.
+	 *
+	 * @return array|WP_Error|true Contact data if it was added, or error otherwise. True if async.
+	 */
+	public static function upsert( $contact, $lists = false, $async = false ) {
+		if ( ! is_array( $lists ) && false !== $lists ) {
+			$lists = [ $lists ];
+		}
+
+		/**
+		 * Trigger an action before contact adding.
+		 *
+		 * @param string[]|false $lists    Array of list IDs the contact will be subscribed to, or false.
+		 * @param array          $contact  {
+		 *          Contact information.
+		 *
+		 *    @type string   $email    Contact email address.
+		 *    @type string   $name     Contact name. Optional.
+		 *    @type string[] $metadata Contact additional metadata. Optional.
+		 * }
+		 */
+		do_action( 'newspack_newsletters_pre_add_contact', $lists, $contact );
+
+		$provider = Newspack_Newsletters::get_service_provider();
+		if ( empty( $provider ) ) {
+			return new WP_Error( 'newspack_newsletters_invalid_provider', __( 'Provider is not set.' ) );
+		}
+
+		if ( defined( 'NEWSPACK_NEWSLETTERS_ASYNC_SUBSCRIPTION_ENABLED' ) && NEWSPACK_NEWSLETTERS_ASYNC_SUBSCRIPTION_ENABLED && true === $async ) {
+			Newspack_Newsletters_Subscription::add_subscription_intent( $contact, $lists );
+			return true;
+		}
+
+		if ( false !== $lists ) {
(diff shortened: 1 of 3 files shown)


includes/class-newspack-newsletters-subscription.php
```diff
@@ -698,22 +547,8 @@ private static function add_contact_to_provider( $contact, $lists = [], $is_upda
...
-		$provider = Newspack_Newsletters::get_service_provider();
-		if ( empty( $provider ) ) {
-			return new WP_Error( 'newspack_newsletters_invalid_provider', __( 'Provider is not set.' ) );
-		}
-		if ( ! method_exists( $provider, 'delete_contact' ) ) {
-			return new WP_Error( 'newspack_newsletters_invalid_provider_method', __( 'Provider does not support deleting user subscriptions.' ) );
-		}
-		return $provider->delete_contact( $user->user_email );
+		_deprecated_function( __METHOD__, '2.21', 'Newspack_Newsletters_Contacts::delete' );
+		return Newspack_Newsletters_Contacts::delete( $contact, $lists, $async );
```
dkoo: Seems like all three of these variables are undefined, so I'm not seeing the deletion happen when I delete from the admin.

dkoo (changes requested):
Mostly working, except for the `delete` method.

dkoo:
@leogermani is removing the async behavior of signup requests via the Subscription Form block also within scope here? Seems like that should be synchronous too unless the constant is present.

EDIT: Looks like this can be a one-line change by changing [this](https://github.com/Automattic/newspack-newsletters/blob/fix/error-handling-my-account/src/blocks/subscribe/index.php#L508) `true` to check the constant instead.

includes/class-newspack-newsletters-subscription.php
leogermani: weird. pretty sure I tested this

fixed in https://github.com/Automattic/newspack-newsletters/pull/1562/commits/fe05adb6f0135cec837db9de2c7d74f2814a2156

also fixed tests in https://github.com/Automattic/newspack-newsletters/pull/1562/commits/0c56849199a977bff8b0e03ea16c064504ec076c

leogermani:
> @leogermani is removing the async behavior of signup requests via the Subscription Form block also within scope here? Seems like that should be synchronous too unless the constant is present.
> 
> EDIT: Looks like this can be a one-line change by changing [this](https://github.com/Automattic/newspack-newsletters/blob/fix/error-handling-my-account/src/blocks/subscribe/index.php#L508) `true` to check the constant instead.

Not sure I get what you mean. I'm checking the constant [here](https://github.com/Automattic/newspack-newsletters/pull/1562/files#diff-2c51024c343e46470774b97265cded37f57e841b1db7a53a0f5a6f3e93053224R61). If the constant is not set to true, that `true` parameter will be ignored. Am I missing something?

dkoo:
@leogermani you're right—I just wasn't seeing the error on the front-end due to [this issue](https://github.com/Automattic/newspack-newsletters/pull/1567). If I merge that fix into this PR, I see the error response being returned in the block front-end as I would expect without the async flag.

includes/class-newspack-newsletters-contacts.php
```diff
@@ -0,0 +1,257 @@
...
+	 * @param int $user_id User ID.
+	 *
+	 * @return bool|WP_Error Whether the contact was deleted or error.
+	 */
+	public static function delete( $user_id ) {
+		$user = get_user_by( 'id', $user_id );
+		if ( ! $user ) {
+			return new WP_Error( 'newspack_newsletters_invalid_user', __( 'Invalid user.' ) );
+		}
+		/** Only delete if email ownership is verified. */
```
dkoo, line 131: Non-blocking question: Do we want this to be the case? If deleting a user account via admin, then unverified reader accounts will be deleted in the site but not in the ESP, creating a data discrepancy. Non-blocking because this is currently true in `trunk` as well.

dkoo (approved):
Working as expected now, but I have a non-blocking question below.

leogermani:
@dkoo I have merged #1567 and had to redo the changes in the new Contacts class. See https://github.com/Automattic/newspack-newsletters/pull/1562/commits/294776ee90ee083d9c9227cf75e82d4a368f50ef

Note that I have removed the [error deletion](https://github.com/Automattic/newspack-newsletters/pull/1567/files#diff-dce6057acc32bc5e49a68179cfc17e3e12a57c347768c34779d25668bd8f16beR680) altogether because it's not necessary. These errors have to do with the Signup Intents (async requests) and are already cleared [here](https://github.com/Automattic/newspack-newsletters/blob/294776ee90ee083d9c9227cf75e82d4a368f50ef/includes/class-newspack-newsletters-subscription.php#L489)

EDIT - I didn't commit all the files in the merge. Had to do it in a new commit -> https://github.com/Automattic/newspack-newsletters/pull/1562/commits/c623e561396fe29706c53a8f3952adee82644ff1

includes/class-newspack-newsletters-contacts.php
leogermani: Yes, agree with you, it looks odd.

But we are not changing behavior in the PR. If we want to change that I'd say let's do it another time

dkoo (approved):
Still working after merging #1567!

PR #1566, opened 2024-07-10
feat(constant-contact): enable subscription management in my account

dkoo:
### All Submissions:

Adds missing methods to the Constant Contact service provider class to enable list/tag subscription self-management via the My Account > Newsletters dashboard. This should allow Constant Contact to be used with Premium Newsletter features.

1. Check out this branch.
2. Connect Constant Contact as your ESP and authorize as needed.
3. In **Newspack > Engagement > Newsletters**, enable at least two lists as  subscription lists and also create and publish at least one local subscription list (which will create a tag in the ESP).
4. As a reader, sign up for the list and the tag via any means.
5. Visit My Account, verify as needed, and confirm you see the "Newsletters" menu item in the sidebar.
6. Visit Newsletters and confirm that you see all enabled subscription lists and that the ones you signed up for as a reader are checked.
7. Unsubscribe from all and confirm that they remain unchecked after the page refreshes. Also confirm that the contact in the ESP has been removed from all tags and lists they were subscribed to.
8. Resubscribe to all and confirm that they remain checked after the page refreshes. Also confirm that the contact in the ESP has been readded to all tags and lists you selected in My Account.
9. Resubscribe only to the local (tag) list and confirm it works
10. Unsubscribe from that and subscribe only to one normal list
11. Test all possible combinations you can

### Other information:

### DIFF
--- a/includes/service-providers/constant_contact/class-newspack-newsletters-constant-contact.php
+++ b/includes/service-providers/constant_contact/class-newspack-newsletters-constant-contact.php
@@ -932,52 +932,21 @@ final class Newspack_Newsletters_Constant_Contact extends \Newspack_Newsletters_
 			}
 		}
 
-		// Existing contact lists/tags.
-		$contact_lists    = $contact_data['list_memberships'] ?? [];
-		$contact_tags     = $contact_data['taggings'] ?? [];
-		$new_contact_data = [
-			'list_ids' => $contact_lists,
-			'taggings' => $contact_tags,
-		];
-
 		// Remove lists or tags from contact.
-		foreach ( $lists_to_remove as $list_id ) {
-			$is_tag = ! is_wp_error( $this->get_tag_by_id( $list_id ) );
-			if ( $is_tag ) {
-				$new_contact_data['taggings'] = array_values(
-					array_filter(
-						$new_contact_data['taggings'],
-						function ( $tag_id ) use ( $list_id ) {
-							return $tag_id !== $list_id;
-						}
-					)
-				);
-			} else {
-				$result = $cc->remove_contacts_from_lists( [ $contact_data['contact_id'] ], [ $list_id ] );
-				if ( is_wp_error( $result ) ) {
-					return $result;
-				}
+		if ( $lists_to_remove ) {
+			$result = $cc->remove_contacts_from_lists( [ $contact_data['contact_id'] ], $lists_to_remove );
+			if ( is_wp_error( $result ) ) {
+				return $result;
 			}
 		}
 
 		// Add lists or tags to contact.
-		foreach ( $lists_to_add as $list_id ) {
-			$is_tag    = ! is_wp_error( $this->get_tag_by_id( $list_id ) );
-			$item_type = $is_tag ? 'taggings' : 'list_ids';
-
-			$new_contact_data[ $item_type ] = array_values(
-				array_unique(
-					array_merge(
-						$new_contact_data[ $item_type ],
-						[ $list_id ]
-					)
-				)
-			);
-		}
-
-		$result = $cc->upsert_contact( $email, $new_contact_data );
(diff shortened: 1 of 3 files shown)


leogermani:
@dkoo I've pushed a commit to the PR

There's no need to handle tags there. This is done by the abstract Provider class in `update_contact_lists_handling_local` which is what's actually called.

Also simplified the code a little and fixed it, beacuse it was always resubscribing me to the list if I tried to leave only the Local list enabled.

Now we could do a round of tests with Premium Newsletters just to confirm it works, but it should

dkoo (commented):
Changes look good to me and work as expected! Just a couple of questions below:

includes/service-providers/constant_contact/class-newspack-newsletters-constant-contact.php
```diff
@@ -887,6 +887,71 @@ public function get_contact_data( $email, $return_details = false ) {
...
+			// Call Newspack_Newsletters_Subscription's method (not the provider's directly),
+			// so the appropriate hooks are called.
+			$contact_data = Newspack_Newsletters_Subscription::add_contact( [ 'email' => $email ] );
+			if ( is_wp_error( $contact_data ) ) {
+				return $contact_data;
+			}
+		}
+
+		// Remove lists or tags from contact.
+		if ( $lists_to_remove ) {
```
dkoo, line 948: Does this need to be `if ( ! empty( $lists_to_remove ) ) {` instead? I believe an empty array will always evaluate to `true`, so this might result in extra requests that don't end up doing anything, right?

includes/service-providers/constant_contact/class-newspack-newsletters-constant-contact.php
```diff
@@ -887,6 +887,71 @@ public function get_contact_data( $email, $return_details = false ) {
...
+			$result = $cc->remove_contacts_from_lists( [ $contact_data['contact_id'] ], $lists_to_remove );
+			if ( is_wp_error( $result ) ) {
+				return $result;
+			}
+		}
+
+		// Add lists or tags to contact.
+		if ( $lists_to_add ) {
```
dkoo, line 956: Same question: should it be `if ( ! empty( $lists_to_add ) ) {`?

includes/service-providers/constant_contact/class-newspack-newsletters-constant-contact.php
leogermani: Empty arrays evaluate to `false`, unless this changed in some php versions.

But I'm not against a more explicit check. Explicit > Implicit :)

dkoo: Ah, you're right! I must have been thinking of JS. I'm okay with it either way

dkoo:
@adekbadek can we ask you for another set of eyes here, since @leogermani and I both worked on the code?

adekbadek (changes requested):
When subscribing, the contact is added with the correct lists/tags, but a PHP Fatal is thrown (in the async handling, so it's not surfaced in the browser):

```
PHP Fatal error:  Uncaught TypeError: get_object_vars(): Argument #1 ($object) must be of type object, bool given in […]newspack-newsletters/includes/service-providers/constant_contact/class-newspack-newsletters-constant-contact.php:867
```

While this does not prevent me from completing the testing steps, this case should be handled (and cause known).

leogermani:
I cannot reproduce the Fatal described by @adekbadek . What is the exact action that triggers it?

dkoo:
> I cannot reproduce the Fatal described by @adekbadek 

Same, so far. Looking at the stack trace, though, I see where the result of `upsert_contact` could be a bool value. Does 05b1333 fix the fatal for you, @adekbadek?

adekbadek (approved):
Yep, that fixes it.

PR #1577, opened 2024-07-19
feat: make "Send To" UI autocomplete instead of dropdowns

dkoo:
### All Submissions:

Changes the "Send To" sidebar UI in the Newsletter editor from a series of dropdowns to an autocomplete input field for Mailchimp, ActiveCampaign, and Constant Contact. This PR doesn't change the way the audience/list/tag/segment/etc. data is fetched and stored in the editor for different ESPs, it only standardizes the UIs for now. The ultimate goal to simplify the data handling across ESPs will happen in future PRs.

Doesn't touch the editor UI for Campaign Monitor as we're planning to deprecate support for this ESP in the near future. Since this implements new editor-facing UI, I'll tag @Automattic/newspack-design for feedback (but note that it only uses existing core WP components).

Also note that this PR breaks error handling for the "Update Sender" UI, but since we're planning to update this UI as well we'll fix it in a future PR.

#### Before

<table>
<tr>
<th>Mailchimp</th>
<th>ActiveCampaign</th>
<th>Constant Contact</th>
</tr>
<tr>
<td>
before-mc
</td>
<td>
before-ac
</td>
<td>
before-cc
</td>
</tr>
</table>

#### After

<table>
<tr>
<th>Mailchimp</th>
<th>ActiveCampaign</th>
<th>Constant Contact</th>
</tr>
<tr>
<td>
after-mc
</td>
<td>
after-ac
</td>
<td>
after-cc
</td>
</tr>
</table>

We'll want to test editing and sending newsletters for each ESP.

1. Connect your site to Mailchimp via the Newsletters settings page or Engagement > Newsletters dashboard.
2. Create a new newsletter.
3. Confirm that the new UI shows an autocomplete field for Audience.
4. Start typing the name of an audience in your connected Mailchimp account (typing `audience` should also work) and confirm that the autocomplete shows suggestions based on the available audiences.
5. Choose an audience and confirm that the autocomplete input changes to show the selected audience, and that a new "Group, Segment, or Tag" autocomplete field appears underneath it. Also confirm in the connected Mailchimp account that the synced campaign is updated whenever you select an audience in the editor. Note that Mailchimp doesn't support unsetting an audience for a campaign via API once it's been set, so you can't reset the audience to empty—you can only edit and choose a different audience.

after-mc

6. Start typing a group, segment, or tag name (typing `group`, `segment`, or `tag` should also work) and confirm that the input shows appropriate suggestions from your Mailchimp account.
7. Choose one and confirm that the UI updates and that the synced campaign also updates to match your selection. Edit and repeat with each sub-audience type to confirm that all work as expected.
8. Click "Reset" and confirm that the autocomplete input appears again, and that the synced campaign is reset to "All subscribers in audience"
9. Send a newsletter to an entire audience, then another to each sub-audience type, and confirm that it gets sent correctly in each case
10. Change your site's connected ESP to ActiveCampaign and repeat all testing steps, then again with Constant Contact. Note the following subtle differences between ESPs:
  - ActiveCampaign won't sync new campaigns in the connected ESP account due to limitations in their API around updating existing campaigns. Instead, list selections are stored as post meta and only synced when sending the live campaign.
  - ActiveCampaign segments have a truly strange UI to create/edit. See [this doc](https://help.activecampaign.com/hc/en-us/articles/221483407-How-to-create-segments-in-ActiveCampaign#h_01HHFAA1Q15DMK9V53GX2YJNCQ) for details on how.
  - Constant Contact's API does support sending to more than one list at a time, a feature we support in the current UI. However, in order to standardize the UI across ESPs, this PR deprecates this ability and enforces selecting only one list per newsletter.
  - Constant Contact also supports sending to either multiple lists OR a single segment, but not both at the same time, so its UI shows only a single autocomplete that covers both data types.

### Other information:

---
  - https://app.asana.com/0/0/1207717004877578

### DIFF
--- a/src/newsletter-editor/sidebar/send-to.js
+++ b/src/newsletter-editor/sidebar/send-to.js
@@ -4,6 +4,7 @@
 import { __ } from '@wordpress/i18n';
 import { FormTokenField, Button, ButtonGroup, Notice } from '@wordpress/components';
 import { useState } from '@wordpress/element';
+import { Icon, external } from '@wordpress/icons';
 
 const SendTo = ( {
 	availableLists,
@@ -31,7 +32,6 @@ const SendTo = ( {
 				<div>
 					<p>
 						<strong>{ getLabel( selectedList ) }</strong>{ ' ' }
-						{ getLink ? getLink( selectedList ) : null }
 					</p>
 				</div>
 			) : (
@@ -67,6 +67,7 @@ const SendTo = ( {
 							<Button
 								disabled={ isUpdating }
 								onClick={ () => setIsEditing( true ) }
+								size="small"
 								variant="secondary"
 							>
 								{ __( 'Edit', 'newspack-newsletters' ) }
@@ -88,15 +89,29 @@ const SendTo = ( {
 												setIsEditing( false );
 											} );
 									} }
+									size="small"
 									variant="secondary"
 								>
 									{ __( 'Reset', 'newspack-newsletters' ) }
 								</Button>
 							) }
+							{ getLink && (
+								<Button
+									disabled={ isUpdating }
+									href={ getLink( selectedList ) }
+									size="small"
+									target="_blank"
+									variant="secondary"
+									rel="noopener noreferrer"
+								>
+									{ __( 'Manage', 'newspack-newsletters' ) }
+									<Icon icon={ external } size={ 14 } />
+								</Button>
+							) }
 						</>
 					) }
 					{ isEditing && (
-						<Button onClick={ () => setIsEditing( false ) } variant="secondary">
+						<Button onClick={ () => setIsEditing( false ) } variant="secondary" size="small">
 							{ __( 'Cancel', 'newspack-newsletters' ) }
 						</Button>
 					) }
--- a/src/newsletter-editor/sidebar/send-to.js
+++ b/src/newsletter-editor/sidebar/send-to.js
@@ -1,21 +1,12 @@
 /**
  * WordPress dependencies
  */
-import { __ } from '@wordpress/i18n';
+import { __, sprintf } from '@wordpress/i18n';
 import { FormTokenField, Button, ButtonGroup, Notice } from '@wordpress/components';
 import { useState } from '@wordpress/element';
 import { Icon, external } from '@wordpress/icons';
 
-const SendTo = ( {
-	availableLists,
-	formLabel,
-	getLabel,
-	getLink = null,
-	onChange,
-	placeholder,
-	reset,
-	selectedList,
-} ) => {
+const SendTo = ( { availableItems, formLabel, onChange, placeholder, reset, selectedItem } ) => {
 	const [ isEditing, setIsEditing ] = useState( false );
 	const [ isUpdating, setIsUpdating ] = useState( false );
 	const [ error, setError ] = useState( false );
@@ -28,12 +19,26 @@ const SendTo = ( {
 				</Notice>
 			) }
 
-			{ selectedList && ! isEditing ? (
-				<div>
-					<p>
-						<strong>{ getLabel( selectedList ) }</strong>{ ' ' }
+			{ selectedItem && ! isEditing ? (
+				<>
+					<p className="newspack-newsletters__send-to-details">
+						{ selectedItem.name }
+						<span>
+							{ sprintf(
+								// Translators: %1$s is the item type, %2$s is details about the item.
+								__( '%1$s %2$s', 'newspack-newsletters' ),
+								isUpdating
+									? sprintf(
+											// Translators: The item type.
+											__( 'Resetting %s…', 'newspack-newsletters' ),
+											selectedItem.typeLabel.toLowerCase()
+									  )
+									: selectedItem.typeLabel,
+								! isUpdating && selectedItem.details ? ' • ' + selectedItem.details : ''
+							).trim() }
+						</span>
 					</p>
-				</div>
+				</>
 			) : (
 				<>
 					<FormTokenField
@@ -52,7 +57,7 @@ const SendTo = ( {
 									setIsEditing( false );
 								} );
 						} }
-						suggestions={ availableLists.map( suggestion => suggestion.label ) }
+						suggestions={ availableItems.map( item => item.label ) }
 						placeholder={ isUpdating ? __( 'Updating campaign…', 'newspack' ) : placeholder }
 						value={ [] }
 						__experimentalExpandOnFocus={ true }
@@ -60,7 +65,7 @@ const SendTo = ( {
 					/>
 				</>
 			) }
-			{ selectedList && (
+			{ selectedItem && (
 				<ButtonGroup>
 					{ ! isEditing && (
 						<>
@@ -95,10 +100,10 @@ const SendTo = ( {
 									{ __( 'Reset', 'newspack-newsletters' ) }
 								</Button>
 							) }
-							{ getLink && (
+							{ selectedItem.editLink && (
 								<Button
 									disabled={ isUpdating }
-									href={ getLink( selectedList ) }
+									href={ selectedItem.editLink }
 									size="small"
 									target="_blank"
 									variant="secondary"
--- a/src/newsletter-editor/sidebar/send-to.js
+++ b/src/newsletter-editor/sidebar/send-to.js
@@ -39,6 +39,7 @@ const SendTo = ( {
 					<FormTokenField
 						disabled={ isUpdating }
 						label={ formLabel || __( 'Select a list', 'newspack' ) }
+						maxSuggestions={ 10 }
 						onChange={ items => {
 							setError( false );
 							setIsUpdating( true );
@@ -54,6 +55,7 @@ const SendTo = ( {
 						suggestions={ availableLists.map( suggestion => suggestion.label ) }
 						placeholder={ isUpdating ? __( 'Updating campaign…', 'newspack' ) : placeholder }
 						value={ [] }
+						__experimentalExpandOnFocus={ true }
 						__experimentalShowHowTo={ false }
 					/>
 				</>
--- a/src/newsletter-editor/sidebar/send-to.js
+++ b/src/newsletter-editor/sidebar/send-to.js
@@ -24,30 +24,21 @@ const SendTo = ( { availableItems, formLabel, onChange, placeholder, reset, sele
 					<p className="newspack-newsletters__send-to-details">
 						{ selectedItem.name }
 						<span>
-							{ sprintf(
-								// Translators: %1$s is the item type, %2$s is details about the item.
-								__( '%1$s %2$s', 'newspack-newsletters' ),
-								isUpdating
-									? sprintf(
-											// Translators: The item type.
-											__( 'Resetting %s…', 'newspack-newsletters' ),
-											selectedItem.typeLabel.toLowerCase()
-									  )
-									: selectedItem.typeLabel,
-								! isUpdating && selectedItem.hasOwnProperty( 'count' )
-									? ' • ' +
-											sprintf(
-												// Translators: %d is the number of contacts in the list.
-												_n(
-													'%d contact',
-													'%d contacts',
-													selectedItem.count,
-													'newspack-newsletters'
-												),
-												selectedItem.count.toLocaleString()
-											)
-									: ''
-							).trim() }
+							{ isUpdating
+								? sprintf(
+										// Translators: Shown while resetting the selected send-to item. %s is the item type.
+										__( 'Resetting %s…', 'newspack-newsletters' ),
+										selectedItem.typeLabel.toLowerCase()
+								  )
+								: selectedItem.typeLabel }
+							{ ! isUpdating && selectedItem.hasOwnProperty( 'count' )
+								? ' • ' +
+								  sprintf(
+										// Translators: If available, show a contact count alongside the selected item's type. %d is the number of contacts in the item.
+										_n( '%d contact', '%d contacts', selectedItem.count, 'newspack-newsletters' ),
+										selectedItem.count.toLocaleString()
+								  )
+								: '' }
 						</span>
 					</p>
 				</>
--- a/src/newsletter-editor/sidebar/send-to.js
+++ b/src/newsletter-editor/sidebar/send-to.js
@@ -1,0 +1,105 @@
+/**
+ * WordPress dependencies
+ */
+import { __ } from '@wordpress/i18n';
+import { FormTokenField, Button, ButtonGroup, Notice } from '@wordpress/components';
+import { useState } from '@wordpress/element';
+
+const SendTo = ( {
+	availableLists,
+	formLabel,
+	getLabel,
+	getLink = null,
+	onChange,
+	placeholder,
+	reset,
+	selectedList,
+} ) => {
+	const [ isEditing, setIsEditing ] = useState( false );
+	const [ isUpdating, setIsUpdating ] = useState( false );
+	const [ error, setError ] = useState( false );
+
+	return (
+		<>
+			{ error && ! isEditing && (
+				<Notice status="error" isDismissible={ false }>
+					{ error }
+				</Notice>
+			) }
+
+			{ selectedList && ! isEditing ? (
+				<div>
+					<p>
+						<strong>{ getLabel( selectedList ) }</strong>{ ' ' }
+						{ getLink ? getLink( selectedList ) : null }
+					</p>
+				</div>
+			) : (
+				<>
+					<FormTokenField
+						disabled={ isUpdating }
+						label={ formLabel || __( 'Select a list', 'newspack' ) }
+						onChange={ items => {
+							setError( false );
+							setIsUpdating( true );
+							onChange( items )
+								.catch( e => {
+									setError( e.message || __( 'Error updating campaign.', 'newspack-newsletters' ) );
+								} )
+								.finally( () => {
+									setIsUpdating( false );
+									setIsEditing( false );
+								} );
+						} }
+						suggestions={ availableLists.map( suggestion => suggestion.label ) }
+						placeholder={ isUpdating ? __( 'Updating campaign…', 'newspack' ) : placeholder }
+						value={ [] }
+						__experimentalShowHowTo={ false }
+					/>
+				</>
+			) }
+			{ selectedList && (
+				<ButtonGroup>
+					{ ! isEditing && (
+						<>
+							<Button
+								disabled={ isUpdating }
+								onClick={ () => setIsEditing( true ) }
+								variant="secondary"
+							>
+								{ __( 'Edit', 'newspack-newsletters' ) }
+							</Button>
+							<Button
+								disabled={ isUpdating }
+								onClick={ () => {
+									setError( false );
+									setIsUpdating( true );
+									reset()
+										.catch( e => {
+											setError(
+												e.message || __( 'Error updating campaign.', 'newspack-newsletters' )
+											);
+										} )
+										.finally( () => {
+											setIsUpdating( false );
+											setIsEditing( false );
+										} );
+								} }
+								variant="secondary"
+							>
+								{ __( 'Reset', 'newspack-newsletters' ) }
+							</Button>
+						</>
+					) }
+					{ isEditing && (
+						<Button onClick={ () => setIsEditing( false ) } variant="secondary">
+							{ __( 'Cancel', 'newspack-newsletters' ) }
+						</Button>
+					) }
+				</ButtonGroup>
+			) }
+		</>
+	);
+};
+
+export default SendTo;
--- a/src/newsletter-editor/sidebar/send-to.js
+++ b/src/newsletter-editor/sidebar/send-to.js
@@ -34,7 +34,7 @@ const SendTo = ( { availableItems, formLabel, onChange, placeholder, reset, sele
 											selectedItem.typeLabel.toLowerCase()
 									  )
 									: selectedItem.typeLabel,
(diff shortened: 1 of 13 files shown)


leogermani:
Nice work @dkoo , a few comments/questions

On Constant Contact, a few times the UI froze with "Retrieving Constant Contact Data...". A refresh would solve the problem, not sure what's going on.

I understand this PR is just updating the UI, preparing the field for the refactor on how we fetch data from the ESPs. But thinking about that, the proposed UI has one single input field to auto complete different kind of data (lists, tags, groups, segments).

Considering that each of these different entities has its own endpoint if the ESP APIs, I was wondering if the UI should not have a SELECT box for the user to select what they will be looking for, so we know which endpoint we will reach. How are you thinking about that?

Also, as a minor suggestion, in Mailchimp I think it would be ok to leave the Audience field as a select box, and do the autocomplete only for the segment/tags/groups.

dkoo:
> Considering that each of these different entities has its own endpoint if the ESP APIs, I was wondering if the UI should not have a SELECT box for the user to select what they will be looking for, so we know which endpoint we will reach. How are you thinking about that?

@leogermani I was thinking we should standardize the data that makes it into the editor for all ESPs. You're right that all of the ESPs have their own data structures and APIs to interact with those, but in my mind they all more or less break down into the following:

- Lists: Also called audiences in MC. The the top-level collection of contacts (and I'm starting to use this word more instead of subscribers as I think it's clearer) in any ESP.
- Sublists: These are all of the different ways the ESPs break lists into smaller groups. They can be segments, tags, interests, etc. These are more varied in their data structure so we'll need ways to disambiguate between different types for each ESP—maybe with a `sublist_type` key and a schema to handle each so we know what data to expect via the ESPs' APIs.

At some point we'll still have to deal with the different APIs and data structures for each ESP, but I'm thinking we should be able to abstract those things deeper down in the PHP classes so that the API endpoints to fetch lists and sublists (and the data they return to the editor) can be the same no matter what ESP is active. It'll be in the PHP methods that handle the data where we decide how to fetch that data from the ESPs and reformat it for the front-end. That way, we can hopefully refactor the editor JS so that all ESPs can share a single set of sidebar components.

There are still some nuances and differences between the ESPs that we'd need to decide how to handle. For example, ActiveCampaign does almost all of its contact organization via lists, and uses segments mostly for automation campaigns that you set up through the ESP. So the use case for sending to segments via AC is less important than for the other ESPs. This and the fact that AC has no limit on lists per account mean that some large sites using AC have many, many lists, so the autocomplete UI becomes more useful for their lists vs. the other ESPs. Meanwhile, when sending content via Constant Contact, you send to either a list or a segment, and there's little hierarchical relationship between the two types (unlike with MC and AC, where segments and other sublist types are subsets of lists). But abstracting on the PHP class side only instead of both there and in the editor JS seems to me like it will make handling these nuances easier.

> Also, as a minor suggestion, in Mailchimp I think it would be ok to leave the Audience field as a select box, and do the autocomplete only for the segment/tags/groups.

This works for Mailchimp, but AC has no limit on lists per account and CC has a limit of 1000, so sites using those ESPs are more likely to have lots and lots of lists, making the select dropdown less useful for them. I'd love to keep the UI as consistent as possible. Maybe we could show up to 10 list suggestions alongside the autocomplete like we do with the `AutocompleteWithSuggestions` component? I initially started trying to implement this feature using that component, but it's designed for WP_Query requests in a way that made it less useful for handling arbitrary data like these newsletter contact lists. However, the UI itself could work well for audiences:

Screenshot 2024-07-23 at 11 24 32 AM

Thoughts?

leogermani:
> Lists and sublists
> At some point we'll still have to deal with the different APIs and data structures for each ESP, but I'm thinking we should be able to abstract those things deeper down in the PHP classes so that the API endpoints to fetch lists and sublists (and the data they return to the editor) can be the same no matter what ESP is active

It makes sense. In the case of Mailchimp, the `sublists` endpoint would have to make 3 requests to MC api and combine the results. Much better than what we have now, but combining results from different sources in a search very often bring us to some edge cases that leave us with some unreachable data. For example, if you have tags and groups with a similar name, and you are looking for a specific group, you might not get to it easily because the results of groups will get in the way of the tags and vice versa. If we want to show 10 results, and there are 3 sources, it will be only 3 results from each source...

But anyways, it might not be a big issue, just a usability inconvenience that will make you type more letters to find what you want.

If each provider informed its "sublists types" we could present the select box before the auto complete.

I'm not sure it would be a better experience, I'm just brain storming here.

> This works for Mailchimp, but AC has no limit on lists per account and CC has a limit of 1000, so sites using those ESPs are more likely to have lots and lots of lists, making the select dropdown less useful for them. I'd love to keep the UI as consistent as possible. Maybe we could show up to 10 list suggestions alongside the autocomplete like we do with the AutocompleteWithSuggestions component?

Makes sense. An initial set of suggestions would be awesome imo, but if that's not something that's easy we can live this for later in the epic

kevinzweerink:
Hey @dkoo, just catching up here. Let me try to articulate the reasoning for this change to make sure I understand:
1. We want to unify the UI as much as possible between different ESPs.
2. The dropdown select component is theoretically not up to the task, as there could be thousands of lists or sublists coming from an ESP.

Is that right?

My main question is whether any of our publishers actually have enough lists or sublists for the dropdown component to become an issue for them. I'm wondering whether we might be solving a problem that wouldn't actually impact any of our publishers, and in doing so making unnecessary compromises for the user experience. I think the autocomplete is probably more confusing and opaque than a dropdown.

I also don't really understand why we're hiding the top-level List selection behind an "Edit" button — it seems harder to use, and visually looks more complicated with all the different types of text and UI going on.

Can you help me understand better what this is accomplishing? I feel like I might be missing something here.

dkoo:
@kevinzweerink thanks for the feedback!

> My main question is whether any of our publishers actually have enough lists or sublists for the dropdown component to become an issue for them.

Yes, the idea for the autocomplete came about because several publishers have hundreds of lists or sublists such that the dropdown becomes unusable.

> I also don't really understand why we're hiding the top-level List selection behind an "Edit" button — it seems harder to use, and visually looks more complicated with all the different types of text and UI going on.

That might be confusing based on the screenshots, but all of the UIs start with just the autocomplete field like this:

Screenshot 2024-07-22 at 1 16 46 PM

When you start typing, the autocomplete field shows suggestions matching your input:

Screenshot 2024-07-23 at 1 09 30 PM

Once you choose one of the suggestions, the autocomplete field hides itself and shows the item you selected with an Edit button (and a Reset button if the API supports it):

Screenshot 2024-07-23 at 1 09 46 PM

I'm very open to more elegant ways of showing the selected state.

EDIT to add:

> 1. We want to unify the UI as much as possible between different ESPs.
> 2. The dropdown select component is theoretically not up to the task, as there could be thousands of lists or sublists coming from an ESP.

Yes, these are basically the main reasons we want to transition away from the very different UIs that we have across ESPs now (some of which are a series of dropdowns, others of which are radio buttons or checklists or some combination thereof). The ultimate goal is to unify not just the UIs but also the data they're consuming, as @leogermani and I were discussing above, but right now that's not feasible because the editor UIs are currently so different between the different ESPs.

kevinzweerink:
@dkoo and I talked this issue through on Zoom, and we've landed on a couple improvements:
1. The autocomplete fields will auto-populate with N most recently used values — this should help offset some of the cognitive load they introduce by forcing users to recall the name of the object they are looking for unassisted.
2. The label for the sublist autocomplete will be changed to "Filter by tag, segment, or group" (specific language can be tailored to ESP terminology). Adding "filter by" helps make it clear to the user that this is a subtractive selection, not additive.
3. Once selected, the panel contents will be displayed as shown in the design below.

CleanShot 2024-07-24 at 13 44 12@2x
This removes some of the typographic noisiness of brackets and parentheses, and gives the user a natural language explanation of the total recipients (when available). The natural language summary can be constructed this way: 

> This newsletter will be sent to all [optional number] members of the [list name] [“audience”|”list”].

> This newsletter will be sent to all [optional number] members of the [list name] [“audience”|”list”] who are part of the [sublist name] [“tag”|”list”|”segment”|”group”].

dkoo:
@kevinzweerink a49d71b implements the design updates as discussed.

@leogermani this is ready for a code review!

leogermani:
It's working fine for CC and AC. For Mailchimp I'm getting this:

image

I'm still concerned on how we are going to handle multiple types of sublists at the same time in tha backend, and the quality of the results it will generate - but I guess we can revisit this later if needed.

dkoo:
@leogermani 990441f should fix the Mailchimp issue.

leogermani (approved):
A very minor comment: When there is only one contact in the selected segment, we could omit the work "all" (we'll need to use `_n()` instead of `__()`

src/service-providers/active_campaign/ProviderSidebar.js
```diff
@@ -146,57 +218,69 @@ const ProviderSidebarComponent = ( {
...
+					dangerouslySetInnerHTML={ {
+						__html: sprintf(
+							// Translators: %1$s is the number of members, %2$s is the item name, %3$s is the item type, %4$s is the parent item name and type (if any).
+							__(
+								'This newsletter will be sent to all %1$s in the %2$s %3$s%4$s.',
+								'newspack-newsletters'
+							),
+							selectedSegment?.details || ( ! selectedSegment && selectedList?.details )
+								? `<strong>${ selectedSegment?.details || selectedList?.details }</strong> `
+								: __( 'contacts', 'newspack-newsletters' ),
```
leogermani: Actually, looking at this again, this is not a very good way to localize the string.

The sentence should be whole, breaking it up and leaving "contacts" as a separate part of the string makes it really hard for translators to figure out the context. 

Tip: https://make.wordpress.org/core/2020/07/17/introducing-createinterpolateelement/

leogermani (changes requested):

src/service-providers/active_campaign/ProviderSidebar.js
leogermani: but in this case I think it's ok to just add the `<strong>` tags to the translatable string

dkoo: Thanks for the suggestion—I think you're right, since we're rendering HTML via `dangerouslySetInnerHTML` anyway. I'm always a little unsure of best practices for interpolating HTML with translatable strings in React...

dkoo: @leogermani 0efabc8 updates the handling of contact counts and the translatable strings in the summaries.

src/service-providers/active_campaign/ProviderSidebar.js
```diff
@@ -124,14 +171,68 @@ const ProviderSidebarComponent = ( {
...
+					__html: sprintf(
+						// Translators: %1$s is the number of members, %2$s is the item name, %3$s is the item type, %4$s is the subitem name and type (if any).
+						__(
+							'This newsletter will be sent to <strong>%1$s</strong> in the <strong>%2$s</strong> %3$s%4$s.',
+							'newspack-newsletters'
+						),
+						! selectedSegment && selectedList?.count // AC doesn't provide contact counts for segments, so we only want to show a count if a list is selected without a segment.
+							? sprintf(
+									// Translators: %d is the number of contacts in the list.
+									_n( '%d contact', '%d contacts', selectedList.count, 'newspack-newsletters' ),
```
leogermani: this is still breaking the translatable string into chunks, making parts of the phrase live in a complete separate string, which is not a good practice.

When translators translate they are not necessarily looking at the code, only at the strings, and they won't know that one string will be concatenated inside another. Also, if a string reads exactly the same in another context, they will share the same translations, but maybe in some languages the whole thing would be better phrased in another way.

src/service-providers/active_campaign/ProviderSidebar.js
```diff
@@ -124,14 +171,68 @@ const ProviderSidebarComponent = ( {
...
+									selectedList.count.toLocaleString()
+							  )
+							: __( 'all contacts', 'newspack-newsletters' ),
+						selectedList.name,
+						selectedList.typeLabel.toLowerCase(),
+						selectedSegment
+							? sprintf(
+									// Translators: %1$s is the parent item name, %2$s is the parent item type.
+									__( ' who are part of the %1$s %2$s', 'newspack-newsletters' ),
```
leogermani: here you are also breaking a phrase into separate translatable strings.

If there are many conditions that can change the elements of the text, bring the checks up in the code and make one completely independent string for each possible situation, even if they have parts that will be identical and you'll have some duplicate code. That's the best practice for i18n.

The main idea is that a phrase should be self contained and have the complete sentence. No "ifs" in the middle. This not only makes the job easier for translators but also gives them flexibility to rephrase things as needed, to adapt and actually localize the string - and not only do a literal translation

leogermani (changes requested):
Since we are here looking at i18n best practices, let's take this right

dkoo:
@leogermani b424eac redoes the summaries so that they're always full strings.

src/newsletter-editor/sidebar/send-to.js
```diff
@@ -0,0 +1,146 @@
...
+			) }
+
+			{ selectedItem && ! isEditing ? (
+				<>
+					<p className="newspack-newsletters__send-to-details">
+						{ selectedItem.name }
+						<span>
+							{ sprintf(
+								// Translators: %1$s is the item type, %2$s is details about the item.
+								__( '%1$s %2$s', 'newspack-newsletters' ),
```
leogermani: I missed this on my first review. But this is something impossible to be translated :)

src/newsletter-editor/utils.js
```diff
@@ -60,3 +61,25 @@ export const refreshEmailHtml = async ( postId, postTitle, postContent ) => {
...
+/**
+ * Get a label for the Send To autocomplete field.
+ *
+ * @param {Object} item A list or sublist item.
+ * @return {string} The autocomplete suggestion label for the item.
+ */
+export const getSuggestionLabel = item => {
+	return sprintf(
+		// Translators: %1$s is the item type, %2$s is the item name, %3$s is more details about the item, if available.
+		__( '[%1$s] %2$s %3$s', 'newspack-newsletters' ),
```
leogermani: Another untranslatable string. Maybe in this case all you need is a simple concatenation, without any i18n function

src/service-providers/active_campaign/ProviderSidebar.js
```diff
@@ -124,14 +171,70 @@ const ProviderSidebarComponent = ( {
...
+	}
+
+	const renderSelectedSummary = () => {
+		if ( ! selectedList ) {
+			return null;
+		}
+
+		const summary = selectedSegment
+			? sprintf(
+					// Translators: A summary of which list and sublist the campaign is set to send to, and the total number of contacts, if available.
```
leogermani:
```suggestion
					// Translators: A summary of which list and sublist the campaign is set to send to, and the total number of contacts, if available. %1$s is the label of the list (ex: Main), %2$s is the label for the type of the list (ex: "list" on Active Campaign and "audience" on Mailchimp). %3$s is the label of the sublist (ex: "paid customers"), and %4$s is the label for the sublist type (ex: tag, group or segment)
```

src/service-providers/active_campaign/ProviderSidebar.js
```diff
@@ -124,14 +171,70 @@ const ProviderSidebarComponent = ( {
...
+					'This newsletter will be sent to <strong>all contacts</strong> in the <strong>%1$s</strong> %2$s who are part of the <strong>%3$s</strong> %4$s.',
+					selectedList.name,
+					selectedList.typeLabel.toLowerCase(),
+					selectedSegment.name,
+					selectedSegment.typeLabel.toLowerCase()
+			  )
+			: sprintf(
+					// Translators: A summary of which list the campaign is set to send to, and the total number of contacts, if available.
```
leogermani:
```suggestion
					// Translators: A summary of which list the campaign is set to send to, and the total number of contacts, if available. %1$s is the number of contacts. %2$s us the label of the list (ex: Main), %3$s is the label for the type of the list (ex: "list" on Active Campaign and "audience" on Mailchimp).
```

src/service-providers/constant_contact/ProviderSidebar.js
```diff
@@ -89,8 +136,43 @@ const ProviderSidebar = ( {
...
-	const { current_status: status } = campaign || {};
-	if ( 'DRAFT' !== status ) {
+	const renderSelectedSummary = () => {
+		if ( ! selected ) {
+			return null;
+		}
+
+		const summary = ! selected.hasOwnProperty( 'count' )
+			? sprintf(
+					// Translators: A summary of which list or segment the campaign is set to send to, and the total number of contacts, if available.
```
leogermani:
```suggestion
					// Translators: A summary of which list or segment the campaign is set to send to, and the total number of contacts, if available.  %1$s us the label of the list (ex: Main), %2$s is the label for the type of the list (ex: "list" on Active Campaign and "audience" on Mailchimp).
```

src/service-providers/constant_contact/ProviderSidebar.js
```diff
@@ -89,8 +136,43 @@ const ProviderSidebar = ( {
...
+					'This newsletter will be sent to <strong>all contacts</strong> in the <strong>%1$s</strong> %2$s.',
+					selected.name,
+					selected.typeLabel.toLowerCase()
+			  )
+			: sprintf(
+					// Translators: A summary of which list the campaign is set to send to, and the total number of contacts, if available.
```
leogermani:
```suggestion
					// Translators: A summary of which list the campaign is set to send to, and the total number of contacts, if available.  %1$s is the number of contacts. %2$s us the label of the list (ex: Main), %3$s is the label for the type of the list (ex: "list" on Active Campaign and "audience" on Mailchimp).
```

src/service-providers/mailchimp/ProviderSidebar.js
```diff
@@ -213,6 +191,47 @@ const ProviderSidebar = ( {
...
 	}, [ stringifiedLayoutDefaults.length ] );
 
+	const renderSelectedSummary = () => {
+		if ( ! selectedAudience ) {
+			return null;
+		}
+
+		const summary = selectedSubAudience?.count
+			? sprintf(
+					// Translators: A summary of which list and sublist the campaign is set to send to, and the total number of contacts, if available.
```
leogermani:
```suggestion
					// Translators: A summary of which list and sublist the campaign is set to send to, and the total number of contacts, if available.  %1$s is the number of contacts. %2$s is the label of the list (ex: Main), %3$s is the label for the type of the list (ex: "list" on Active Campaign and "audience" on Mailchimp). %4$s is the label of the sublist (ex: "paid customers"), and %5$s is the label for the sublist type (ex: tag, group or segment)
```

src/service-providers/mailchimp/ProviderSidebar.js
```diff
@@ -213,6 +191,47 @@ const ProviderSidebar = ( {
...
+						'newspack-newsletters'
+					),
+					selectedSubAudience.count.toLocaleString(),
+					selectedAudience.name,
+					selectedAudience.typeLabel.toLowerCase(),
+					selectedSubAudience.name,
+					selectedSubAudience.typeLabel.toLowerCase()
+			  )
+			: sprintf(
+					// Translators: A summary of which list the campaign is set to send to, and the total number of contacts, if available.
```
leogermani:
```suggestion
					// Translators: A summary of which list the campaign is set to send to, and the total number of contacts, if available. %1$s is the number of contacts. %2$s is the label of the list (ex: Main), %3$s is the label for the type of the list (ex: "list" on Active Campaign and "audience" on Mailchimp).
```

leogermani (changes requested):
@dkoo thanks for the changes. I hope you agree that it looks so much more readable, not only for translators but also for developers.

I'm sorry but I found 2 other strings that need to be fixed, and left some suggestions to improve the doc block to help translators even more

src/newsletter-editor/utils.js
dkoo: Agreed, there's no reason this whole string would need to be translated. Just the `(n contacts)` string in this case. Updated in b33ab3d

src/newsletter-editor/sidebar/send-to.js
dkoo: Thanks for highlighting—should make more sense in b33ab3d. There's probably no need for this whole thing to be translated, only the `Resetting [item type]...` and `(n contacts)` strings.

src/newsletter-editor/utils.js
leogermani: Did you forget to push?

dkoo: Yes, yes I did

leogermani (approved):

dkoo:
@leogermani thanks for the guidance in improving the handling of translated strings!

src/newsletter-editor/sidebar/send-to.js
```diff
@@ -0,0 +1,137 @@
...
+				<>
+					<FormTokenField
+						disabled={ isUpdating }
+						label={ formLabel || __( 'Select a list', 'newspack' ) }
+						maxSuggestions={ 10 }
+						onChange={ items => {
+							setError( false );
+							setIsUpdating( true );
+							onChange( items )
+								.catch( e => {
```
miguelpeixe, line 55: If I select an item, click to edit, and select the same item again I get the following error and the input fails to change:

```
Cannot read properties of undefined (reading 'catch')
```

dkoo: I think we'll fix this in the next round of changes in #1605

PR #1582, opened 2024-07-29
Release Jul 29

adekbadek:
A new release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,30 @@
+## [2.19.1](https://github.com/Automattic/newspack-newsletters/compare/v2.19.0...v2.19.1) (2024-07-09)
+
+
+### Bug Fixes
+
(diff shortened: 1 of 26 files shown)


dkoo (approved):
It looks like most of the changes in the commit list were actually (accidentally?) released as part of [this hotfix](https://github.com/Automattic/newspack-newsletters/pull/1556) last week. So this merge may not be needed at all.

adekbadek:
That's true, this hotfix accidentally contained the whole alpha branch. I'll merge this PR anyway, let's see how this edge case will be handled.

adekbadek:
And so we do have a new release, because the [hotfix commit](https://github.com/Automattic/newspack-newsletters/commit/6d028ffc7e34101dd1fb30fd5c1d04736ce4ce78) was a squash-merge. But it's for the better, the changelog and releases page now reflect reality.

PR #1587, opened 2024-07-31
fix(layouts): settings defaults from layout (mailchimp)

adekbadek:
### All Submissions:

Fixes an issue with setting the default sender & audience data when Mailchimp is set as the ESP. 

1. On `trunk` or `release`
1. Set Mailchimp as the ESP
2. Create a new newsletter, set sender data and choose an audience, save that as a new layout
3. Create a new newsletters, use the newly created layout
4. Observe two error notices in the editor and that the sender and audience are not filled
5. Switch to this branch, create a new newsletter from that same layout again, observe the sender and audience are filled, and no error notices are displayed

### Other information:

---
  - https://app.asana.com/0/0/1207937306459773

### DIFF
--- a/src/service-providers/mailchimp/ProviderSidebar.js
+++ b/src/service-providers/mailchimp/ProviderSidebar.js
@@ -100,8 +100,43 @@ const ProviderSidebar = ( {
 		: [];
 	const folders = newsletterData?.folders || [];
 
+	const setDefaultsFromLayout = () => {
+		try {
+			const layoutDefaults = JSON.parse( stringifiedLayoutDefaults );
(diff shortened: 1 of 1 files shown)


miguelpeixe (approved):
If I understand correctly, the sender and list data from the layout required `fetchListsAndSegments()` before being set, correct?

Fixes the issue!

adekbadek:
> If I understand correctly, the sender and list data from the layout required fetchListsAndSegments() before being set, correct?

That's correct. Fetching lists and segments sets up the campaign in MC, as a side effect. And it also makes sense logically to set the audience (list and segments) from layout _after_ fetching the related data.

PR #1590, opened 2024-08-01
fix: human-readable timestamp of sent newsletters

dkoo:
### All Submissions:

Fixes a bug where the human-readable timestamp shown next to sent newsletters doesn't respect the site's local timezone.

1. Ensure your site has a time zone set in **Settings > General** that's different from GMT/Universal Time.
2. On `trunk`, send a newsletter.
3. Go back to the **Newsletters** post list and observe that the sent label says something inaccurate. For instance, my site is set to Denver/Mountain Time, and so seconds after sending the newsletter the human-readable timestamp reads like this:

Screenshot 2024-08-01 at 4 20 30 PM

4. Check out this branch and refresh. Confirm that the timestamp is now more accurate based on when the newsletter was actually sent:

Screenshot 2024-08-01 at 4 23 55 PM

### Other information:

---
  - https://app.asana.com/0/0/1207957618908225

### DIFF
--- a/includes/class-newspack-newsletters.php
+++ b/includes/class-newspack-newsletters.php
@@ -1175,7 +1175,8 @@ final class Newspack_Newsletters {
 
 		/** Detect meta that determines the sent state */
 		$sent         = get_post_meta( $post_id, 'newsletter_sent', true );
-		$is_published = 'publish' === get_post_status( $post_id ) || $sent;
+		$post_status  = get_post_status( $post_id );
+		$is_published = 'publish' === $post_status || 'private' === $post_status;
 		$publish_date = $is_published ? get_post_datetime( $post_id, 'date', 'gmt' )->getTimestamp() : 0;
 		if ( 0 < $sent && $sent === $publish_date ) {
 			return $sent;
--- a/includes/class-newspack-newsletters.php
+++ b/includes/class-newspack-newsletters.php
@@ -1181,7 +1181,6 @@ final class Newspack_Newsletters {
 			return $sent;
 		}
 
-		/** Legacy check for sent/publish newsletters without meta. */
 		if ( $publish_date ) {
 			self::set_newsletter_sent( $post_id, $publish_date );
 			return $publish_date;
--- a/includes/class-newspack-newsletters.php
+++ b/includes/class-newspack-newsletters.php
@@ -1175,7 +1175,7 @@ final class Newspack_Newsletters {
 
 		/** Detect meta that determines the sent state */
 		$sent         = get_post_meta( $post_id, 'newsletter_sent', true );
-		$is_published = 'publish' === get_post_status( $post_id );
+		$is_published = 'publish' === get_post_status( $post_id ) || $sent;
 		$publish_date = $is_published ? get_post_datetime( $post_id, 'date', 'gmt' )->getTimestamp() : 0;
 		if ( 0 < $sent && $sent === $publish_date ) {
 			return $sent;
--- a/includes/class-newspack-newsletters.php
+++ b/includes/class-newspack-newsletters.php
@@ -546,8 +546,6 @@ final class Newspack_Newsletters {
 
 		$post_status = get_post_status_object( $post->post_status );
 		$sent        = self::is_newsletter_sent( $post->ID );
-		$is_public   = get_post_meta( $post->ID, 'is_public', true );
-
 		if ( $sent ) {
 			$time_diff = time() - $sent;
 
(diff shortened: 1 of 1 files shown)


includes/class-newspack-newsletters.php
```diff
@@ -1176,17 +1174,17 @@ public static function is_newsletter_sent( $post_id ) {
 		}
 
 		/** Detect meta that determines the sent state */
-		$sent = get_post_meta( $post_id, 'newsletter_sent', true );
-		if ( 0 < $sent ) {
+		$sent         = get_post_meta( $post_id, 'newsletter_sent', true );
+		$is_published = 'publish' === get_post_status( $post_id );
+		$publish_date = $is_published ? get_post_datetime( $post_id, 'date', 'gmt' )->getTimestamp() : 0;
```
miguelpeixe: This only considers the legacy check, which doesn't account for private sent newsletters (which has post status set to `private`).

```suggestion
		$sent         = get_post_meta( $post_id, 'newsletter_sent', true );
		$is_published = 'publish' === get_post_status( $post_id ) || $sent;
		$publish_date = $is_published ? get_post_datetime( $post_id, 'date', 'gmt' )->getTimestamp() : 0;
```

includes/class-newspack-newsletters.php
```diff
@@ -1176,17 +1174,17 @@ public static function is_newsletter_sent( $post_id ) {
...
+		if ( 0 < $sent && $sent === $publish_date ) {
 			return $sent;
 		}
 
 		/** Legacy check for sent/publish newsletters without meta. */
```
miguelpeixe:
```suggestion
```

With this change, this comment is no longer correct.

miguelpeixe (changes requested):
Works well after the suggested tweak!

includes/class-newspack-newsletters.php
dkoo: You're right, thanks!

miguelpeixe (approved):

miguelpeixe:
Not sure why tests started failing now 🤔

dkoo:
> Not sure why tests started failing now 🤔

It's because `$sent` could have been `false` if the post didn't have the `newsletter_sent` meta. 442c522 should fix this by explicitly checking for `publish` or `private` post status. Also, even if a post already has the `newsletter_sent` meta, it could be the wrong date, so we want to overwrite it in that case if it doesn't equal the actual publish date timestamp.

includes/class-newspack-newsletters.php
```diff
@@ -1176,17 +1174,17 @@ public static function is_newsletter_sent( $post_id ) {
...
 			return $sent;
 		}
 
-		/** Legacy check for sent/publish newsletters without meta. */
-		if ( 'publish' === get_post_status( $post_id ) ) {
-			$post = get_post( $post_id );
-			$sent = strtotime( $post->post_date );
-			self::set_newsletter_sent( $post_id, $sent );
-			return $sent;
+		if ( $publish_date ) {
```
miguelpeixe, line 1185: It's odd that we're using the $publish_date to determine whether the newsletter should be marked as sent. Why does this PR change the original behavior?

dkoo: It's because for any sites that exhibit the bug described in the PR description (where the `newsletter_sent` timestamp stored in meta doesn't match the actual sent time due to the timezone difference) we want to overwrite the existing `newsletter_sent` meta with the actual timezone-correct timestamp. To walk through the code step-by-step:

1. `$sent         = get_post_meta( $post_id, 'newsletter_sent', true );` - we check if the already has a sent timestamp saved as post meta
2. `$post_status  = get_post_status( $post_id );` and `$is_published = 'publish' === $post_status || 'private' === $post_status;` - this uses the post status to determine whether the newsletter was sent
3. `$publish_date = $is_published ? get_post_datetime( $post_id, 'date', 'gmt' )->getTimestamp() : 0;` - if it was sent, we get the publish date as a timestamp using [`get_post_datetime`](https://developer.wordpress.org/reference/functions/get_post_datetime/) (which returns a `DateImmutable` object and which will return a timestamp correctly calculated based on the site's timezone setting)
4. `if ( 0 < $sent && $sent === $publish_date ) { return $sent; }` - if the timestamp saved to post meta already matches the publish date, then we're good—just return that
5. If not, then we have the correct publish date from step 3, so save that as the `newsletter_sent` meta value for the next time.

As always I'm open to improvements if you have ideas on how to accomplish this better!

miguelpeixe: Ah, I see! Sorry for my confusion!

miguelpeixe (approved):

PR #1593, opened 2024-08-02
feat: introduce "subscribe" method

miguelpeixe:
### All Submissions:

This PR implements a new `Newspack_Newsletters_Contacts::subscribe()` method to differentiate contact updates from intentional newsletter subscriptions. Contact updates should continue to use `::upsert()`, also called by the new `::subscribe()` method.

The async strategy was moved from `::upsert()` to `::subscribe()`, given it's a strategy aimed at improving the UX for the newsletter subscription form. Most `::upsert()` calls are already performed asynchronously via data events, this method should be predictable and up to the caller to handle performance.

Unit test were also updated to include the sync and async usage of the new `::subscribe()` method. The test class was renamed and refactored a bit to reflect its relationship with the `Newspack_Newsletters_Contacts` class.

Related PRs:
- https://github.com/Automattic/newspack-plugin/pull/3304
- https://github.com/Automattic/newspack-blocks/pull/1827

1. Check out this branch, https://github.com/Automattic/newspack-plugin/pull/3304, and https://github.com/Automattic/newspack-blocks/pull/1827
2. Make sure you have Newspack Newsletters configured and properly connected to an ESP
3. Add a Newsletter Subscription Form and Reader Registration (with newsletter lists) blocks to a page
4. In a fresh session, register using the Reader Registration block and watch the logs from https://github.com/Automattic/newspack-manager/pull/240
5. Confirm the subscription goes through and includes the context for each action
6. In another fresh session, repeat step 4 and 5 with the Newsletter Subscription Form block
7. Using the Donate Block, make a donation with your newly created account
8. Confirm, via logs and the ESP dashboard, that data events is able to update the contact metadata with the donation data

### Other information:

### DIFF
--- a/includes/class-newspack-newsletters-contacts.php
+++ b/includes/class-newspack-newsletters-contacts.php
@@ -17,11 +17,6 @@ class Newspack_Newsletters_Contacts {
 	/**
 	 * Upserts a contact to lists.
 	 *
-	 * A contact can be added asynchronously, which means the request will return
-	 * immediately and the contact will be added in the background. In this case
-	 * the response will be `true` and the caller must handle it optimistically.
-	 * NEWSPACK_NEWSLETTERS_ASYNC_SUBSCRIPTION_ENABLED must be defined as true for this feature to be available.
-	 *
 	 * @param array          $contact {
 	 *          Contact information.
 	 *
@@ -30,11 +25,10 @@ class Newspack_Newsletters_Contacts {
 	 *    @type string[] $metadata Contact additional metadata. Optional.
 	 * }
 	 * @param string[]|false $lists   Array of list IDs to subscribe the contact to. If empty or false, contact will be created but not subscribed to any lists.
-	 * @param bool           $async   Whether to add the contact asynchronously. Default is false.
 	 *
 	 * @return array|WP_Error|true Contact data if it was added, or error otherwise. True if async.
 	 */
-	public static function upsert( $contact, $lists = false, $async = false ) {
+	public static function upsert( $contact, $lists = false ) {
 		if ( ! is_array( $lists ) && false !== $lists ) {
 			$lists = [ $lists ];
 		}
@@ -56,11 +50,6 @@ class Newspack_Newsletters_Contacts {
 		$provider = Newspack_Newsletters::get_service_provider();
 		if ( empty( $provider ) ) {
 			return new WP_Error( 'newspack_newsletters_invalid_provider', __( 'Provider is not set.' ) );
-		}
-
-		if ( defined( 'NEWSPACK_NEWSLETTERS_ASYNC_SUBSCRIPTION_ENABLED' ) && NEWSPACK_NEWSLETTERS_ASYNC_SUBSCRIPTION_ENABLED && true === $async ) {
-			Newspack_Newsletters_Subscription::add_subscription_intent( $contact, $lists );
-			return true;
 		}
 
 		if ( false !== $lists ) {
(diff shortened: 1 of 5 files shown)


includes/class-newspack-newsletters-contacts.php
```diff
@@ -17,24 +17,19 @@ class Newspack_Newsletters_Contacts {
...
 	 * }
-	 * @param string[]|false $lists   Array of list IDs to subscribe the contact to. If empty or false, contact will be created but not subscribed to any lists.
-	 * @param bool           $async   Whether to add the contact asynchronously. Default is false.
+	 * @param string[]|false $lists            Array of list IDs to subscribe the contact to. If empty or false, contact will be created but not subscribed to any lists.
+	 * @param array|null     $existing_contact Optional existing contact data.
 	 *
 	 * @return array|WP_Error|true Contact data if it was added, or error otherwise. True if async.
 	 */
-	public static function upsert( $contact, $lists = false, $async = false ) {
+	public static function upsert( $contact, $lists = false, $existing_contact = null ) {
```
miguelpeixe: The async strategy was removed from this method to be handled by the new `subscribe_contact` method. The async strategy is meant to improve UX exclusive to newsletter subscriptions and should not be available for every upsert.

The `$existing_contact` was added as a parameter in case the caller already fetched the data, which is the case for the new `subscribe_contact` method.

includes/class-newspack-newsletters-contacts.php
```diff
@@ -113,7 +106,57 @@ public static function upsert( $contact, $lists = false, $async = false ) {
...
+		 *
+		 *    @type string   $email    Contact email address.
+		 *    @type string   $name     Contact name. Optional.
+		 *    @type string[] $metadata Contact additional metadata. Optional.
+		 * }
+		 * @param string[]|false      $lists    Array of list IDs to subscribe the contact to.
+		 * @param array|WP_Error      $result   Array with data if the contact was added or error if failed.
+		 * @param bool                $is_updating Whether the contact is being updated. If false, the contact is being created.
+		 */
+		do_action( 'newspack_newsletters_upsert', $provider->service, $contact, $lists, $result, $is_updating );
```
miguelpeixe: The `newspack_newsletters_add_contact` is now split into this `newspack_newsletters_upsert` and `newspack_newsletters_contact_subscribed`, meant for actual newsletter subscriptions

includes/class-newspack-newsletters-subscription.php
```diff
@@ -489,6 +554,23 @@ public static function process_subscription_intents( $intent_id = null ) {
...
+				 *
+				 *    @type string   $email    Contact email address.
+				 *    @type string   $name     Contact name. Optional.
+				 *    @type string[] $metadata Contact additional metadata. Optional.
+				 * }
+				 * @param string[]|false      $lists    Array of list IDs to subscribe the contact to.
+				 * @param array|WP_Error      $result   Array with data if the contact was added or error if failed.
+				 * @param bool|null           $is_updating Whether the contact is being updated. If false, the contact is being created.
+				 */
+				do_action( 'newspack_newsletters_contact_subscribed', $provider->service, $contact, $lists, $result, $is_updating );
```
miguelpeixe: The hook is triggered again here to handle the successful async subscription. I take suggestions to avoid duplication. Maybe a `finish_subscription` method?

includes/class-newspack-newsletters-subscription.php
```diff
@@ -473,11 +478,16 @@ public static function process_subscription_intents( $intent_id = null ) {
...
+
+			$contact = $intent['contact'];
+			$email   = $contact['email'];
+			$lists   = $intent['lists'];
+			$context = $intent['context'];
+
+			$existing_contact = self::get_contact_data( $contact['email'], true );
+			$is_updating      = \is_wp_error( $existing_contact ) ? false : true;
+
+			$result = Newspack_Newsletters_Contacts::upsert( $contact, $lists, $context . ' (ASYNC)', $existing_contact );
```
leogermani: Why not call `::subscribe` here?

miguelpeixe: The intent is initiated by ::subscribe(). I don't like the circular pattern this would create.

leogermani: It looks perfectly fine for me, as it avoids duplication of code. It's not circular.

`subscribe` with `async` initializes the request. But the request gets short circuited and delayed. Then, when the subscription intent is processed, the request is resumed, by calling `subscribe` with `async=false`. It looks beautiful!

I actually implemented this pattern in one of the first commits of the epic branch. The process subscription intent was doing some duplication and this looks much simpler and straight forward. We are calling `subscribe` with `async` explicitly set to `false` - there is no way of an infinite loop or anything.

miguelpeixe: I'm convinced! Updated in 08b930e592a9cf1c7cdf8b51836d839507588ca1

leogermani (approved):
Looks and works well.

One thing I saw in my tests that we might want to consider:

If you:
* Register (without subscribing)
* you are added as transactional contact
* Then you verify your email
* Go to My Account -> Newsletters
* And subscribe to a newsltter from there

I don't think the `update_lists`method is triggering the `subscribe` hook. And I don't know if we want to

miguelpeixe:
Good catch! I'll look into now

PR #1574, opened 2024-07-16
Add/logging

leogermani:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/trunk/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Adds logging to contact syncing

### How to test the changes in this Pull Request:

See https://github.com/Automattic/newspack-manager/pull/240

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

PR #1594, opened 2024-08-02
feat: add php sniffer for contact methods

leogermani:
### All Submissions:

In the effort of consolidating our data flows, we want to enforce that every contact update goes through the same route, starting at `Newspack_Newsletters_Contacts` class to make sure we have the propper logging.

However, we can't make all the internal methods private or protected.

This PHP Sniffer helps us avoid adding any new code that will bypass our intended flow. We'll need to add this sniffers to other Newspack plugins as well. - TBD how to do that

1. checkout this branch
2. run `./vendor/bin/phpcs`
3. Confirm you see some error in 2 files. (These errors are fixed in https://github.com/Automattic/newspack-newsletters/pull/1574)
4. Does the message look good?
5. Try to add a method to one of the forbidden methods anywhere in the code.
6. Make sure the sniffer catches it

PS - the linting errors in the CI job are expected until we merge https://github.com/Automattic/newspack-newsletters/pull/1574

### Other information:

---
  - https://app.asana.com/0/0/1207942575891805


miguelpeixe (changes requested):
Super cool!

One issue caught by the linter is from [src/blocks/subscribe/index.php](https://github.com/Automattic/newspack-newsletters/blob/bf3e26880368650baa5af7245ef0991481fdfc17/src/blocks/subscribe/index.php#L501), which doesn't call the forbidden method from a class. This causes the sniffer to throw a misleading error:

```
FILE: /home/circleci/project/src/blocks/subscribe/index.php
----------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
----------------------------------------------------------------------
 501 | ERROR | Method
     |       | Newspack_Newsletters_Subscription::add_contact()
     |       | should not be called from class
     |       | NewspackNewslettersContactsMethodsSniff. These methods
     |       | are for internal use only. Use methods in
     |       | Newspack_Newsletters_Contacts class instead.
     |       | (..NewspackNewslettersContactsMethods.ForbiddenContactsMethods)
----------------------------------------------------------------------
```

Given it includes the filename, it's not too bad, but is there a way to catch that and adjust the messaging?

leogermani:
> Super cool!
> 
> One issue caught by the linter is from [src/blocks/subscribe/index.php](https://github.com/Automattic/newspack-newsletters/blob/bf3e26880368650baa5af7245ef0991481fdfc17/src/blocks/subscribe/index.php#L501), which doesn't call the forbidden method from a class. This causes the sniffer to throw a misleading error:
> 
> ```
> FILE: /home/circleci/project/src/blocks/subscribe/index.php
> ----------------------------------------------------------------------
> FOUND 1 ERROR AFFECTING 1 LINE
> ----------------------------------------------------------------------
>  501 | ERROR | Method
>      |       | Newspack_Newsletters_Subscription::add_contact()
>      |       | should not be called from class
>      |       | NewspackNewslettersContactsMethodsSniff. These methods
>      |       | are for internal use only. Use methods in
>      |       | Newspack_Newsletters_Contacts class instead.
>      |       | (..NewspackNewslettersContactsMethods.ForbiddenContactsMethods)
> ----------------------------------------------------------------------
> ```
> 
> Given it includes the filename, it's not too bad, but is there a way to catch that and adjust the messaging?

Thanks!

Yes, we could get the filename and add to the error message if we are not inside a class. But looking at it, I think it makes more sense to make it simpler and more generic. https://github.com/Automattic/newspack-newsletters/pull/1594/commits/99425a45ea54abbf599f1d2063c284a1b9ead5ac does it

miguelpeixe (approved):

leogermani:
We forgot to merge this to the epic before merging the epic to trunk and now, because of Squash, it was easier to rebuild this PR in a new branch -> https://github.com/Automattic/newspack-newsletters/pull/1611

PR #1603, opened 2024-08-08
Fix/status if new

leogermani:
### All Submissions:

Fixes and improves the handling of `status` and `status_if_new` in Mailchimp

Also fixes the Mailchimp Notes hook

Test different flows and make sure they work as expected.

Suggestions

1. Enable RAS
2. Set Mailchimp as ESP
3. As an anonymous reader, register to the site
4. Confirm the contact is added to Mailchimp as non-subscriber
5. Confirm there are Notes added to the contact with our logging
6. With the same user, make a Donation
7. Confirm the contact is updated on Mailchimp but is still not subscribed
8. Subscribe to a newsletter via newsletter subscription block
9. confirm that now the contact is subscribed
10. Make another donation
11. Make sure the contact is still subscribed

### Other information:

### DIFF
--- a/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php
+++ b/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php
@@ -1366,10 +1366,21 @@ final class Newspack_Newsletters_Mailchimp extends \Newspack_Newsletters_Service
 		}
 
 		$new_contact_status = 'subscribed';
-		if ( isset( $contact['metadata'] ) && ! empty( $contact['metadata']['status'] ) ) {
+		$status_if_new_informed = false;
+
+		if ( ! empty( $contact['metadata']['status_if_new'] ) ) {
+			$new_contact_status = $contact['metadata']['status_if_new'];
+			$status_if_new_informed = true;
+			unset( $contact['metadata']['status_if_new'] );
+		}
+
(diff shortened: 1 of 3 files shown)


includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php
```diff
@@ -1366,10 +1366,21 @@ public function add_contact( $contact, $list_id = false, $sublists = [] ) {
 		}
 
 		$new_contact_status = 'subscribed';
-		if ( isset( $contact['metadata'] ) && ! empty( $contact['metadata']['status'] ) ) {
+		$status_if_new_informed = false;
```
miguelpeixe: I find it confusing to add another variable to the mix. If `status_if_new` is provided, we can just add to `$update_payload`, same with `status`. I don't think we need `$new_contact_status` at all.

When populating those items, the developer should be as worried as they would be when actually making the Mailchimp request.

leogermani: yes, that was because we started handling statuses before handling metadata, but then would only add to the payload later, because the payload variable didn't exist at this point.

I changed that and it looks simpler now

miguelpeixe (approved):
Working as described!

PR #1606, opened 2024-08-09
fix(editor): render notice on sync error

miguelpeixe:
### All Submissions:

1200530782742699-as-1207981078972791

Creates an endpoint to retrieve the sync error message to be fetched once the editor has finished saving a post.

This PR also cleans up some unused react hooks from the newsletter editor.

1. Configure the plugin to use Mailchimp
2. Draft a new newsletter, save and confirm the draft is synced to Mailchimp
3. Add the following to line `784` of `includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php`

```php
			throw new Exception( 'Test Error.' );
```

4. Save the draft campaign again and confirm the notice appears after saving
5. Refresh the page and confirm the notice doesn't render
6. Save again and confirm it reappears
7. Without refreshing the editor, remove the added line, make changes to the draft and save once again
8. Confirm the notice disappears after saving and the changes are synced to Mailchimp

### Other information:


leogermani (changes requested):
It works but I have to save the draft twice before the error shows up and also have to save it twice for the error to disappear.

This is the only blocking comment, but since I'm here, I few notes on the implementation.

`get_transient_name` is a pretty generic method name. Transient for what? I think it should be more specific.

And actually, instead of just having a method to get the transient name, why not have methods to `[set|get|clear]_sync_errors` instead of having that logic spread across all the other methods. Looks cleaner to me and also makes it easier to change the approach in the future if needed.

For example: Why transient and not post meta? Why tied to a specific user, if the sync affects the campaign itself? etc...

miguelpeixe:
> It works but I have to save the draft twice before the error shows up and also have to save it twice for the error to disappear.

This was happening on 716213fb15e8d456f7900600f918e3665f78387e, before I changed strategy with 85fab56d54aca50e0cda31f683df9fa80006b498. Can you confirm you had the latest changes when you tested?

> get_transient_name is a pretty generic method name. Transient for what? I think it should be more specific.
> 
> And actually, instead of just having a method to get the transient name, why not have methods to [set|get|clear]_sync_errors instead of having that logic spread across all the other methods. Looks cleaner to me and also makes it easier to change the approach in the future if needed.

This transient implementation to handle sync errors has been part of Newsletters for years, I don't think this PR's goal is to refactor that. Nonetheless, I wholeheartedly agree that the API could be better.

> For example: Why transient and not post meta? Why tied to a specific user, if the sync affects the campaign itself? etc...

I understand that using a transient and adding the user ID to its key highlights the ephemeral nature of the error message. It's only meant to tell the editor, at that moment, that a specific attempt to sync has failed.

leogermani:
> This was happening on https://github.com/Automattic/newspack-newsletters/commit/716213fb15e8d456f7900600f918e3665f78387e, before I changed strategy with https://github.com/Automattic/newspack-newsletters/commit/85fab56d54aca50e0cda31f683df9fa80006b498. Can you confirm you had the latest changes when you tested?

Yes, I am. Just tested again.

> This transient implementation to handle sync errors has been part of Newsletters for years

Ah, I didn't realize. I thought it had been added in the recent changes that added the errors on refresh

miguelpeixe:
Got it! 31e5a262418409370f6b07b5909d8e4e4fed7213 fixes it.

The post meta update needed to be returned so it could nest promises and only finalize when this request is complete.

leogermani (approved):

PR #1608, opened 2024-08-12
fix(subscription-block): remove default list when misconfigured

miguelpeixe:
### All Submissions:

1200530782742699-as-1207876225097086

Changes how we handle the subscription form block when a subscription list becomes unavailable. The current behavior defaults to the first list available when none of the configured lists are found, which is not ideal.

If none of the lists are found, the form should not render and the editor should display a notice so it can be revised and fixed:

image

If the form has multiple lists selected and at least 1 one exists, the render is preserved with the lists available. While editing the page, the block should still display a notice to alert the editor that the block is not reflecting all the intended lists.

When adding a Newsletter Subscription Form block for the first time, the default behavior should still be auto-selecting the first available list for convenience.

1. Check out this branch and make sure you have an ESP connected with at least 3 subscription lists available
2. Edit a page and add the Newsletter Subscription Form block
3. Confirm the first available list is selected by default
4. Deactivate the subscription list in the Newsletters' settings page (make sure you have others enabled as well)
5. Visit the page and confirm the form no longer renders
6. Edit the page and confirm you see the notice as in the image above
7. Change the selection and confirm the notice disappears
8. Save the page, visit, and confirm the form renders
9. Edit the page again and select multiple lists
10. Deactivate one of them in the Newsletters' settings
11. Visit the page and confirm the deactivated list is not listed but others are
12. Edit the page and confirm the selection is preserved and you also get the same notice as above

### Other information:

---
  - https://app.asana.com/0/0/1207876225097086

### DIFF
--- a/src/blocks/subscribe/edit.js
+++ b/src/blocks/subscribe/edit.js
@@ -4,7 +4,6 @@
  * External dependencies.
  */
 import classnames from 'classnames';
-import { intersection } from 'lodash';
 
 /**
  * WordPress dependencies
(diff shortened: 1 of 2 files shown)


leogermani (approved):
Works as described

PR #1614, opened 2024-08-14
fix: add preview text fallback to make sure HTML is removed

laurelfulford:
### All Submissions:

This PR adds a fallback for the newsletter preview text that uses part of the body text with the HTML stripped out.

This is to work around an odd issue with Gmail: if you have an image with a link near the beginning of your email, the preview will include the image's alt text or description.

This happens when you draft regular emails from one Gmail account to another -- it's not unique to Newsletters or our plugin. 

See: 1207860156851023-as-1206907385452695

1. On `trunk`, draft a test newsletter with a linked image at the very top (this can be either an Image Block or Site Logo block). Make sure that the image has either alt text or a description entered in the Media Library.
2. Send a test newsletter (or the actual newsletter) to a Gmail email address.
3. Note the alt text appears in the preview:

CleanShot 2024-08-13 at 20 48 36

4. Apply this PR, and repeat steps 1 and 2.
5. Confirm that the preview no longer includes the alt text.
6. Draft a third newsletter, and add text to the Preview field in the right sidebar. 
7. Send that as a test (or real) newsletter, and confirm your custom text is still used in the preview. Note: if your text is short, the rest of the space will still be filled with the body in Gmail, so it may still include alt text/descriptions, but (thankfully!) not as the very first couple words.

### Other information:

### DIFF
--- a/includes/class-newspack-newsletters-renderer.php
+++ b/includes/class-newspack-newsletters-renderer.php
@@ -1266,6 +1266,9 @@ final class Newspack_Newsletters_Renderer {
 		$background_color = get_post_meta( $post->ID, 'background_color', true );
 		$preview_text     = get_post_meta( $post->ID, 'preview_text', true );
 		$custom_css       = get_post_meta( $post->ID, 'custom_css', true );
+		if ( ! $preview_text ) {
+			$preview_text = wp_trim_words( wp_strip_all_tags( $body ), 60 );
+		}
 		if ( ! $background_color ) {
 			$background_color = '#ffffff';
 		}


leogermani (approved):
Nice!

If the image has a caption, the caption still shows up in the preview, but I guess that's expected!

laurelfulford:
Thanks @leogermani! 

> If the image has a caption, the caption still shows up in the preview, but I guess that's expected!

Ah good catch! That is expected in this case, but could be worth digging into down the road, since the markup (I think) would be pretty consistent!

PR #1629, opened 2024-08-29
feat: send list classes

dkoo:
### All Submissions:

Implements a new `Send_List` class for abstacting ESP-specific list and sublist data into a single schema. Also implements a `Send_Lists` class for shared util methods and REST API routes (the latter currently won't work because this PR doesn't include the `get_send_lists()` methods on each provider class).

To support the new REST API routes, moves the `newsletters/v1` namespace and the API permissions callback to the main `Newspack_Newsletters` class, so these can be shared outside of the `Newspack_Newsletters_Subscription` class.

1. Confirm that all tests pass (and review the tests for expected `Send_List` behavior)
2. In `wp_shell`, try the new constructor:

```bash
> $send_list = new Newspack\Newsletters\Send_List( [ 'id' => '123', 'name' => 'My Send List', 'type' => 'list', 'entity_type' => 'something', 'count' => 100, 'provider' => 'mailchimp' ] )
= Newspack\Newsletters\Send_List {#14997
    +"provider": "mailchimp",
    +"type": "list",
    +"entity_type": "something",
    +"id": "123",
    +"name": "My Send List",
    +"count": 100,
    +"value": "123",
    +"label": "[SOMETHING] My Send List (100 contacts)",
  }
```

3. Try out the new methods:

```bash
> $send_list->get_id()
= "123"

> $send_list->get_label()
= "[SOMETHING] My Send List (100 contacts)"

> $send_list->set_name( 'My Send List (renamed)' )
= "My Send List (renamed)"

> $send_list->get_name()
= "My Send List (renamed)"

> Newspack\Newsletters\Send_Lists::matches_search( 'renamed', [ $send_list->get( 'name' ) ] )
= true

> Newspack\Newsletters\Send_Lists::matches_id( [ 123 ], $send_list->get( 'id' ) )
= true

> $send_list->to_array()
= [
    "provider" => "mailchimp",
    "type" => "list",
    "entity_type" => "something",
    "id" => "123",
    "name" => "My Send List",
    "count" => 100,
    "label" => "[SOMETHING] My Send List (100 contacts)",
    "value" => "123",
  ]
```

### Other information:

---
  - https://app.asana.com/0/0/1207992056607959

### DIFF
--- a/includes/class-send-list.php
+++ b/includes/class-send-list.php
@@ -27,13 +27,6 @@ class Send_List {
 	protected $config;
 
 	/**
-	 * If there were errors building the Send_List object due to invalid config, this will contain the error messages.
-	 *
-	 * @var WP_Error
-	 */
-	protected $errors;
-
-	/**
 	 * Initializes a new Send_List.
 	 *
 	 * @param array $config {
@@ -43,23 +36,25 @@ class Send_List {
 	 *   @type string $type        The type of list. Can be either 'list' or 'sublist'. If the latter, must specify a `parent` property (required).
 	 *   @type string $entity_type The type of entity this list or sublist is associated with in the ESP. Controls which ESP API endpoints to use to fetch/update (required).
 	 *   @type string $id          The ID of the list or sublist in the ESP (required).
-	 *   @type string $value       The value to use for autocomplete inputs. If not passed, will match `id`.
 	 *   @type string $parent_id   If this is a sublist of another list, the ID of the parent list.
 	 *   @type string $name        The name of the list or sublist as identified in the ESP (required).
 	 *   @type string $local_name  If the list is also a Subscription List, it could have a locally edited name.
-	 *   @type string $label       The label to use for autocomplete inputs. If not set, will be generated from other properties.
 	 *   @type int    $count       If available, the number of contacts associated with this list or sublist.
 	 *   @type string $edit_link   If available, the URL to edit this list or sublist in the ESP.
+	 *   @type string $label       The label to use for autocomplete inputs. If not set, will be generated from other properties.
+	 *   @type string $value       The value to use for autocomplete inputs. If not passed, will match `id`.
 	 * }
+	 *
+	 * @throws \InvalidArgumentException In case the Send_List object can't be built from the given config data.
 	 */
 	public function __construct( $config ) {
-		$this->errors = new WP_Error();
-		$schema       = self::get_config_schema();
+		$errors = new WP_Error();
+		$schema = self::get_config_schema();
 
 		foreach ( $schema['properties'] as $key => $property ) {
 			// If the property is required but not set, throw an error.
 			if ( ! empty( $property['required'] ) && ! isset( $config[ $key ] ) ) {
-				$this->errors->add( 'newspack_newsletters_send_list_invalid_config', __( 'Missing required config property: ', 'newspack-newsletters' ) . $key );
+				$errors->add( 'newspack_newsletters_send_list_invalid_config', __( 'Missing required property: ', 'newspack-newsletters' ) . $key );
 				continue;
 			}
 
@@ -69,7 +64,15 @@ class Send_List {
 			}
 
 			// Set the property.
-			$this->set( $key, $config[ $key ] );
+			$result = $this->set( $key, $config[ $key ] );
+			if ( \is_wp_error( $result ) ) {
+				$result->export_to( $errors );
+			}
+		}
+
+		// Throw an exception if there are errors.
+		if ( $errors->has_errors() ) {
+			throw new \InvalidArgumentException( esc_html( __( 'Error creating send list: ', 'newspack-newsletters' ) . implode( '. ', $errors->get_error_messages() ) ) );
 		}
 	}
 
@@ -157,14 +160,10 @@ class Send_List {
 	 * @return mixed|WP_Error The property value or null if not set/not a supported property.
 	 */
 	private function get( $key ) {
-		// Clear previous errors for this key.
-		$this->errors->remove( 'newspack_newsletters_send_list_missing_required_property_' . $key );
 		$value  = $this->{ $key } ?? null;
 		$schema = $this->get_config_schema();
 		if ( ! empty( $schema['properties'][ $key ]['required'] ) && null === $this->{ $key } ) {
-			$error_message = new WP_Error( 'newspack_newsletters_send_list_missing_required_property_' . $key, __( 'Could not get required property: ', 'newspack-newsletters' ) . $key );
-			$error_message->export_to( $this->get_errors() );
-			return $error_message;
+			return new WP_Error( 'newspack_newsletters_send_list_missing_required_property_' . $key, __( 'Could not get required property: ', 'newspack-newsletters' ) . $key );
 		}
 
 		return $value;
@@ -180,22 +179,15 @@ class Send_List {
 	 * @return mixed|WP_Error The property value or WP_Error if not set/not a supported property.
 	 */
 	private function set( $key, $value ) {
-		// Clear previous errors for this key.
-		$this->errors->remove( 'newspack_newsletters_send_list_invalid_property_' . $key );
 		$schema = $this->get_config_schema();
 		if ( ! isset( $schema['properties'][ $key ] ) ) {
-			$error_message = new WP_Error( 'newspack_newsletters_send_list_invalid_property_' . $key, __( 'Could not set invalid property: ', 'newspack-newsletters' ) . $key );
-			$error_message->export_to( $this->get_errors() );
-			return $error_message;
+			return new WP_Error( 'newspack_newsletters_send_list_invalid_property_' . $key, __( 'Could not set invalid property: ', 'newspack-newsletters' ) . $key );
 		}
 
 		// If the passed value isn't in the enum, throw an error.
-		$this->errors->remove( 'newspack_newsletters_send_list_invalid_property_value_' . $key );
 		$property = $schema['properties'][ $key ];
 		if ( isset( $property['enum'] ) && ! in_array( $value, $property['enum'], true ) ) {
-			$error_message = new WP_Error( 'newspack_newsletters_send_list_invalid_property_value_' . $key, __( 'Invalid value for property: ', 'newspack-newsletters' ) . $key );
-			$error_message->export_to( $this->get_errors() );
-			return $error_message;
+			return new WP_Error( 'newspack_newsletters_send_list_invalid_property_value_' . $key, __( 'Invalid value for property: ', 'newspack-newsletters' ) . $key );
 		}
 
 		// Cast value to the expected type.
@@ -203,15 +195,6 @@ class Send_List {
 
 		$this->{ $key } = $value;
 		return $this->get( $key );
-	}
-
-	/**
-	 * Get any errors associated with this Send_List.
-	 *
-	 * @return WP_Error
-	 */
-	public function get_errors() {
-		return $this->errors;
 	}
 
 	/**
@@ -299,7 +282,7 @@ class Send_List {
 
 	/**
 	 * Get a manually set or dynamic label for autocomplete inputs.
-	 * If not manually set, generate from entity_type, name, and count and store.
+	 * If not passed via __construct(), will be generated from entity_type, name, and count.
 	 *
 	 * @return string
 	 */
@@ -324,7 +307,8 @@ class Send_List {
 	}
 
 	/**
-	 * Get the value for autocomplete inputs. Defaults to the id.
+	 * Get the value for autocomplete inputs.
+	 * If not passed via __construct(), defaults to the id.
 	 *
 	 * @return string
 	 */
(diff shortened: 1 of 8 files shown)


includes/class-send-list.php
```diff
@@ -0,0 +1,213 @@
...
+	 * The configuration associated with this Send_List.
+	 *
+	 * @var array
+	 */
+	protected $config;
+
+	/**
+	 * Initializes a new Send_List.
+	 *
+	 * @param array $config The configuration for the Send_List.
```
miguelpeixe: It'd be nice to have the accepted values for this assoc array in the docblock. E.g.:

```
	 * @param array $config {
	 *   The configuration for the Send_list.
	 *
	 *   @type string $provider The slug of the ESP for which this list or sublist originates.
	 *   @type string $type     The type of list. Can be either 'list' or 'sublist'. If the latter, must specify a `parent` property.
```

includes/class-send-list.php
```diff
@@ -0,0 +1,213 @@
...
+	/**
+	 * Get a specific property's value.
+	 *
+	 * @param string $key The property to get.
+	 *
+	 * @return mixed The property value or null if not set/not a supported property.
+	 */
+	public function get( $key ) {
+		return $this->{ $key } ?? null;
+	}
```
miguelpeixe: Following the steps from the `Subscription_List`, I wonder if it's worth being more explicit using named methods like `get_provider()` rather than `get( 'provider' )`.

I understand it requires more boilerplate but the encapsulation might be better in the long run.

Maybe @leogermani could share his thoughts?

includes/class-send-list.php
```diff
@@ -0,0 +1,213 @@
...
+
+	/**
+	 * Initializes a new Send_List.
+	 *
+	 * @param array $config The configuration for the Send_List.
+	 * @throws \InvalidArgumentException In case the Send_List object can't be built from the given config data.
+	 */
+	public function __construct( $config ) {
+		$schema = self::get_config_schema();
+		$errors = [];
```
miguelpeixe: NIT suggestion: the `WP_Error` class provides a nice API with `$errors->add( $code, $message )` and `$errors->has_errors()`. I find it better than handling this with try/catch.

includes/class-send-list.php
```diff
@@ -0,0 +1,213 @@
...
+	 *
+	 * @param string $key The property to get.
+	 * @param mixed  $value The value to set.
+	 *
+	 * @return mixed The property value or null if not set/not a supported property.
+	 */
+	public function set( $key, $value ) {
+		$schema = $this->get_config_schema();
+		if ( ! isset( $schema['properties'][ $key ] ) ) {
+			return null;
```
miguelpeixe: We might want to return an error here.

includes/class-send-list.php
```diff
@@ -0,0 +1,213 @@
...
+		}
+		$this->{ $key } = $value;
+		return $this->get( $key );
+	}
+
+	/**
+	 * Set the label and value properties for autocomplete inputs.
+	 */
+	public function set_label_and_value() {
```
miguelpeixe: Given this is always dynamic, based on the Send_List config, is there a specific reason we're setting and storing? A `get_label_and_value()` would work well and not require every config change to call this method for an update.

miguelpeixe (changes requested):
The class looks great and I love how clear the tests are!

I have a few comments:

includes/class-send-list.php
dkoo: I stored them as properties mainly because these are for the `Autocomplete` component in the editor JS, and having them automatically generated and set by the constructor means that I never have to make sure they're set before using them there. But you're right that if we ever update any of the properties that are used to generate these, we would need to regenerate the dynamic values. Let me think about this a bit more.

includes/class-send-list.php
dkoo: Great idea! Moved this info to the docblock in 7f4906e.

includes/class-send-list.php
dkoo: I agree. 7f4906e adds the following methods:

- `get_errors`
- `get_provider`
- `get_id`
- `get_type`
- `get_entity_type`
- `get_name`
- `get_local_name` (this isn't used anywhere currently, but I included local name so we can have the option to inherit the local name from the `Subscription_List` created from the same entity)
- `get_label`
- `get_value`
- `set_name`
- `set_local_name`
- `get_parent_id` (3469cf4)
- `get_count` (3469cf4)
- `set_count` (f103484)

This makes it clearer which properties you might need to get and which can be set after the object has been constructed. e.g. you can't reset `id` since it's a required property. I kept the `get` and `set` methods around as private helpers which also helps with standardizing the error handling.

includes/class-send-list.php
dkoo: Good idea. 7f4906e refactors the error handling to use a `WP_Error` object for this instead of an array of strings. The constructor will no longer throw an exception, but will add error data that can be fetched all at once using the new `get_errors()` method. `get_*` and `set_*` methods will also return any applicable errors in addition to adding their own messages to the class instance's errors object.

dkoo: #1631 will need to be refactored to take advantage of these improvements, but I'll hold off on that until this PR is merged.

includes/class-send-list.php
dkoo: Done in 7f4906e!

includes/class-send-list.php
dkoo: 7f4906e introduces `get_label()` and `get_value()` methods which will either retrieve the stored values (if passed in the constructor config) or dynamically generate them. 7f4906e also takes inspiration from @leogermani's `Subscription_List` object with a `to_array()` method which will return the send list's data as an associative array. This gives us a chance to check whether we need to generate the dynamic values and append them to the response, and it also will let us tailor the data we pass back as REST responses (to be handled in #1631).

includes/class-send-list.php
```diff
@@ -0,0 +1,358 @@
...
+	 *
+	 * @return mixed|WP_Error The property value or WP_Error if not set/not a supported property.
+	 */
+	private function set( $key, $value ) {
+		// Clear previous errors for this key.
+		$this->errors->remove( 'newspack_newsletters_send_list_invalid_property_' . $key );
+		$schema = $this->get_config_schema();
+		if ( ! isset( $schema['properties'][ $key ] ) ) {
+			$error_message = new WP_Error( 'newspack_newsletters_send_list_invalid_property_' . $key, __( 'Could not set invalid property: ', 'newspack-newsletters' ) . $key );
+			$error_message->export_to( $this->get_errors() );
```
miguelpeixe: This leaves the impression that a get/set failing has the entire Send_List object in a broken state, which I don't think should be interpreted that way. Non-blocking, but I think it's fine to just return the error to the caller without modifying the object error.

includes/class-send-list.php
```diff
@@ -0,0 +1,358 @@
...
+		}
+		$config = array_filter( $config ); // Remove empty values.
+
+		// Ensure label + value properties are set for JS components.
+		if ( ! isset( $config['label'] ) ) {
+			$config['label'] = $this->get_label();
+		}
+		if ( ! isset( $config['value'] ) ) {
+			$config['value'] = $this->get_value();
+		}
```
miguelpeixe: Won't these conditionals defeat the purpose of updating these values with the latest config data?

includes/class-send-list.php
```diff
@@ -0,0 +1,358 @@
...
+	 * Get a manually set or dynamic label for autocomplete inputs.
+	 * If not manually set, generate from entity_type, name, and count and store.
+	 *
+	 * @return string
+	 */
+	public function get_label() {
+		$stored_label = $this->get( 'label' );
+		if ( ! empty( $stored_label ) ) {
+			return $stored_label;
+		}
```
miguelpeixe, line 299: Same as in `to_array()`, why bail if the value exists if we want it up to date with dynamic data?

includes/class-send-list.php
miguelpeixe: I failed to recognize that this is the constructor 🤦 

I think we can still leverage the WP_Error class to manage multiple errors but it's best practice to maintain the throw here... Tough decision because I really like how you did it 😅

includes/class-send-list.php
dkoo: My thought with allowing the `label` and `value` properties to be set in the constructor was that we might want to manually set those values to override the dynamically generated ones. But I'm not actually doing that anywhere in the code, so maybe that's overkill?

The way it works now, if you don't create the `Send_List` with `label` or `value` props on the config object, then they'll always be dynamically generated when you call `to_array()`. Note how `get_label()` and `get_value()` don't set the generated values, they just return them.

includes/class-send-list.php
dkoo: Ah, no problem! 7a3d8b3 restores the throw but keeps the use of `WP_Error` to manage the errors.

includes/class-send-list.php
dkoo: If this isn't needed then we can remove `label` and `value` from the schema and remove all the conditionals. Otherwise we can keep as-is to retain the future option for manual overrides.

includes/class-send-list.php
dkoo: 7a3d8b3 simplifies this by just returning the error.

includes/class-send-list.php
miguelpeixe: Aaaah I see now! It makes sense to me, I think it's a nice aspect, not overkill.

dkoo: Awesome, I wasn't sure myself! Let's keep it, then 😄

miguelpeixe (approved):
Thank you for the revisions! Class and tests look great!

Let's get this in to review it in action 😎

includes/class-send-list.php
```diff
@@ -0,0 +1,338 @@
...
+		// If the passed value isn't in the enum, throw an error.
+		$property = $schema['properties'][ $key ];
+		if ( isset( $property['enum'] ) && ! in_array( $value, $property['enum'], true ) ) {
+			return new WP_Error( 'newspack_newsletters_send_list_invalid_property_value_' . $key, __( 'Invalid value for property: ', 'newspack-newsletters' ) . $key );
+		}
+
+		// Cast value to the expected type.
+		settype( $value, $property['type'] );
+
+		$this->{ $key } = $value;
```
leogermani: Setting the property directly into an undeclared object property has the downside of letting this prop public.

So I can still access `$send_list->label` which would skip our getter.

We could come up with some ideas to work around that, maybe using [overloading](https://www.php.net/manual/en/language.oop5.overloading.php) somehow, but I think the easiest thing to do is to store all properties inside one `private $config` property, which will be an array holding all the properties...

makes sense?

miguelpeixe:
@leogermani reviewed the same time as me 😄

tests/test-send-lists.php
```diff
@@ -0,0 +1,37 @@
...
+		$this->assertTrue( Send_Lists::matches_id( '123', '123' ) ); // Single ID matches.
+		$this->assertTrue( Send_Lists::matches_id( [ '123', '456' ], '123' ) ); // Array of IDs matches.
+		$this->assertFalse( Send_Lists::matches_id( '456', '123' ) ); // Single of ID doesn't match.
+		$this->assertFalse( Send_Lists::matches_id( [ '456', '789' ], '123' ) ); // Array of IDs doesn't match.
+	}
+
+	/**
+	 * Test matching by search term(s).
+	 */
+	public function test_match_by_search() {
```
leogermani, line 33: Nice!

When testing methods like this it is always desirable to test some reallly unhappy tests.

For example, what's the expected return if I pass NULL as one of the parameters? or an Object of any kind? or an integer, etc.

includes/class-send-list.php
```diff
@@ -0,0 +1,338 @@
...
+
+namespace Newspack\Newsletters;
+
+use Newspack_Newsletters;
+use WP_Error;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Class used to represent one Send_List.
```
leogermani, line 16: Maybe too verbose.. but just trying to add some rationale for our future selves... 

```suggestion
 * Class used to represent one Send_List.
 * 
 * A Send List is any list that can be used as a segment of contacts that you can send a Newsletter campaign to
 * The difference between Send_List and Subscription_List is that Send Lists are all the existing lists, audiences, tags, groups, segments, etc in the provider. We don't store them locally. Some of these can also be Subscription Lists, but not all are.
```

leogermani (changes requested):
This is looking pretty good Derrick! Thanks for all the work building it!

I left only one comment regarding how we store the properties in the object. Other than that this is awesome!

includes/class-send-list.php
dkoo: Ah, I didn't realize that! 69e8232 refactors to use a single protected `$config` property instead of the separate undeclared properties. This also makes the `to_array()` method a bit simpler.

tests/test-send-lists.php
dkoo: Great idea! f9bd959 adds some of these tests (as well as some missing handling for those weird cases!).

leogermani: That's the beauty of writing tests :)

leogermani (approved):

leogermani (approved):

PR #1530, opened 2024-06-14
fix: refactor for updating dependencies in newspack-scripts

dkoo:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/trunk/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Refactoring required for changes in https://github.com/Automattic/newspack-scripts/pull/209. Note that this PR will not pass CI tests and should NOT be merged until https://github.com/Automattic/newspack-scripts/pull/209 is released as an NPM package, so this repo can install it as a dependency.

### How to test the changes in this Pull Request:

Follow instructions in https://github.com/Automattic/newspack-scripts/pull/209.

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

PR #1632, opened 2024-08-29
feat: refactor editor JS to use new standardized data

dkoo:
### All Submissions:

To be reviewed after #1631 (the base for this branch).

Implements editor JS changes to support standardizations in the data structure across ESPs in #1631.

- Deprecates the `newsletterData` post meta field in favor of storing `retrieve` output in a Redux store to be fetched on editor mounting, shared across editor components for the life of the session, and rehydrated when needed (upon fetching additional sublists or after saving/syncing, for example)
- Refactors `SendTo` components to use the new `Send_List` data schema for all ESPs
- Refactors Sender (From email/name) UI to use new standardized meta fields for all ESPs
- Eliminates most provider-specific code that's no longer needed due to the above abstraction layers (as sender and send-to data is now handled at the editor/sidebar level rather than in the provider-specific sidebars). Also moves validation checks to the higher level since these can now be shared across ESPs.
- Improves "Sender" UI so that ESP-specific quirks are handled on the front-end (e.g. Mailchimp requires verified domains, CC requires verfiied email addresses), but in an abstract way—no ESP-specific code required except for adding certain data to the retrieve payload
- Standardizes sync behavior across all ESPs and types of data: sender and send-to info are no longer synced on the fly, but only after clicking "save" along with all other newsletter campaign data. This also includes provider-specific data such as Mailchimp's campaign folders.
- Implements a true two-way sync for ESPs that maintain a synced draft campaign alongside the newsletter post. When fetching campaign data from the ESP, if sender or send-to info differs from the fetched campaign, we update the stored post meta to match.

1. On `epic/update-editor-uis`, create four new newsletters: one for each ESP. You'll need to connect an ESP, create the newsletter, edit the info, change the ESP, create the next newsletter, etc. In each newsletter you create, set a sender name + email, and select a list and sublist (only MC & AC support sublists). Make sure to name each newsletter so that it's easy to identify which ESP they're for.
2. Check out this branch and rebuild.
3. Edit the newsletter for the currently active ESP. Confirm that all sender and send-to info that you set in step 1 remain intact.
4. **(For AC + MC + CC only)** Visit the connected ESP account dashboard and find the synced campaign. Ensure that all info in the synced campaign matches the newsletter post. Update the sender and send-to info in the ESP dashboard and save, then refresh the newsletter post in WP and confirm that they sender and send-to info are updated in the editor with a message:

Screenshot 2024-08-21 at 2 01 37 PM

5. **(For MC only)** Confirm that when entering a sender email, you're shown a warning if the input doesn't contain one of the account's required/verified domains.

Screenshot 2024-09-09 at 9 44 41 PM

6. **(For CC only)** Confirm that the sender email is a dropdown UI that only allows you to select email addresses that have been added and confirmed in your connected CC account.

Screenshot 2024-08-26 at 5 38 48 PM

7. Smoke test making further edits to the sender and send-to info, saving the post, and sending a test campaign.
8. Send a live campaign and confirm that it gets sent to the correct recipients and with the correct sender info and content.
9. Repeat for all four ESPs.
10. Switch ESP to "manual/other" mode and smoke test creating, editing, and previewing a newsletter, and copying the email HTML from the "Send" modal.
11. Smoke test creating, editing, inserting, and previewing newsletter ads.
12. Smoke test editing non-newsletter CPT emails such as the custom Reader Revenue receipt and RAS emails.

### Other information:

---
  - https://app.asana.com/0/0/1207717004877614
  - https://app.asana.com/0/0/1207717004877591
  - https://app.asana.com/0/0/1207799991082606
  - https://app.asana.com/0/0/1207717004877578


src/editor/mjml/index.js
```diff
@@ -105,19 +111,28 @@ function MJML() {
...
-								isDismissible: true,
-							} );
-						} else {
-							removeNotice( 'newspack-newsletters-newsletter-sync-error' );
-						}
-					} );
+
+					const { name: serviceProviderName } = getServiceProvider();
+					const { supported_esps: suppportedESPs } = newspack_email_editor_data || {};
+					const isSupportedESP = serviceProviderName && 'manual' !== serviceProviderName && suppportedESPs?.includes( serviceProviderName );
```
miguelpeixe: `supportedESPs` is an object, which doesn't support `.includes()`.  This needs to be transformed into an array.

includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php
```diff
@@ -704,7 +704,13 @@ public function get_send_lists( $args = [] ) {
...
 		}
 
-		return $send_lists;
+		// Convert to arrays.
+		return array_map(
+			function ( $list ) {
+				return $list->to_array();
+			},
+			$send_lists
+		);
```
miguelpeixe: This also warrants an update in the docblock, which currently states it returns `Send_List[]`.

I'm not sure it's worth losing access to the send list object in this method. Is there another reason to use an array other than API responses? If not, we can just use `->to_array()` on API callbacks and enforce the use of the object across the class, which is safer.

src/service-providers/mailchimp/index.js
```diff
@@ -67,25 +56,37 @@ const renderPreSendInfo = newsletterData => {
...
-				</Fragment>
+				</>
 			) }
 			<strong>
 				{ sprintf(
 					// Translators: subscriber count help message.
-					_n( '%d subscriber', '%d subscribers', listData.subscribers, 'newspack-newsletters' ),
-					listData.subscribers
+					_n( '%d subscriber', '%d subscribers', subscriberCount, 'newspack-newsletters' ),
+					subscriberCount
```
miguelpeixe: This is not considering the selected group/segment/tag and I'm getting the total subscribers from the audience on the pre-send modal:

image

image

src/service-providers/mailchimp/index.js
```diff
@@ -67,25 +56,37 @@ const renderPreSendInfo = newsletterData => {
...
 				) }
 			</strong>
 		</p>
 	);
 };
 
+const isCampaignSent= ( newsletterData, postStatus = 'draft' ) => {
```
miguelpeixe:
```suggestion
const isCampaignSent = ( newsletterData, postStatus = 'draft' ) => {
```

src/service-providers/example/index.js
```diff
@@ -1,83 +1,115 @@
...
+ * @param {Object} newsletterData The data returned from the ESP retrieve method
+ * @param {string} postStatus     The post's current status.
+ * @return {boolean} True if the campaign has been sent, otherwise false.
  */
-const renderPreSendInfo = ( newsletterData = {} ) => (
-	<p>
-		{ __( 'Sending newsletter to:', 'newspack-newsletters' ) } { newsletterData.listName }
-	</p>
-);
+const isCampaignSent= ( newsletterData, postStatus = 'draft' ) => {
```
miguelpeixe:
```suggestion
const isCampaignSent = ( newsletterData, postStatus = 'draft' ) => {
```

src/service-providers/constant_contact/index.js
```diff
@@ -4,67 +4,72 @@
...
+						_n( '%d subscriber', '%d subscribers', subscriberCount, 'newspack-newsletters' ),
+						subscriberCount
+					) }
+				</strong>
+			) }
 		</p>
 	);
 };
 
+const isCampaignSent= ( newsletterData, postStatus = 'draft' ) => {
```
miguelpeixe:
```suggestion
const isCampaignSent = ( newsletterData, postStatus = 'draft' ) => {
```

src/service-providers/active_campaign/index.js
```diff
@@ -46,12 +44,18 @@ const renderPreSendInfo = ( newsletterData = {} ) => {
 				) }
 			</p>
 			<p>{ __( 'Are you sure you want to proceed?', 'newspack-newsletters' ) }</p>
-		</Fragment>
+		</>
 	);
 };
 
+const isCampaignSent= ( newsletterData, postStatus = 'draft' ) => {
```
miguelpeixe:
```suggestion
const isCampaignSent = ( newsletterData, postStatus = 'draft' ) => {
```

Is our linter broken?

miguelpeixe (changes requested):
Great use of the redux store to get rid of `newsletterData`. It's a great approach and I love seeing so much duplicated code going away with the new Send Lists being used.

Tests went well with Mailchimp, but it's a large PR and I'll continue my review after my initial comments are addressed.

includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php
dkoo: 7a48870 refactors things so that the `get_send_lists()` methods now take a second optional arg, `$to_array`. If passed as `true`, the `Send_List` objects will be converted to arrays before returning, otherwise they'll be returned as regular objects. This lets us get the arrays for REST API and `localize_script` contexts and regular `Send_List` objects for PHP-only contexts.

src/editor/mjml/index.js
dkoo: Whoops, it's actually an array, not an object, I was just setting the wrong fallback value. be52969 should fix this.

src/service-providers/mailchimp/index.js
dkoo: Ah, good catch—this is because the pre-send modal for MC only was using the `campaign.recipients.recipient_count` value from `newsletterData`, but this isn't updated until after a sync is completed AND we rehydrate the Redux store with another `retrieve` request. To avoid the miscount and also not add another roundtrip request delay before showing the modal, 3ffcd6b refactors this to use the `Send_List` and `Send_Sublist` counts, just like in the sidebar summary. There's a small chance that this count could be slightly off if the count changes in the ESP after the most recent `retrieve`, but the difference would likely be negligible, and since clicking the "Send" button does trigger a post save + campaign sync + MJML re-render + `retrieve` request + store rehydration, any incorrect value would likely update itself after a few seconds.

src/service-providers/mailchimp/index.js
dkoo: Whoops, fixed in a22b721

src/service-providers/example/index.js
dkoo: Fixed in a22b721

src/service-providers/constant_contact/index.js
dkoo: a22b721

src/service-providers/active_campaign/index.js
dkoo: Our linter isn't broken, but we decided to temporarily disable the linter from reporting Prettier issues because of the switch to `@wordpress/scripts` (which has different Prettier rules than what we were using before) in #1530.

src/service-providers/constant_contact/index.js
dkoo: Fixed in a22b721

src/editor/mjml/index.js
miguelpeixe: In my case the newspack_email_editor_data.supported_esps is also coming as an object:

```
{
  '0': 'mailchimp',
  '1': 'constant_contact',
  '3': 'active_campaign',
  '4': 'manual'
}
```

That's because of this:
https://github.com/Automattic/newspack-newsletters/blob/a22b7219cbaaaea6e69175bfa42307e4a867cb7e/includes/class-newspack-newsletters.php#L123-L126

`array_diff()` won't reset the indexes and json encoding an indexed array with a skipped index will transform it into an object.

An `array_values()` there should suffice.

Also, there's an extra P in `suppportedESPs`.

dkoo: Ah, I see what you mean! I wasn't getting the object because I had the `NEWSPACK_NEWSLETTERS_SUPPORT_DEPRECATED_CAMPAIGN_MONITOR` constant enabled, so the array wasn't getting passed through `array_diff`. c00aca5 should fix both issues. :ppp

miguelpeixe:
You forgot to update [this bit](https://github.com/Automattic/newspack-newsletters/blob/c00aca535f4b0b1d3f48947f845149c493918124/includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign-controller.php#L122) and it's throwing an error when trying to send a test campaign via AC:

```
PHP Fatal error:  Uncaught TypeError: call_user_func(): Argument #1 ($callback) must be a valid callback, class Newspack_Newsletters_Active_Campaign does not have a method "api_authoring_permissions_check"
```

dkoo:
a43989c

miguelpeixe:
While testing Constant Contact, I ran into issues with account permissions that uncovered some error-handling problems. I pushed 6f85a8b7d2a91f03410a1ee17d4a87c62ee90780, which introduces some error handling to `retrieve()`, `sync()`, and  `get_sync_payload()`.

dkoo:
@miguelpeixe thanks for highlighting missing or inconsistent error handling. I've attempted to plug more holes and unify things a bit better across ESPs in these PRs:

f5555c2 - Implements missing error handling in the Redux store. Conceptually the store's error state can describe an error state from any attempt to get or set data via the 3rd-party ESP's API. Any editor component can use the `useNewsletterDataError` hook to check for the error state and act accordingly. So far the store's error state is used in the following contexts:

  - The `fetchNewsletterData` helper, if the `retrieve` fetch fails
  - The `sync-error` fetch to get any errors resulting from the last sync
  - The `NewsletterEditor` component will display the store's error state as a top-level editor error notice (which essentially acts like an alternative way to show generic errors—compare with the existing `apiFetchWithErrorHandling` HOC)
  - In the editor `Sidebar` component, which will now show an error if the `retrieve` request fails instead of the `Retrieving campaign data...` state forever, as well as a button which will let you trigger another `retrieve` request without reloading the whole editor. (We might want to consider offering a refresh button even if there are no errors—maybe along with the link to the synced campaigns in another PR)

Screenshot 2024-09-12 at 5 30 33 PM

5627680 - Now that f5555c2 allows editors to manually trigger another `retrieve` request in the event of a failure to fetch or create campaign data, 5627680 removes recursive calls of `retrieve()` in the event of failing to fetch a synced campaign by ID. Instead, for the ESPs that maintain a synced campaign in the ESP side-by-side with the newsletter post (Mailchimp and Constant Contact), we now have a `fetch_synced_campaign()` helper method that will fetch the campaign via ID from the ESP's API and return either the campaign data or a `WP_Error` if it can't find the campaign. This lets the `retrieve()` method delete the campaign ID meta value only in the event of a failure to fetch a campaign by ID, in which case the editor can click the "Retrieve campaign data" button to trigger another attempt which will recreate the campaign.

5627680 also unifies the error handling in `retrieve()` methods across all ESPs. Before there was a mix of thrown exceptions and `WP_Error`s. Now all `retrieve()` methods will handle `WP_Error`s as exceptions in the same way, in all scenarios.

You can test the above scenarios by adding `throw new Exception( 'Test error message' );` in various places throughout each ESP's `retrieve()`, `sync()`, `get_send_lists()`, etc. methods as well as hardcoding an `$error_message` in the `api_get_sync_error()` REST API handler.

miguelpeixe (approved):
Awesome! Thank you for the revisions and new error handling strategy!

Tested on all ESPs and it's working well.

PR #1633, opened 2024-08-29
fix: default sender/send-to for saved layouts

dkoo:
### All Submissions:

To be reviewed after #1632 (the base for this PR)

Fixes auto-population of saved sender and send-to info when creating a new newsletter from a custom layout containing that info.

Note that this is a **BREAKING CHANGE** as layouts that were saved prior to the introduction of the new sender and send-to meta fields will no longer auto-populate this info from the old `newsletterData` meta fields. However, it should be possible to set sender and send-to info and update the existing layouts to restore those defaults for new newsletters going forward.

1. Create a new newsletter and save as a new custom layout for each ESP. In each layout, set a sender name/email and send list and sublist (for those ESPs that support sublists).
2. Create a new newsletter using the custom layouts created in step 1. Confirm that the sender and send-to info are prepopulated from the saved layouts. 
3. Save and send this newsletter, and confirm that the campaign sends correctly with the right sender and send-to info.

### Other information:

### DIFF
--- a/includes/class-newspack-newsletters-layouts.php
+++ b/includes/class-newspack-newsletters-layouts.php
@@ -77,7 +77,7 @@ final class Newspack_Newsletters_Layouts {
 		\register_meta( 'post', 'font_body', $meta_default_params );
 		\register_meta( 'post', 'background_color', $meta_default_params );
 		\register_meta( 'post', 'custom_css', $meta_default_params );
-		\register_meta( 'post', 'layout_defaults', $meta_default_params );
+		\register_meta( 'post', 'campaign_defaults', $meta_default_params );
 	}
 
 	/**
--- a/includes/class-newspack-newsletters.php
+++ b/includes/class-newspack-newsletters.php
@@ -182,6 +182,10 @@ final class Newspack_Newsletters {
 		];
 		$fields = [
 			[
+				'name'               => 'stringifiedCampaignDefaults',
+				'register_meta_args' => $default_register_meta_args,
+			],
+			[
 				'name'               => 'newsletter_send_errors',
 				'register_meta_args' => [
 					'show_in_rest' => [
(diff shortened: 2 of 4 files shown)


miguelpeixe (changes requested):
I'm not able to reproduce the testing steps.

I'm drafting a new newsletter, saving it, and creating a new layout from it. After creating a new newsletter based on the layout, the sender info and "send to" options are empty.

Also behaving strangely, the saved layout is not persisted to the original draft. Once I first create the layout I get the proper layout info:

image

After refreshing the page, it's gone:

image

Mind that this is not the case for the derived newsletter, the one created from the saved layout has its selected layout persisted.

dkoo:
@miguelpeixe looks like this branch had out-of-date code. Let me re-test after updating with the latest from the epic and see if this addresses the issues you saw.

dkoo:
> I'm drafting a new newsletter, saving it, and creating a new layout from it. After creating a new newsletter based on the layout, the sender info and "send to" options are empty.

> Also behaving strangely, the saved layout is not persisted to the original draft.

@miguelpeixe a0aaa98 should fix these issues. Looks like the `campaign_defaults` meta field name got accidentally reverted to the legacy `layout_defaults` by a recent merge from `trunk`. And the saved layout is now persisted to the original draft as soon as you save the layout, by triggering a `savePost()` request after the layout is created (it wasn't persisting before unless you clicked "Save draft" because it would only update the edited `template_id` meta value without persisting it to the server).

dkoo:
There's still a bug here where if you create a new newsletter using a custom layout specifying a sublist, or set the layout of a newsletter to that custom layout, the sublist will be shown in a deselected state until you focus the sublist Autocomplete field. #1657 should fix this, so I'll try to rebase this PR to that one to confirm.

EDIT: Yep, looks like #1657 fixes this issue. Reopening for review.

miguelpeixe (approved):
Nice! Working great now.

PR #1635, opened 2024-08-29
fix: handle encoded URLs and block patterns in redirect check

dkoo:
### All Submissions:

Fixes a bug in the click handler for URLs within newsletter content. Fixes the following scenarios:

1. When inserting a link in a synced block pattern and then inserting the synced block into a post, the contents of the synced block is not rendered in the raw post content, only a block ref like `<!-- wp:block {"ref":604} /-->`. This PR checks if the link is contained in the final rendered HTML instead, which is stored in the `newspack_email_html` meta field.
2. If you're copying and pasting links from another email, or using a block pattern, the link ends up being saved with the full redirect wrapper, e.g. `https://url?np_newsletters_click=1&id=546&url=https%3A%2F%2Fnewspack.com%2Fabout%2F%3Futm_medium%3Demail&em=*|EMAIL|*` instead of simply `https://newspack.com/about/`. This PR checks the newsletter content for both unencoded and encoded forms of the URL (`https%3A%2F%2Fnewspack.com%2Fabout`) to account for this scenario.
3. While authoring a newsletter, if you preview the newsletter without first saving, any links in unsaved content won't be present in the saved newsletter content for the check, and will return the `Invalid URL` error. This PR accounts for this scenario by letting the redirect happen if `current_user_can( 'edit_others_posts' )` passes.
4. Not an error per se, but currently both `seen` and `click` events are tracked for logged-in editor/admin users, which might be inflating the numbers. This PR does not track either if `current_user_can( 'edit_others_posts' )` passes.

1. In `release`, while editing a newsletter post, create a synced block pattern containing a link to any URL and save it.
2. Insert the synced block pattern into another newsletter post.
3. Send a test to yourself and click the link inside the synced block pattern. Observe that the redirect fails with an `Invalid URL` error.
4. Also preview the newsletter from inside the editor and observe that the redirect fails here, too.
5. Also add a different link anywhere in the newsletter content, then preview without saving, and observe that this link also fails to redirect.
6. Check out this branch. Confirm that the links from inside the test email and the preview both redirect correctly now. 
7. Repeat step 5 (previewing an unsaved link) and confirm that the link does redirect.
8. Confirm that the "clicked" count for the post has not changed since you're logged in as an admin user.

### Other information:

---
  - https://app.asana.com/0/0/1207688164534491

### DIFF
--- a/includes/tracking/class-pixel.php
+++ b/includes/tracking/class-pixel.php
@@ -148,11 +148,6 @@ final class Pixel {
 	 * @return void
 	 */
 	public static function track_seen( $newsletter_id, $tracking_id, $email_address ) {
-		// Don't track for logged-in editor or admin users.
-		if ( current_user_can( 'edit_others_posts' ) ) {
-			return;
-		}
-
 		$newsletter_tracking_id = \get_post_meta( $newsletter_id, 'tracking_id', true );
 
 		// Bail if tracking ID mismatch.
--- a/includes/tracking/class-pixel.php
+++ b/includes/tracking/class-pixel.php
@@ -148,6 +148,11 @@ final class Pixel {
 	 * @return void
 	 */
 	public static function track_seen( $newsletter_id, $tracking_id, $email_address ) {
+		// Don't track for logged-in editor or admin users.
+		if ( current_user_can( 'edit_others_posts' ) ) {
+			return;
+		}
+
 		$newsletter_tracking_id = \get_post_meta( $newsletter_id, 'tracking_id', true );
 
 		// Bail if tracking ID mismatch.
--- a/includes/tracking/class-click.php
+++ b/includes/tracking/class-click.php
@@ -160,7 +160,7 @@ final class Click {
 		if (
 			false === stripos( $newsletter_content, $url_without_query_args ) &&
 			false === stripos( $newsletter_content, urlencode( $url_without_query_args ) ) && // URL might be encoded via a block pattern.
-			! $is_admin_user // Admin user might be testing the link from a preview.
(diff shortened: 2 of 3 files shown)


leogermani (approved):

includes/tracking/class-pixel.php
```diff
@@ -148,6 +148,11 @@ public static function get_pixel_url( $post_id ) {
 	 * @return void
 	 */
 	public static function track_seen( $newsletter_id, $tracking_id, $email_address ) {
+		// Don't track for logged-in editor or admin users.
+		if ( current_user_can( 'edit_others_posts' ) ) {
+			return;
+		}
+
```
miguelpeixe: This is called on `::process_logs()`, which happens asynchronously and does not have a session.

dkoo: Ah, you're right—I guess we would need to move this to where the logfile is being written to?

miguelpeixe: That would be pretty complex. The purpose of the separate PHP file for the pixel is never to load WP on a pixel hit. Although the cookie exists in the browser, there's nothing to validate it.

miguelpeixe: One way of doing it is to create a new secret token to be stored as a cookie on users that the pixel should skip.

The pixel php can just match that single secret on a cookie and bail.

I'm not a fan, but it works.

dkoo: We can skip this for now—I doubt the extra hits from admins and editors is so great that it's skewing the numbers in a significant way

dkoo: e43d4b2

miguelpeixe (approved):

PR #1657, opened 2024-09-13
refactor: consolidate ESP data getters into Redux store

dkoo:
### All Submissions:

Just some reorganization to consolidate things. All newsletter data fetched from or written to via the connected ESP's API should be routed through our `newspack/newsletters` Redux store. This now includes:

- Fetching `newsletterData` from the ESP
- Fetching send lists or sublists from the ESP (this has been moved from the `NewsletterEdit` component to the store with its own actions, dispatchers, and hook functions, so there's no longer a need to pass the function down the component tree)
- Fetching sync errors from the most recent sync attempt with the ESP (this has been moved from the `MJML` component to the store, too)

Hopefully this reduces the effort to trace where this ESP data is coming from or updated no matter where you are in the editor component hierarchy.

There should be no major changes in functionality or behavior compared with #1632, except that we're now able to more explicitly show error messages in the `<SendTo />` component if it fails to fetch send lists for any reason. All other changes are purely conceptual and organizational.

### Other information:

### DIFF
--- a/src/editor/mjml/index.js
+++ b/src/editor/mjml/index.js
@@ -11,7 +11,7 @@ import { refreshEmailHtml } from '../../newsletter-editor/utils';
 /**
  * Internal dependencies
  */
-import { fetchNewsletterData, updateNewsletterDataError } from '../../newsletter-editor/store';
+import { fetchNewsletterData, fetchSyncErrors } from '../../newsletter-editor/store';
 import { getServiceProvider } from '../../service-providers';
 
 /**
@@ -119,13 +119,7 @@ function MJML() {
 						fetchNewsletterData( postId );
 
 						// Check for sync errors after refreshing the HTML.
-						apiFetch( {
-							path: `/newspack-newsletters/v1/${ postId }/sync-error`,
-						} ).then( ( error ) => {
-							if ( error?.message ) {
-								updateNewsletterDataError( error );
-							}
-						} );
+						fetchSyncErrors( postId );
 					}
 				} );
 		}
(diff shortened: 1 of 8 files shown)


miguelpeixe (changes requested):
Something strange happens to me:

https://github.com/user-attachments/assets/bfeccc69-82eb-4d5e-8363-e22f6b5c79cf

I'm able to select the list and sublist without issues, but then it clears the sublist after a few seconds and after a few more seconds it restores it. No interaction from my part after the initial selection.

dkoo:
@miguelpeixe I had a hard time replicating this bug exactly as it appears in your video, but I think it's happening because there are multiple requests to `retrieve` or fetch send lists happening at the same time.

- 6c82b6e attempts to avoid this issue by not starting new fetch requests from the store if one is already in progress, and also makes the data chain a bit easier to follow by moving the `<SendTo />` component from the top-level `NewsletterEdit` component into the `Sidebar` component.
- 2a28758 implements a trigger to prefetch recent sublists as soon as you select a new list, to make the Autocomplete for sublists appear to be more performant.
- a8e10da implements some checks for if we have a selected list or sublist but no matching info, and fetches the info if needed.
- b27f195 fixes a minor issue caused by a8e10da where `retrieve` requests after saving can overwrite previously fetched list and sublist info, possibly triggering unnecessary requests to refetch that info for the selected items. Now any list/sublist info in store data will be retained in the `fetchNewsletterData` payload.

miguelpeixe (approved):
Thank you for the revisions and thorough explanation. It's working flawlessly now!

PR #1660, opened 2024-09-16
fix: automatically retry failed scheduled sends

dkoo:
### All Submissions:

Implements an auto-retry system for scheduled newsletter sends, and adds logging for scheduled send failures.

#831 fixed handling for failed scheduled sends by resetting the post to `draft` status if a scheduled send failed to happen, but by resetting to `draft` status the newsletter must be manually sent or rescheduled. This PR will attempt to automatically retry the send in exponentially increasing intervals up to a maximum of 10 times.

1. Check out this branch.
2. Add `return new WP_Error( 'newspack_newsletters_error', 'Temporary error' );` to the first line of the `Newspack_Newsletters_Service_Provider::send_newsletter()` method to force a failure upon scheduled send.
3. Schedule a newsletter to be sent at some point in the near future.
4. Refresh the newsletter in the editor after the scheduled send time has passed and confirm you see an error message about the failed send at the top of the editor that says "Error sending newsletter: Temporary error". Also confirm that the post remains in "Future" status with a scheduled send time of some minutes in the future.
5. Wait a while or bump the `MAX_SCHEDULED_RETRIES` number down to something like 3 and wait a bit less, then refresh the newsletter in the editor again. Confirm that the post has reverted to "Draft" status with the error message "Failed to send 10 times. Please check the provider connection and try sending again."
6. Schedule the newsletter to be sent again at some point in the future, then wait for it to fail once.
7. Remove the temporary error message you added in step 2 and wait for the next attempt—this time confirm the scheduled send succeeds and that the newsletter post successfully transitions to "Private" or "Published" status.

### Other information:

### DIFF
--- a/includes/service-providers/class-newspack-newsletters-service-provider.php
+++ b/includes/service-providers/class-newspack-newsletters-service-provider.php
@@ -194,14 +194,12 @@ abstract class Newspack_Newsletters_Service_Provider implements Newspack_Newslet
 			update_post_meta( $post->ID, 'sending_scheduled', true );
 			$result = $this->send_newsletter( $post );
 			if ( is_wp_error( $result ) ) {
-				$prior_attempts = intval( get_post_meta( $post->ID, 'scheduled_send_attempts', true ) );
-				update_post_meta( $post->ID, 'scheduled_send_attempts', $prior_attempts + 1 );
 				$this->add_send_error( $post->ID, $result );
+				$send_errors   = get_post_meta( $post->ID, 'newsletter_send_errors', true );
+				$send_attempts = is_array( $send_errors ) ? count( $send_errors ) : 0;
 
 				// If we've already tried to send this post too many times, give up.
-				if ( self::MAX_SCHEDULED_RETRIES <= $prior_attempts ) {
-					delete_post_meta( $post->ID, 'sending_scheduled' );
-					delete_post_meta( $post->ID, 'scheduled_send_attempts' );
+				if ( self::MAX_SCHEDULED_RETRIES <= $send_attempts ) {
 					wp_update_post(
 						[
 							'ID'          => $post->ID,
@@ -216,6 +214,7 @@ abstract class Newspack_Newsletters_Service_Provider implements Newspack_Newslet
 							self::MAX_SCHEDULED_RETRIES
 						)
 					);
+					$this->add_send_error( $post->ID, $max_attempts );
 					do_action(
 						'newspack_log',
 						'newspack_esp_scheduled_send_error',
@@ -416,12 +415,12 @@ abstract class Newspack_Newsletters_Service_Provider implements Newspack_Newslet
 			$existing_errors = [];
 		}
 		$error_message = $error->get_error_message();
-		$errors[] = [
+		$existing_errors[] = [
 			'timestamp' => time(),
 			'message'   => $error_message,
 		];
-		$errors   = array_slice( $errors, -10, 10, true );
-		update_post_meta( $post_id, 'newsletter_send_errors', $errors );
+		$existing_errors   = array_slice( $existing_errors, -10, 10, true );
+		update_post_meta( $post_id, 'newsletter_send_errors', $existing_errors );
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


includes/service-providers/class-newspack-newsletters-service-provider.php
```diff
@@ -187,14 +189,77 @@ public function transition_post_status( $new_status, $old_status, $post ) {
...
+						'post_date_gmt' => gmdate( 'Y-m-d H:i:s', strtotime( current_time( 'Y-m-d H:i:s', true ) . ' +5 minutes ' ) ),
+						'post_status'   => 'future', // Reset status to `future` so the newspack_scheduled_post_checker job retries it.
+					]
+				);
+
+				do_action(
+					'newspack_log',
+					'newspack_esp_scheduled_send_error',
+					sprintf(
+						'Error scheduling send for post ID: %d',
```
miguelpeixe: This message could be improved. The scheduling part was successful, it's the sending that failed.

```suggestion
						'Error sending scheduled newsletter ID: %d',
```

includes/service-providers/class-newspack-newsletters-service-provider.php
```diff
@@ -187,14 +189,77 @@ public function transition_post_status( $new_status, $old_status, $post ) {
 			return;
 		}
 
+		// Handle scheduled newsletters.
 		if ( in_array( $new_status, self::$controlled_statuses, true ) && 'future' === $old_status ) {
 			update_post_meta( $post->ID, 'sending_scheduled', true );
 			$result = $this->send_newsletter( $post );
 			if ( is_wp_error( $result ) ) {
+				$prior_attempts = intval( get_post_meta( $post->ID, 'scheduled_send_attempts', true ) );
```
miguelpeixe: Why not leverage the existing `newsletter_send_errors` meta for this? It's persisted and each error entry is already an attempt.

miguelpeixe (changes requested):
When a newsletter send fails, it'll always [send an email to the admin](https://github.com/Automattic/newspack-newsletters/blob/82c032c90c779cb6b8bd5e5aecb2cecb240605bd/includes/service-providers/class-newspack-newsletters-service-provider.php#L452-L481). Given that we're now adding a new retry layer for scheduled newsletters, it's probably best to suppress these emails until the attempts exhaust, otherwise they might get 5 emails regarding the same newsletter. More confusingly, they might get an email when the first attempt fails, only to get resolved by the second attempt.

includes/service-providers/class-newspack-newsletters-service-provider.php
miguelpeixe: Maybe also add `Attempt x/5`

miguelpeixe:
It might be a good idea to add some exponential backoff. If the first attempt failed because of a server hiccup, it would be fine to try again sooner, like 1 minute, then exponentially try on 2, 4, and 8 minutes. We use something similar for webhooks:

https://github.com/Automattic/newspack-plugin/blob/6e8d7205852f82367fe49c748aebb8a9d09491c9/includes/data-events/class-webhooks.php#L763-L765

miguelpeixe:
I left the attempts running while I reviewed, after it exhausted all 5 attempts the newsletter ended up being marked as sent 🤔 

image

image

includes/service-providers/class-newspack-newsletters-service-provider.php
dkoo: That's a good idea!

dkoo: f430275 changes the approach to use the `newsletter_send_errors` meta.

dkoo:
@miguelpeixe thanks for the suggestions!

> I left the attempts running while I reviewed, after it exhausted all 5 attempts the newsletter ended up being marked as sent 🤔

f430275 should fix this. It was happening because we were deleting the `sending_scheduled` meta when we hit max retries. 

> When a newsletter send fails, it'll always [send an email to the admin](https://github.com/Automattic/newspack-newsletters/blob/82c032c90c779cb6b8bd5e5aecb2cecb240605bd/includes/service-providers/class-newspack-newsletters-service-provider.php#L452-L481). Given that we're now adding a new retry layer for scheduled newsletters, it's probably best to suppress these emails until the attempts exhaust, otherwise they might get 5 emails regarding the same newsletter. More confusingly, they might get an email when the first attempt fails, only to get resolved by the second attempt.

f430275 also addresses this. If the post has a `sending_scheduled` meta, it will only send an email after the last attempt, and it will include all error messages in the email body. If the post doesn't have the meta, it means it failed on a regular send attempt, so the email will get sent. We could consider implementing the retry system for regular sends too, but I think that's out of scope for this PR.

> It might be a good idea to add some exponential backoff. If the first attempt failed because of a server hiccup, it would be fine to try again sooner, like 1 minute, then exponentially try on 2, 4, and 8 minutes.

bbb656b implements the exponential backoff and also bumps the max retries up to 10, because why not?

miguelpeixe (approved):
Awesome! It's working like a charm now. Thank you for the revisions.

miguelpeixe:
Whoops, seems like something is up with the email notification.

When sending regularly, the error message includes the link to edit the newsletter. When scheduled, the link is missing:

image

That's because `get_edit_post_link()` returns `null` if `current_user_can( 'edit_post', $post->ID )` is `false`, which will be the case in a cron/scheduled action. Since there's also a small bug, unrelated to this PR, in which the link is currently broken because it's formatted with `&amp;` instead of `&` [due to the default `$context` second argument](https://developer.wordpress.org/reference/functions/get_edit_post_link/#parameters), I think it's ok to switch to `admin_url( 'post.php?post=' . $post_id . '&action=edit' );`. This will fix the existing bug and allow the scheduled action to retrieve the URL.

dkoo:
@miguelpeixe ad2246c makes the suggested change and seems to work well for me.

miguelpeixe (approved):

PR #1661, opened 2024-09-18
feat: add link to campaign in provider

miguelpeixe:
### All Submissions:

Add a link to view/edit the campaign in the provider:

image

It works for Mailchimp, ActiveCampaign, and Constant Contact.

Campaign Monitor is not included because we don't sync the newsletter campaign data with the provider. It's only pushed when sending the campaign.

1. For each ESP, start a draft and save
2. Note that the button appears once the campaign is synced to the ESP

For ActiveCampaign, the campaign is only synced once you have filled sender info and a list, so it should only become available after saving these fields.

### Other information:

### DIFF
--- a/includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign.php
+++ b/includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign.php
@@ -799,6 +799,14 @@ final class Newspack_Newsletters_Active_Campaign extends \Newspack_Newsletters_S
 				'sublists'                          => $send_sublists,
 			];
 
+			if ( $campaign_id ) {
+				$newsletter_data['link'] = sprintf(
+					'https://%s.activehosted.com/app/campaigns/%d',
+					explode( '.', str_replace( 'https://', '', $this->api_credentials()['url'] ) )[0],
+					$campaign_id
+				);
+			}
+
 			// Handle legacy sender meta.
 			$from_name   = get_post_meta( $post_id, 'senderName', true );
 			$from_email  = get_post_meta( $post_id, 'senderEmail', true );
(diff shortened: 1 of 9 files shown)


dkoo (approved):
This is working great, with the caveat that for ActiveCampaign I could only get the button to appear after sending (even after filling in Sender and List info and refreshing). But I think this is just due to limitations on the AC side.

One other bit of non-blocking feedback: do you think it would make more sense to move the button to the "Newsletter" sidebar panel? We could also consider renaming this panel to "Newsletter Campaign". All of the info here except for the "Public newsletter" option pertains to campaign settings. In fact, maybe "Public newsletter" should go up to where you put the campaign link button. 😆 WDYT?

Screenshot 2024-09-18 at 4 10 41 PM

leogermani:
> for ActiveCampaign I could only get the button to appear after sending (even after filling in Sender and List info and refreshing). But I think this is just due to limitations on the AC side.

That's my experience as well. You have to at least send a test email

miguelpeixe:
> One other bit of non-blocking feedback: do you think it would make more sense to move the button to the "Newsletter" sidebar panel? We could also consider renaming this panel to "Newsletter Campaign". All of the info here except for the "Public newsletter" option pertains to campaign settings. In fact, maybe "Public newsletter" should go up to where you put the campaign link button. 😆 WDYT?

That makes a lot of sense! Thanks, Derrick!

I switched the options around in 8aaa5c67ac9b88d73d9e5392b9f0b6c49efc1cf3, and it looks much better.

Regarding AC, as far as the "Campaign Link" is concerned, whenever a campaign is synced and becomes available it will use it. I don't think we should make this PR go around enforcing a sync just for the sake of the link. AC is a bit more complicated to manage the campaign. For the test email we have to create a disposable campaign and delete it afterwards because some of the fields are not editable after the campaign is created.

Side note: isn't it impressive that none of the links can be a simple `sprintf( 'https://esp.com/edit/%d', $campaign_id );`? 😄

dkoo (approved):
Looking great!

PR #1658, opened 2024-09-13
epic: consolidate newsletter editor data and UI

dkoo:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/trunk/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

This PR contains changes from #1629, #1630, #1631, and #1632 which were already tested and code reviewed, so it should only require a quick review of the changeset instead of a full code review.

### How to test the changes in this Pull Request:

If anything, just ensure that the newsletter editor loads correctly and lets you select/save sender and send list info without any fatals.

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

PR #1663, opened 2024-09-20
fix: parse legacy layout defaults

dkoo:
### All Submissions:

This PR parses legacy `newsletterData` metadata on preexisting custom newsletter layouts to migrate sender and send-to settings to the newer, leaner, cross-ESP-compatible `campaign_defaults` meta field for layout CPT posts. This should make #1658 NOT a breaking change, as publishers' preexisting custom layouts will be able to seamlessly their data.

1. On `trunk` or `release`, create a new newsletter, set sender email + name and list/sublist options, then save as a new custom layout.
2. Check out this branch, then create a new newsletter using the custom layout from step 1. Confirm that the Sender Name/Email and Send To options are prepopulated from the layout settings.
3. Repeat for Mailchimp, ActiveCampaign, and ~Constant Contact~ (Campaign Monitor is not handled as there is no preexisting legacy data to handle for this ESP). EDIT: also, Constant Contact does not seem to work as creating a new campaign always seems to set the top-level list by default, but this is also the case in `trunk`, so I think this is non-blocking for now.

### Other information:

### DIFF
--- a/includes/class-newspack-newsletters-layouts.php
+++ b/includes/class-newspack-newsletters-layouts.php
@@ -231,6 +231,12 @@ final class Newspack_Newsletters_Layouts {
 					if ( ! empty( $campaign_info['sublist_id'] ) ) {
 						$campaign_defaults['send_sublist_id'] = $campaign_info['sublist_id'];
 					}
+					if ( ! empty( $campaign_info['senderName'] ) ) {
+						$campaign_defaults['senderName'] = $campaign_info['senderName'];
+					}
+					if ( ! empty( $campaign_info['senderEmail'] ) ) {
+						$campaign_defaults['senderEmail'] = $campaign_info['senderEmail'];
+					}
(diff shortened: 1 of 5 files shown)


adekbadek (changes requested):
WFM with Mailchimp, but not with AC – the list was saved, but the sender not:

image

dkoo:
@adekbadek good catch, I had a mistake with the `senderName` meta field name for the AC method, plus the legacy `newsletterData` has a different shape for AC vs. the other ESPs. c567728 should fix both issues.

adekbadek (approved):
Works as expected, but I see file changes – changelog, package.json, etc – which should not be part of this PR. Pre-approving, but these should be removed.

dkoo:
Whoops, that's because I had updated `epic/update-editor-uis` locally but forgot to push. Should be good now, thanks!

PR #1664, opened 2024-09-20
fix: ignore (but log) local list errors when updating contact lists

dkoo:
### All Submissions:

This addresses something we've seen out in the wild where the `Newspack_Newsletters\Plugins\Woocommerce_Memberships::handle_membership_status_change()` method can fail a bit too easily if there are old local subscription lists hanging around on the site. This PR will log any errors that result from updating a user's local subscription lists, but will not halt the operation for those errors, which should allow updates for other non-local lists to continue. Such errors could happen in cases like the following:

- If you set up a subscription list (local or otherwise) with a membership plan and then change the connected ESP platform or account (this is the easiest way to replicate the issue and the basis of the testing instructions below)
- If you set up a local subscription list with a membership plan and then subsequently delete the corresponding tag in the connected ESP
- If you set up a local subscription list with a membership plan and then delete the local subscription list without removing it from the membership plan

1. On `trunk`, connect your site to an ESP and create a local subscription list in **Engagement > Newsletters**. Then, switch your ESP to another platform and create another local subscription list for this ESP.
2. Create a membership plan restricting content and tied to subscription product ownership and add the subscription lists for both ESPs.
3. As a reader, purchase the required subscription product.
4. As an admin, change the reader's subscription to "On Hold", "Cancelled", or "Expired" status.
5. Observe that the reader is likely not unsubscribed from the subscription list. This is because unlike the error handling for non-local subscription lists in the ESP classes' `update_contact_lists()` methods (which gather errors that occur and report them all after processing all passed lists), any error encountered when updating local subscription lists will halt the update process. Processing restricted subscription lists for other ESPs will throw errors because they're not configured for the current ESP.
6. Check out this branch.
7. As an admin, reactivate the reader's subscription, then deactivate again with an inactive status.
8. This time, confirm that the reader is removed from the restricted list.

### Other information:

---
  - https://app.asana.com/0/0/1206548767599372

### DIFF
--- a/includes/service-providers/class-newspack-newsletters-service-provider.php
+++ b/includes/service-providers/class-newspack-newsletters-service-provider.php
@@ -687,6 +687,7 @@ Error message(s) received:
 								'file'       => 'newspack_' . $this->service,
 							]
 						);
+						continue;
 					}
 					$list_settings = $list->get_provider_settings( $this->service );
 
--- a/includes/service-providers/class-newspack-newsletters-service-provider.php
+++ b/includes/service-providers/class-newspack-newsletters-service-provider.php
@@ -649,8 +649,8 @@ Error message(s) received:
 			return true;
 		}
 		if ( static::$support_local_lists ) {
-			$lists_to_add            = $this->update_contact_local_lists( $email, $lists_to_add, 'add', true );
-			$lists_to_remove         = $this->update_contact_local_lists( $email, $lists_to_remove, 'remove', true );
+			$lists_to_add    = $this->update_contact_local_lists( $email, $lists_to_add, 'add', true );
+			$lists_to_remove = $this->update_contact_local_lists( $email, $lists_to_remove, 'remove', true );
 		}
 		return $this->update_contact_lists( $email, $lists_to_add, $lists_to_remove );
 	}
--- a/includes/service-providers/class-newspack-newsletters-service-provider.php
+++ b/includes/service-providers/class-newspack-newsletters-service-provider.php
@@ -687,6 +687,7 @@ Error message(s) received:
 								'file'       => 'newspack_' . $this->service,
 							]
 						);
+						unset( $lists[ $key ] );
 						continue;
 					}
 					$list_settings = $list->get_provider_settings( $this->service );
--- a/includes/service-providers/class-newspack-newsletters-service-provider.php
+++ b/includes/service-providers/class-newspack-newsletters-service-provider.php
@@ -649,8 +649,8 @@ Error message(s) received:
 			return true;
 		}
 		if ( static::$support_local_lists ) {
-			$lists_to_add    = $this->update_contact_local_lists( $email, $lists_to_add, 'add', true );
-			$lists_to_remove = $this->update_contact_local_lists( $email, $lists_to_remove, 'remove', true );
+			$lists_to_add    = $this->update_contact_local_lists( $email, $lists_to_add, 'add' );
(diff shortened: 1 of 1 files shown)


includes/service-providers/class-newspack-newsletters-service-provider.php
```diff
@@ -649,13 +649,47 @@ public function update_contact_lists_handling_local( $email, $lists_to_add = [],
...
+							'provider' => $this->service,
+							'errors'   => $removed_result->get_error_messages(),
+						],
+						'user_email' => $email,
+						'file'       => 'newspack_' . $this->service,
+					]
+				);
 			}
 		}
 		return $this->update_contact_lists( $email, $lists_to_add, $lists_to_remove );
```
leogermani, line 655: This method expects a modified version of the `$lists_to_add` variable. But you renamed the variable up there and now you're passing the raw value.

Some additional handling is needed there to pass only the remote lists.

Easiest way of doing this is if we use the `Subscription_List` objects instead of the ID, because then you have the `is_local` method and you can call the appropriate method depending on the list type....

You can also get the objects in the `Newspack_Newsletters_Contacts::add_and_remove_lists` method, similar to what we do in the `upsert` method (https://github.com/Automattic/newspack-newsletters/blob/trunk/includes/class-newspack-newsletters-contacts.php#L188). And then this method would receive objects instead of strings

leogermani (changes requested):

includes/service-providers/class-newspack-newsletters-service-provider.php
dkoo: I think the renaming of the variable isn't needed, and I don't want to change more than necessary to avoid the errors halting the operation. 16efc54 moves the error logging to the `update_contact_local_lists` method, which is what's causing the issue anyway. This preserves the behavior of modifying the `$lists_to_add` and `$lists_to_remove` arrays while also letting the loop proceed in the event of errors.

includes/service-providers/class-newspack-newsletters-service-provider.php
```diff
@@ -676,7 +670,25 @@ protected function update_contact_local_lists( $email, $lists = [], $action = 'a
 					$list = Subscription_List::from_public_id( $list_id );
 
 					if ( ! $list->is_configured_for_provider( $this->service ) ) {
-						return new WP_Error( 'List not properly configured for the provider' );
+						do_action(
+							'newspack_log',
+							'newspack_esp_update_contact_lists_error',
+							sprintf(
+								'Error handling local list: %s',
```
leogermani: NIT: We're still learning how to use the logs, so this is just a suggestion, not sure what's the best way.

But it seems that the "message" here is a bit redundant. Since it's the same in all cases, it has the same purpose of the `error_code` up there (`newspack_esp_update_contact_lists_error`)

Maybe the string that is down there as `errors` in the `data` array could be added here and removed from there.

Also. Not sure if it's a good thing to add this dynamic value of the ID here in the message. This will give us an infinite number of possible values making it a little more difficult to filter by one specific error. Maybe the ID can go down there in the data?

leogermani (approved):
Nice! 

I left one non-blocking comment

includes/service-providers/class-newspack-newsletters-service-provider.php
dkoo: @leogermani makes sense. 710ea24 updates the error logs so that the messages are evergreen in both cases, and moves the `$list_id` to a data param instead.

leogermani (approved):

PR #1666, opened 2024-10-03
Fix: Dynamically fetch message ID in ActiveCampaign test emails

claudiulodro:
### All Submissions:

**Note: This is a hotfix PR.**

This PR fixes an issue I discovered on some client sites which is also reproducible in my local environment. 

When a new temporary campaign is created in ActiveCampaign to send a test email, it uses [the message ID saved onto the post](https://github.com/Automattic/newspack-newsletters/blob/616666ced5b44c15e67a67eaed4b4bc7721a8c31/includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign.php#L777). This message is not necessarily associated with the temporary campaign, so the test send will fail.

I'm not sure if the approach I took here is the way to go, or if it would be better to somehow associate the message from the "real" campaign with the temporary campaign, so that this check isn't needed. Let me know what you think.

1. This part happens naturally for every new newsletter on my local running only the trunk version of Plugin, Blocks, Ads, and Newsletters, but it did not happen in @adekbadek's testing. Start a new newsletter from a template, add all the necessary information, save it as draft, and send a test email.
2. Before applying this patch you should get a "Sending test campaign failed: Message not sent" error:
Screenshot 2024-10-03 at 2 20 24 PM
4. After applying this patch it should successfully send:
Screenshot 2024-10-03 at 2 21 05 PM

### Other information:

### DIFF
--- a/includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign.php
+++ b/includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign.php
@@ -659,6 +659,24 @@ final class Newspack_Newsletters_Active_Campaign extends \Newspack_Newsletters_S
 			return $campaign;
 		}
 		add_post_meta( $post_id, 'ac_test_campaign', $campaign['id'] );
+
+		/** Get the latest message ID from the temporary campaign. */
+		$campaign_data = $this->api_v1_request(
+			'campaign_list',
+			'GET',
+			[
+				'query' => [
+					'action' => 'test',
+					'ids'    => $campaign['id'],
+				],
+			]
+		);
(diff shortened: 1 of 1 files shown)


dkoo (approved):
I was able to replicate the issue after repeatedly sending multiple tests. This fixed the issue for me.

dkoo:
@claudiulodro if this is supposed to be a hotfix, should it be rebased on `release`?

claudiulodro:
Thanks for the quick review! 

I always remember to branch off release but never remember to merge to release. 😆 Fixed now. I'll merge Monday so it doesn't deploy Friday afternoon.

PR #1673, opened 2024-10-16
fix(mailchimp): unsetting sublists in synced campaign

dkoo:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/trunk/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

#1658 had a bug in the `get_sync_payload` method for Mailchimp. If you intend to send a newsletter campaign to the entire list, the sync payload must include a `segment_opts` param with a value of an empty object, otherwise the campaign recipients won't be updated.

Also fixes a bug in the handling for cached list data in which the cache for audiences is never expired.

### How to test the changes in this Pull Request:

1. On `trunk`, using Mailchimp, create a new newsletter post and set an audience and a sublist, then save.
2. Clear the sublist and save. Observe that in the synced campaign in Mailchimp, the recipients are still set to the sublist selected in step 1.
3. Check out this branch and resave the newsletter without any sublist. Confirm that the synced campaign in Mailchimp is updated to send to all contacts in the selected audience.
4. To test the cache expiration for audiences, note the contact count shown for your audience in the editor sidebar. Add some contacts to your list, wait >20 minutes (or run `wp option delete newspack_nl_mailchimp_cache_date_lists` to invalidate the cache immediately) and refresh the editor, then confirm that the contact count is updated.

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

PR #1676, opened 2024-10-21
fix: lists cache flushing

leogermani:
### All Submissions:

Fixes two issues with flushing the cache for listings.

First, the `is_cache_expired` method, when called without an argument, will call `get_cache_date_key( NULL )`. In this case, NULL is considered an argument with causes `get_cache_date_key()` not to use the default `lists` value, and we end up returning an invalid cache key.

The result is that the check to see if the Lists cache was expired was returning `false` for ever.

Second, when we trigger a cache refresh via cron, the lists were not being refreshed right away. We would refresh each lists cache but based on a cached list of lists. 

1. Configure the plugin and make sure you have lists cached
2. Enter a different api key for another account
3. run `wp cron event run newspack_nl_mailchimp_refresh_cache` and make sure the cache is properly refreshed

### Other information:


dkoo:
This is handled in #1673, so closing! if that one doesn't solve the issue entirely for whatever reason, feel free to reopen.

PR #1680, opened 2024-10-22
Adjust nonce logic

claudiulodro:
### All Submissions:

**Note: This is a hotfix**

This PR adjusts the nonce logic for newsletter signups. Now it will only check the nonce if reCaptcha is inactive. Technically we don't even need a nonce for logged-out forms, but I like having the little extra security to protect from bot signups.

1. Try the newsletter subscribe block. It should continue working nicely with no visible change.

### Other information:

### DIFF
--- a/src/blocks/subscribe/index.php
+++ b/src/blocks/subscribe/index.php
@@ -180,7 +180,7 @@ function render_block( $attrs ) {
 	>
 		<?php if ( ! $subscribed ) : ?>
 			<form id="<?php echo esc_attr( get_form_id() ); ?>">
-				<?php \wp_nonce_field( FORM_ACTION, FORM_ACTION ); ?>
+				<input type="hidden" name="<?php echo esc_attr( FORM_ACTION ); ?>" value="1" />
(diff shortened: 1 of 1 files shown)


leogermani:
>  Technically we don't even need a nonce for logged-out forms, but I like having the little extra security to protect from bot signups.

But doesn't that mean that the cached nonce will still be an issue for sites not using recaptcha?

Also, I don't see any added protection there, since the nonce is part of the form as a hidden field anyways

claudiulodro:
It's at least as much protection as the honeypot field! :) Good points though, I'll just remove it

claudiulodro:
Adjusted in https://github.com/Automattic/newspack-newsletters/pull/1680/commits/f1aed62d5b5d7af78ef1ab1d4d5890d617b45f2f

leogermani (approved):
Thanks for revisiting it!

Tip for your future self, in cases like this you can use `phpcs:disable Rule....` and then `phpcs:enable` at the end to avoid all these repetitions

PR #1681, opened 2024-10-23
fix: preview & send buttons on WP 6.7

adekbadek:
### All Submissions:

WP 6.7 updates some styling, which breaks our custom buttons:

| 6.6 | 6.7 | 
|--------------|-----------|
| image | image |

This PR fixes that, while ensuring compatibility with 6.6.

1. On `trunk`, switch WP to 6.7 (`wp core update --version=6.7-beta3`)
2. Open a newsletter or create a new on, observe broken UI
3. Switch to this branch, observe UI is now fine

### Other information:

### DIFF
--- a/src/components/send-button/index.js
+++ b/src/components/send-button/index.js
@@ -298,7 +298,7 @@ export default compose( [
 		}
 
 		return (
-			<Fragment>
+			<div style={{ display: 'flex' }}>
 				<PreviewHTMLButton />
 				<Button
 					className="editor-post-publish-button"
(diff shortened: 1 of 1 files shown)


laurelfulford (approved):
Such a good find, @adekbadek! 🙌 

The buttons definitely look better.

One additional thing I noticed is that the buttons have also been reordered for posts and pages in WP 6.7, so they appear at the end: 

CleanShot 2024-10-23 at 10 32 42

Do you see any issues with doing the same for Newsletters? If we did, we'd probably want to do it a little closer to the actual WP 6.7 release (so the change drops around the same time); if it sounds like an okay move I can spin up a separate issue for it. 

Thanks!

adekbadek:
> if it sounds like an okay move I can spin up a separate issue for it.

Please do, thanks! :)

PR #1682, opened 2024-10-24
fix: strip Mailchimp merge tags from email preview text

dkoo:
### All Submissions:

If you don't fill in the preview text field, preview text will be generated from the first 60 words of post content. However, if that generated preview text contains an unclosed [Mailchimp conditional merge tag](https://mailchimp.com/help/use-conditional-merge-tag-blocks/), with an `ELSE` statement in between, this will break Mailchimp's email renderer, potentially resulting in a totally blank email being sent. Not good!

To avoid this issue, this PR strips _all_ Mailchimp merge tags from the preview text string, generated or not. 

1. On `release`, using Mailchimp as the connected ESP, create a new newsletter and add a conditional `IF...ELSE` merge tag block at the beginning of the post content. Make sure the preview text will contain the `IF` and `ELSE` tags before the first 60 words, but after that is long enough to push the closing tag out of the first 60 words, e.g.:

```
*|IF:MMERGE7=|*

Lorem ipsum dolor sit amet, consectetur adipiscing elit. Quisque ut lorem at ex hendrerit finibus. Pellentesque habitant morbi tristique senectus et netus et malesuada fames ac turpis egestas.

*|ELSEIF:MMERGE7=Testing|*

Suspendisse posuere enim et libero vehicula, non faucibus tortor vehicula. Nam sit amet iaculis sem, et porta elit. Phasellus pretium aliquet ligula at elementum. Proin vitae mauris eget ex aliquet vulputate. Lorem ipsum dolor sit amet, consectetur adipiscing elit. Proin non scelerisque neque. Etiam iaculis ligula non pellentesque dignissim. Sed at ultricies massa, eget varius nisl.

*|END:IF|*
```

2. Save the draft and send yourself a test email. It should appear completely blank. You can also view the raw HTML of the email (`wp post meta get <post ID> newspack_email_html`) and see that the beginning of the content contains the unclosed `IF...ELSE` tags.

Screenshot 2024-10-24 at 3 30 24 PM

3. Check out this branch, refresh the editor and save again, then send another test email. Confirm that the email's content renders correctly and that the beginning of the raw HTML contains the preview text but stripped of the `IF...ELSE` tags.

### Other information:

---
  - https://app.asana.com/0/0/1208622148217455

### DIFF
--- a/includes/class-newspack-newsletters-renderer.php
+++ b/includes/class-newspack-newsletters-renderer.php
@@ -1269,11 +1269,8 @@ final class Newspack_Newsletters_Renderer {
 		 */
 		$body             = self::post_to_mjml_components( $post ); // phpcs:ignore WordPressVIPMinimum.Variables.VariableAnalysis.UnusedVariable
 		$background_color = get_post_meta( $post->ID, 'background_color', true );
-		$preview_text     = get_post_meta( $post->ID, 'preview_text', true );
+		$preview_text     = self::get_preview_text( $post );
 		$custom_css       = get_post_meta( $post->ID, 'custom_css', true );
-		if ( ! $preview_text ) {
-			$preview_text = wp_trim_words( wp_strip_all_tags( $body ), 60 );
-		}
 		if ( ! $background_color ) {
 			$background_color = '#ffffff';
 		}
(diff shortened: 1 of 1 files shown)


adekbadek (changes requested):
> Confirm that the email's content renders correctly and that the beginning of the raw HTML contains the preview text but stripped of the IF...ELSE tags.

The preview is now not-empty, but the email content still is. Looking at the code here, I see only the preview text is affected.

dkoo:
> The preview is now not-empty, but the email content still is.

@adekbadek can you share the email editor content you're using to test?

> Looking at the code here, I see only the preview text is affected.

That's true, but the issue is happening because the preview text is a part of the rendered HTML, so unclosed tags in the preview text affect the rest of the content.

adekbadek (approved):
> can you share the email editor content you're using to test?

I was pasting the content from PR description in the code view of the editor to prevent escaping. This resulted in empty content. When pasting the content into the visual editor, it works as described.

dkoo:
Thanks for clarifying, @adekbadek! ~I'm going to hold this one for now since it's kind of an edge case~. Nevermind, it's going out now!

PR #1684, opened 2024-10-29
fix: check for error type before logging

dkoo:
### All Submissions:

Fixes a potential fatal. We log an error if a `$lists_response` variable is a WP error OR if it lacks a `lists` key, but in that latter case (not an error and lacks the key) a fatal will be thrown.

A code review should be sufficient, but to test you can add a line `$lists_response = [];` before [this line](https://github.com/Automattic/newspack-newsletters/blob/trunk/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp-cached-data.php#L559), then visit a newsletter post in the editor to trigger the fatal on `release` and confirm that the fatal doesn't happen on this branch.

### Other information:

### DIFF
--- a/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp-cached-data.php
+++ b/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp-cached-data.php
@@ -556,8 +556,8 @@ final class Newspack_Newsletters_Mailchimp_Cached_Data {
 			),
 			__( 'Error retrieving Mailchimp lists.', 'newspack_newsletters' )
 		);
-		if ( is_wp_error( $lists_response ) || empty( $lists_response['lists'] ) ) {
-			Newspack_Newsletters_Logger::log( 'Mailchimp cache: Error refreshing cache: ' . ( is_wp_error( $lists_response ) ? $lists_response->getMessage() : __( 'Error retrieving Mailchimp lists.', 'newspack_newsletters' ) ) );
+		if ( method_exists( $lists_response, 'getMessage' ) || empty( $lists_response['lists'] ) ) {
+			Newspack_Newsletters_Logger::log( 'Mailchimp cache: Error refreshing cache: ' . ( method_exists( $lists_response, 'getMessage' ) ? $lists_response->getMessage() : __( 'Error retrieving Mailchimp lists.', 'newspack_newsletters' ) ) );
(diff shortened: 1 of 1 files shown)


includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp-cached-data.php
```diff
@@ -557,7 +557,7 @@ public static function fetch_lists( $limit = null ) {
 			__( 'Error retrieving Mailchimp lists.', 'newspack_newsletters' )
 		);
 		if ( is_wp_error( $lists_response ) || empty( $lists_response['lists'] ) ) {
-			Newspack_Newsletters_Logger::log( 'Mailchimp cache: Error refreshing cache: ' . ( $lists_response->getMessage() ?? __( 'Error retrieving Mailchimp lists.', 'newspack_newsletters' ) ) );
+			Newspack_Newsletters_Logger::log( 'Mailchimp cache: Error refreshing cache: ' . ( is_wp_error( $lists_response ) ? $lists_response->getMessage() : __( 'Error retrieving Mailchimp lists.', 'newspack_newsletters' ) ) );
```
leogermani: I don't think this is a `WP_Error` object. WP_Error method is `get_message`, not `getMessage`. I think this is something from the Mailchimp lib

dkoo: How about this, then? f928095

dkoo: No wait, this 63a9178

dkoo: Argh, cb62cfe

leogermani:
@dkoo I pushed a change.

The actual problem was when I added `fetch_lists` to the cron callback. 

`->validate()` will either return a successful response or throw an Exception. You can see that all calls to the fetch methods are inside `try catch` clauses. The `getMessage` (camelCase) method comes from the `Exception` object.

You can test this PR by, for example, changing the path of the "fetch lists" request to something invalid at line 556, and then invoke the cron event. You'll see the error being logged safely. Without these changes - and with your previous changes - you would see a Fatal error

dkoo:
Thanks for untangling that, @leogermani!

claudiulodro (approved):
Solves it nicely in my testing

PR #1688, opened 2024-10-31
fix: avoid Redux usage and related errors in non-newsletter email editors

dkoo:
### All Submissions:

#1679 introduced errors in non-newsletter CPT email editors. All email editors use the MJML component to convert post block content into MJML and then into email-compliant HTML. Only the newsletter CPT editor registers a Redux store to handle newsletter campaign data, so store methods should only be called while editing a newsletter CPT post while connected to a supported ESP.

This PR makes the `MJML` component more standalone once again, and only lets it call Redux store methods when it detects a supported ESP connection (which will only happen in the newsletter editor, not other email editors). It also does some light refactoring of this component to avoid a Promise chain in favor of a less verbose and more readable `async` function.

1. On `trunk`, edit a custom Newspack email (such as the custom receipt in Reader Revenue > Emails, or the RAS emails in Engagement > Reader Activation) and a newsletter ad. In both cases, observe an error notice: `Uncaught TypeError: Cannot read properties of undefined (reading 'getIsRefreshingHtml')`
2. Check out this branch and refresh both editors and confirm that the error doesn't appear.
3. Smoke test making edits to the custom email and the newsletter ad and confirm that the changes are persisted.
4. Smoke test sending a test email from the custom email editor and confirm that the changes are reflected in the sent test email.
5. Smoke test updating a newsletter post, saving, sending a test email, and confirm that all changes made are reflected in test emails and in synced campaign in the connected ESP.

### Other information:

---
  - https://app.asana.com/0/0/1208638777229783

### DIFF
--- a/src/components/send-button/index.js
+++ b/src/components/send-button/index.js
@@ -18,8 +18,9 @@ import { get } from 'lodash';
  * Internal dependencies
  */
 import { getServiceProvider } from '../../service-providers';
-import { refreshEmailHtml, validateNewsletter } from '../../newsletter-editor/utils';
+import { validateNewsletter } from '../../newsletter-editor/utils';
 import { useNewsletterData } from '../../newsletter-editor/store';
+import { refreshEmailHtml } from '../../editor/mjml';
 import './style.scss';
 
 function PreviewHTML() {
--- a/src/editor/mjml/index.js
+++ b/src/editor/mjml/index.js
@@ -3,19 +3,52 @@
 /**
  * WordPress dependencies
  */
+import { __ } from '@wordpress/i18n';
 import apiFetch from '@wordpress/api-fetch';
 import { useSelect, useDispatch } from '@wordpress/data';
 import { useEffect } from '@wordpress/element';
-import { refreshEmailHtml, usePrevious } from '../../newsletter-editor/utils';
+import { usePrevious } from '../../newsletter-editor/utils';
 
 /**
  * Internal dependencies
  */
-import { fetchNewsletterData, fetchSyncErrors, updateNewsletterDataError, updateIsRefreshingHtml, useIsRefreshingHtml } from '../../newsletter-editor/store';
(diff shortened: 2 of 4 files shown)


leogermani (approved):
Works well.

I've noticed two little things, not sure if related or fixable in this PR

1. On custom emails: Changes done while on trunk were not present in the sent email. (I guess this is expected because there was an error there)

2. On custom emails: If I send a test email before saving, a little popup says the email was saved but the sent newsletter does not include the latest changes

dkoo:
> 1. On custom emails: Changes done while on trunk were not present in the sent email. (I guess this is expected because there was an error there)

Yes, this PR should fix that issue!

> 2. On custom emails: If I send a test email before saving, a little popup says the email was saved but the sent newsletter does not include the latest changes

This is the same race condition that #1679 attempted to fix for newsletter CPT posts. The reason the race condition happens is:

- You click "test" and it initiates a save request
- When the save request completes, the `MJML` editor component triggers a refresh of email HTML via the MJML JS library, and then saves the refreshed HTML to post meta. We need to wait until the save is complete before doing this, otherwise recent changes in the editor may not be available on the server when the updated HTML is written.
- But also when the save request completes, the custom email editor sends a `test` request to send the test email, but this is likely to get sent before the refreshed email HTML gets saved to post meta as it can start at the same time as the MJML request

It's relatively easy to fix this race condition for newsletter CPTs, in which the MJML and newsletter editor components can both access the Redux store to check when the MJML request completes (and that's when we send the test email for newsletters).

But because the custom email editor is controlled by JS in the main Newspack plugin, it's more complicated to fix this for custom emails as the main plugin can't access the Redux store registered by Newsletters. I suggest we leave this as-is for now, unless you have any suggestions for fixing the race condition. I thought about adding a `setTimeout` delay to the test request for custom emails to try and give the MJML request time to complete, but that seems ugly to me.

PR #1690, opened 2024-10-31
fix(mailchimp): fatal error with invalid API keys

dkoo:
### All Submissions:

Fixes a fatal error when trying to fetch audience/list data using an invalid Mailchimp API key. The fatal was happening because our Mailchimp integration's API response validation logic only considered status codes of `400` or `404` to be errors, but an invalid API key error is returned with a status code of `401`, and unhandled WP errors are returned with status codes of `500`. So this PR will handle any response with a status code of `400` or greater as an error.

1. On `trunk`, connect your site to Mailchimp using a valid API key.
2. In the Mailchimp dashboard, revoke that API key.
3. Using WP CLI, trigger the cron job to refresh the Mailchimp cache (or just wait >20 minutes): `wp cron event run newspack_nl_mailchimp_refresh_cache`
4. Visit **Newspack > Engagement** and **Newsletters > Settings** admin pages and observe fatal errors in each.
5. Check out this branch and refresh both admin pages, and confirm no fatal error and a more helpful error message shown in both places:

Screenshot 2024-10-31 at 3 15 38 PM
Screenshot 2024-10-31 at 3 16 25 PM

6. Confirm that you can update the API key to a new valid key and that saving persists the key and restores the connection to the ESP.

### Other information:

---
  - https://app.asana.com/0/0/1208644983311128

### DIFF
--- a/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp-cached-data.php
+++ b/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp-cached-data.php
@@ -310,7 +310,7 @@ final class Newspack_Newsletters_Mailchimp_Cached_Data {
 	 * @param string|null $list_id The List ID, or null for all lists.
 	 * @param string      $error The error message.
 	 */
-	private static function maybe_add_error( $list_id = null, $error ) {
+	private static function maybe_add_error( $list_id = null, $error = '' ) {
 		Newspack_Newsletters_Logger::log(
 			sprintf(
 				'Mailchimp cache: handling error while fetching cache for %s',
@@ -319,6 +319,9 @@ final class Newspack_Newsletters_Mailchimp_Cached_Data {
 		);
 		if ( ! $list_id ) {
 			$list_id = 'lists';
+		}
+		if ( ! $error ) {
+			$error = __( 'Unknown error', 'newspack_newsletters' );
 		}
 		$cache_date = get_option( self::get_cache_date_key( $list_id ) );
 		if ( $cache_date && ( time() - $cache_date ) > self::SURFACE_ERRORS_AFTER ) {
--- a/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp-cached-data.php
+++ b/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp-cached-data.php
@@ -525,10 +525,18 @@ final class Newspack_Newsletters_Mailchimp_Cached_Data {
 	 */
 	public static function handle_cron() {
 		Newspack_Newsletters_Logger::log( 'Mailchimp cache: Handling cron request to refresh cache' );
+		delete_option( self::get_lists_cache_key() );
+		delete_option( self::get_cache_date_key() );
+
 		try {
 			$lists = self::fetch_lists(); // Force a cache refresh.
 		} catch ( Exception $e ) {
 			Newspack_Newsletters_Logger::log( 'Mailchimp cache: Error refreshing lists cache: ' . $e->getMessage() );
+			return;
+		}
+
+		if ( is_wp_error( $lists ) ) {
+			Newspack_Newsletters_Logger::log( 'Mailchimp cache: Error refreshing lists cache: ' . $lists->get_error_message() );
 			return;
 		}
 
@@ -549,7 +557,7 @@ final class Newspack_Newsletters_Mailchimp_Cached_Data {
 	public static function fetch_lists( $limit = null ) {
 		$mc = self::get_mc_api();
 		if ( \is_wp_error( $mc ) ) {
-			return [];
+			return $mc;
 		}
 		$lists_response = ( self::get_mc_instance() )->validate(
 			$mc->get(
--- a/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp-cached-data.php
+++ b/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp-cached-data.php
@@ -288,10 +288,14 @@ final class Newspack_Newsletters_Mailchimp_Cached_Data {
 	/**
 	 * Clears the cache errors for a given list
 	 *
-	 * @param string $list_id The List ID.
+	 * @param string|null $list_id The List ID, or null for all lists.
 	 * @return void
 	 */
-	private static function clear_errors( $list_id ) {
+	private static function clear_errors( $list_id = null ) {
+		if ( ! $list_id ) {
+			delete_option( self::ERRORS_OPTION );
+			return;
+		}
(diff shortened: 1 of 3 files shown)


includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php
```diff
@@ -1426,7 +1426,7 @@ public function validate( $result, $preferred_error = null ) {
 				throw new Exception( esc_html__( 'A Mailchimp error has occurred.', 'newspack-newsletters' ) );
 			}
 		}
-		if ( ! empty( $result['status'] ) && in_array( $result['status'], [ 400, 404 ] ) ) {
+		if ( ! empty( $result['status'] ) && 400 <= (int) $result['status'] ) {
```
adekbadek: Maybe it's just me, but I find this version more readable:

```suggestion
		if ( ! empty( $result['status'] ) && (int) $result['status'] >= 400  ) {
```

adekbadek (commented):
I was not able to complete the testing steps until I've emptied the cache (`newspack_nl_mailchimp_cache_lists` option). With the wrong key, the list retrieval doesn't update the lists and fails silently, keeping the cached lists.

dkoo:
> I was not able to complete the testing steps until I've emptied the cache (`newspack_nl_mailchimp_cache_lists` option). With the wrong key, the list retrieval doesn't update the lists and fails silently, keeping the cached lists.

@adekbadek aha, you're correct. 41db9b6 should fix this by returning a proper error in this scenario. So now, if you run `wp cron event run newspack_nl_mailchimp_refresh_cache` or wait 20 minutes for the cache to refresh, the cached data will no longer be returned if we fail to refresh the cache, and the error message should appear in the UIs.

includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php
dkoo: Yoda conditionals you do not like, see I do! (Me too, honestly!)

leogermani (changes requested):
> > I was not able to complete the testing steps until I've emptied the cache (`newspack_nl_mailchimp_cache_lists` option). With the wrong key, the list retrieval doesn't update the lists and fails silently, keeping the cached lists.
> 
> @adekbadek aha, you're correct. [41db9b6](https://github.com/Automattic/newspack-newsletters/commit/41db9b69e22e12c6d38a42f8402b334ce8b0a4d9) should fix this by returning a proper error in this scenario. So now, if you run `wp cron event run newspack_nl_mailchimp_refresh_cache` or wait 20 minutes for the cache to refresh, the cached data will no longer be returned if we fail to refresh the cache, and the error message should appear in the UIs.

Wait! That's by design!

From the [docs](https://github.com/Automattic/newspack-newsletters/blob/72897f206058af5c63afdd7fd2ab9946a9a0b923/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp-cached-data.php#L25):

> If the cache refresh fails, we will store the error in a separate option, and will only surface it to the user after 20 minutes.
> In every admin page we will display a generic Warning message, telling the user to go to Newsletters > Settings to see the errors.
> In Newsletters > Settings we will output the errors details.

Maybe what we can do is clear the cache if the API key changes

dkoo:
Thanks for clarifying, @leogermani! I'll spin up another commit with that change

dkoo:
@leogermani 84044fc and 6d7602e implement the cache clear on API key change, and also add an error message to be shown when the cron job to refresh the cache fails. This will result in the following error being shown in both the Newsletters > Settings and Engagement > Newsletters pages 20 minutes after the cache refresh fails:

Screenshot 2024-11-07 at 11 51 43 AM

It will also avoid a fatal error in the newsletter editor. Instead, errors will be shown at the top of the editor and in the Newsletter Campaign sidebar:

Screenshot 2024-11-07 at 11 54 55 AM
Screenshot 2024-11-07 at 11 54 51 AM

Testing instructions remain the same.

includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp-cached-data.php
```diff
@@ -300,11 +307,19 @@ private static function clear_errors( $list_id ) {
...
 	 * Stores the last error for a given list, if the cache is older than self::SURFACE_ERRORS_AFTER
 	 *
-	 * @param string $list_id The List ID.
-	 * @param string $error The error message.
+	 * @param string|null $list_id The List ID, or null for all lists.
+	 * @param string      $error The error message.
 	 */
-	private static function maybe_add_error( $list_id, $error ) {
-		Newspack_Newsletters_Logger::log( 'Mailchimp cache: handling error while fetching cache for list ' . $list_id );
+	private static function maybe_add_error( $list_id = null, $error ) {
```
leogermani: Having an optional parameter before a required paramater throws a PHP deprecated error
```suggestion
	private static function maybe_add_error( $list_id, $error ) {
```

leogermani (dismissed):
All looks good to me. The addition to clear the cache on api key change is a good one. ( I left one minor suggestion )

There is just one problem not related to this PR.

I think we need to revisit what we did [here](https://github.com/Automattic/newspack-newsletters/pull/1636/files) (cc @adekbadek):

I've changed the api key, invalidated the cache, and then added an invalid key.

I then triggered the cron job to refresh the cache but I never saw an error message being stored and displayed (I modified the code to skip that time before showing errors)

The reason was [this](https://github.com/Automattic/newspack-newsletters/blob/4e2688e01600376d7c0e8dc397c7c9c9f748860d/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp-cached-data.php#L552).

With invalid API keys, `fetch_lists` simply returns an empty array and an error is never triggered.

If you look at the code in that PR, all the time after instantiating the api, we call the `validate` method, which throws in case of errors. That's why these methods are always invoked in a try catch statement. We should let the API instantiation throw as well, otherwise it won't trigger errors when it should

Update: Here's a PR to fix this -> https://github.com/Automattic/newspack-newsletters/pull/1698

includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp-cached-data.php
dkoo: Ah, `$list_id` needs to be optional because it's `null` if we're adding an error about refreshing the cache for audience data. 017efcc should fix it by making `$error` also optional.

dkoo:
Thanks, @leogermani! Originally this PR tried to avoid returning an empty array, but that meant that a workflow-blocking error would be thrown for temporary Mailchimp API hiccups, too. Glad you figured out a less obtrusive solution in #1698.

leogermani (approved):

PR #1698, opened 2024-11-08
fix: let the exceptions be thrown

leogermani:
### All Submissions:

See https://github.com/Automattic/newspack-newsletters/pull/1690#pullrequestreview-2424810964

This PR reverts https://github.com/Automattic/newspack-newsletters/pull/1636

The pattern we use in Newsletters is not something we use in other plugins and we keep forgetting about it.

The pattern is that instead of returning errors, the methods will throw exceptions that we need to catch. This is the basis of the `validate` method so I followed the same thing for the whole Mailchimp Cache class.

In #1636, we are supressing this errors, and the result is that they are not being surfaced when they should.

The thing that we were solving in #1636 was not an issue, it was expected behavior, you can see by the methods doc blocks that they are expected to throw.

Smoke test

### Other information:

### DIFF
--- a/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp-cached-data.php
+++ b/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp-cached-data.php
@@ -570,6 +570,7 @@ final class Newspack_Newsletters_Mailchimp_Cached_Data {
 		if ( ! $limit ) {
 			update_option( self::get_lists_cache_key(), $lists_response['lists'], false ); // auto-load false.
 			update_option( self::get_cache_date_key(), time(), false ); // auto-load false.
+			self::clear_errors();
 		}
 
 		return $lists_response['lists'];
--- a/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp-cached-data.php
+++ b/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp-cached-data.php
@@ -108,29 +108,6 @@ final class Newspack_Newsletters_Mailchimp_Cached_Data {
 	}
 
 	/**
-	 * Retrieves an instance of the Mailchimp api
-	 *
-	 * @return DrewM\MailChimp\MailChimp|WP_Error
-	 */
-	private static function get_mc_api() {
-		$api_key = self::get_mc_instance()->api_key();
-		if ( empty( $api_key ) ) {
-			return new WP_Error(
-				'newspack_newsletters_mailchimp_error',
-				__( 'Missing Mailchimp API key.', 'newspack-newsletters' )
-			);
-		}
-		try {
-			return new Mailchimp( $api_key );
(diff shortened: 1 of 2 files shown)


dkoo (changes requested):
Hmm, on this branch after changing the Mailchimp API key to something invalid and refreshing the cache, I don't see any errors in the Settings page after the `SURFACE_ERRORS_AFTER` amount of time has passed. I also see some warnings in the debug log that seem related:

```
[11-Nov-2024 23:14:17 UTC] PHP Warning:  Undefined array key "lists" in /newspack-repos/newspack-newsletters/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp-cached-data.php on line 542
[11-Nov-2024 23:14:17 UTC] PHP Warning:  Undefined array key "lists" in /newspack-repos/newspack-newsletters/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp-cached-data.php on line 546
[11-Nov-2024 23:14:17 UTC] PHP Warning:  foreach() argument must be of type array|object, null given in /newspack-repos/newspack-newsletters/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php on line 629
```

> The pattern we use in Newsletters is not something we use in other plugins and we keep forgetting about it.
> 
> The pattern is that instead of returning errors, the methods will throw exceptions that we need to catch. This is the basis of the `validate` method so I followed the same thing for the whole Mailchimp Cache class.

I wonder if we should retain this pattern since it keeps breaking our brains? Could we refactor the `validate` method to use `WP_Error`s instead of exceptions that we need to catch and handle manually? Maybe outside the scope of this PR, but I feel this pattern has anecdotally bitten us more than helped us.

leogermani:
@dkoo  I can't reproduce what you described. This branch was outdated, so it might have contributed.

Here is what I see:

* Having a healthy connection
* I change the mailchimp api key to an invalid one
* I see the error right away (because the cache is invalidated)
* I go to line 304 and comment the `if` out, so the errors are stored right away
* trigger the cron to refresh the cache (`wp cron event run newspack_nl_mailchimp_refresh_cache`)
* I see that persistent error alert in all pages

Also, I noticed that we were never clearing the `lists` error. Fixed in https://github.com/Automattic/newspack-newsletters/pull/1698/commits/e0c700ba259bd3275334824ff8ec341376f0f53c

leogermani:
> I wonder if we should retain this pattern since it keeps breaking our brains? Could we refactor the validate method to use WP_Errors instead of exceptions that we need to catch and handle manually? Maybe outside the scope of this PR, but I feel this pattern has anecdotally bitten us more than helped us.

ah, and about this. You're probably right. We should use a pattern we use everywhere else... but I personally don't want to engage in this refactor right now :)

dkoo (approved):
This is working for me after the refresh. Thanks, @leogermani!

PR #1699, opened 2024-11-11
feat(cli): add command to synchronize premium lists

adekbadek:
### All Submissions:

Adds a CLI command to synchronize premium newsletter subscribers. 

1. Ensure you have a premium newsletter list:
    1. Create a membership plan with "Restrict Content" -> choose "Subscription Lists" -> pick a list
3. Create two memberships for two different users, with this plan 
4. Cancel one of the memberships 
5. In the ESP, break things: unsubscribe the user with the active membership, and subscribe the user with the inactive membership
6. Run `wp newspack-newsletters sync-membership-tied-subscribers --live --verbose`, and observe it realigns the status in the ESP with the memberships status 

### Other information:

---
  - https://app.asana.com/0/0/1208660941160628

### DIFF
--- a/includes/cli/class-sync-premium-subscribers.php
+++ b/includes/cli/class-sync-premium-subscribers.php
@@ -155,7 +155,7 @@ class Sync_Premium_Subscribers {
 									$email,
 									$lists_to_add,
 									$lists_to_remove,
-									'Updating contact when running the CLI sync script.'
+									'Updating contact when running the sync-premium-subscribers CLI sync script.'
 								);
 							}
 
--- a/includes/cli/class-sync-premium-subscribers.php
+++ b/includes/cli/class-sync-premium-subscribers.php
@@ -1,0 +1,176 @@
+<?php
+/**
+ * Newspack Newsletters Premium Subscribers CLI.
+ *
+ * @package Newspack
+ */
+
+namespace Newspack_Newsletters\CLI;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Manages Settings page.
+ */
+class Sync_Premium_Subscribers {
+	/**
+	 * Initialize the class
+	 *
+	 * @codeCoverageIgnore
+	 */
+	public static function init() {
+		add_action( 'init', [ __CLASS__, 'initialize_cli_commands' ] );
+	}
+
+	/**
+	 * Initialize CLI commands.
+	 *
+	 * @codeCoverageIgnore
+	 */
+	public static function initialize_cli_commands() {
+		if ( ! defined( 'WP_CLI' ) || ! WP_CLI ) {
+			return;
+		}
+
+		\WP_CLI::add_command(
+			'newspack-newsletters sync-premium-subscribers',
+			[ __CLASS__, 'cli_sync_premium_subscribers' ],
+			[
+				'shortdesc' => 'Synchronizes the premium newsletter lists with the memberships.',
+				'synopsis'  => [],
+			]
+		);
+	}
+
+	/**
+	 * Can a memberships be considered active?
+	 *
+	 * @param \WC_Memberships_User_Membership $user_membership User membership.
+	 */
+	public static function is_membership_active( $user_membership ): bool {
+		$active_statuses = wc_memberships()->get_user_memberships_instance()->get_active_access_membership_statuses();
+		return in_array( $user_membership->get_status(), $active_statuses );
+	}
+
+	/**
+	 * CLI handler for premium newsletter lists synchronization.
+	 *
+	 * @param array $args Indexed array of args.
+	 * @param array $assoc_args Associative array of args.
+	 * @return void
+	 *
+	 * ## OPTIONS
+	 *
+	 * [--live]
+	 * : Run the command in live mode, updating the data.
+	 *
+	 * [--verbose]
+	 * : More output.
+	 *
+	 * ## EXAMPLES
+	 *
+	 *     wp newspack-newsletters sync-premium-subscribers
+	 */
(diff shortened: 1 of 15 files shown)


leogermani (changes requested):
This mostly works! 

One thing that doesn't work is: if the member does not exits in the ESP, the sync fails. It first needs to be added to the ESP. (I have a suggestion down below that might help solving this).

Other feedback:

1. Link to where the main feature is

I wonder how we can make sure that we make it clear that this CLI command is related to [this](https://github.com/Automattic/newspack-newsletters/blob/39e1aa55332d30cee525fbee1006a55ff3d53db0/includes/plugins/class-woocommerce-memberships.php#L21). Each thing live in a different folder and we don't mention "Premium" anywhere in the original file to identify this feature.

Maybe both files living in the same folder? Maybe adding a command at the beginning of both of them?

My goal is that any developer that arrives in any of these files will know that the other exists.

2. extend the command description (and highlight the caveat it has)

Add a longer description, explaining what the command does. Also, there is one caveat:

If a reader went to "My Account" and intentionally unsubscribed to a Premium list, they will be re-subscribed if we run this command.

I thought we could do something similar to[ what we do here](https://github.com/Automattic/newspack-newsletters/blob/39e1aa55332d30cee525fbee1006a55ff3d53db0/includes/plugins/class-woocommerce-memberships.php#L325), but it's not that simple... Also, I think they can go to their Mailchimp profile and unsubscribe from there, and we wouldn't know... Not sure if there is a reliable way to avoid that. Maybe just highlight that this is a side effect of this script.

3. Maybe check the user status before updating it

WDYT of getting the user data and checking if their subscription is wrong and needs to be fixed, instead of simply updating all the users as if they all needed to be fixed.

This would help solve the issue I found in the beginning of this comment, because we would know if they user does not exist in the ESP and we could then create it.

Also, the `dry-run`  command would be much more powerful, actually giving us a report of how many users are "wrong"

includes/cli/class-sync-premium-subscribers.php
```diff
@@ -0,0 +1,176 @@
...
+									$live ? '      - Updating the contact in the ESP…' : '      - Would update the contact in the ESP.'
+								);
+							}
+
+							if ( $live ) {
+								$result = \Newspack_Newsletters_Contacts::add_and_remove_lists(
+									$email,
+									$lists_to_add,
+									$lists_to_remove,
+									'Updating contact when running the CLI sync script.'
```
leogermani:
```suggestion
									'Updating contact when running the sync-premium-subscribers CLI sync script.'
```

leogermani:
ah, one last NIT

The "Warning" and "Success" messages don't respect the indentation, making it hard to read sometimes and know which block the message is related to. Maybe just add one additional line break between blocks

Captura de tela de 2024-11-11 11-48-49

adekbadek:
> 0. One thing that doesn't work is: if the member does not exits in the ESP, the sync fails. It first needs to be added to the ESP. 

Good catch, added in eedd6471d52bc1d60aecc563e3754aa9aae514ed

> 1. Link to where the main feature is […] we don't mention "Premium" anywhere in the original file to identify this feature

That's right, I was inspired by the naming in #1255. The "Premium" is mentioned in `includes/plugins/class-woocommerce-memberships.php`, though.
Updated the naming to "membership-tied" instead of "premium" in 71d3f01b09b9314a530944b08994c2a3e0c76438 and the file structure in f51c54539b94d392e716521c70a5faf5e2c4cdd2. That should make more sense, WDYT?

> 2. extend the command description (and highlight the caveat it has)

Done in 670fa3615d9e8b083d3b2c211969de917e2dcd68. I also don't see a way to reliably and across all ESPs know if the user was unsubscribed or never subscribed, so I went ahead with the suggestion and added it to the  

> 3. Maybe check the user status before updating it

Good call, added in efdd76bcf646b2c0ef2ddf6f5799da2ddc174f9c

> 3. WDYT of getting the user data and checking if their subscription is wrong and needs to be fixed

With MC, what we have stored on site is a "remote ID" like `group-3d0ab673dd-a35c2fe658`. This is a different format than what's coming from MC API, it would have to be parsed. I don't think it's worth the added complexity, considering this is not code that's supposed to be run on a regular basis.

leogermani:
I had to add https://github.com/Automattic/newspack-newsletters/pull/1699/commits/a8877a0e041bf3d75e59e86297d686832505edbd for the contact addition to work. It wasn't working for me

> With MC, what we have stored on site is a "remote ID" like group-3d0ab673dd-a35c2fe658. This is a different format than what's coming from MC API, it would have to be parsed. I don't think it's worth the added complexity, considering this is not code that's supposed to be run on a regular basis.

I took the liberty of pushing https://github.com/Automattic/newspack-newsletters/pull/1699/commits/db9f06134f90d41920fbb364918139d94830bb4f

Just because it's pretty simple now with the Subscription_List objects. I think this is super valuable because it also turns this into a good diagnosis script.

The change from `remote_id` to `public_id` is just to properly support Local lists. For remote lists both ids are the same.

(This difference in the implementation of Local lists is some legacy we want to get rid at some point...)

adekbadek:
@leogermani – thank you for these additions, that makes sense 👍

leogermani:
> @leogermani – thank you for these additions, that makes sense 👍 

One thing that occurred to me is that if we looped through users first, and then by lists in the membership, we would save ourselves a lot of requests to the esp In case there are more than one list behind a membership. It would make a big difference for large sites

leogermani (approved):

adekbadek:
Yeah, that optimization would make sense. Let's revisit it if we run into issues with too many ESP API requests.

PR #735, opened 2022-02-09
feat: remove subject input (title) from the sidebar

thomasguillot:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Inspired by pcD9cT-I4-p2, this PR completely removes the "Subject" field from the sidebar and instead reintroduces the standard Post Title in the editor, just slightly restyled.

newsletter-editor

### How to test the changes in this Pull Request:

1. Switch to this branch
2. Edit the subject
3. Send newsletter and check if subject is correct

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

PR #1701, opened 2024-11-11
fix: remove behavior to hide post title in newsletter editor

dkoo:
### All Submissions:

This removes the behavior which attempts to hide the editor post title element in newsletter CPTs when the newsletter is NOT marked to be a "public post" after sending. The hiding behavior is inconsistent between WP 6.6 vs. WP 6.7 (it's actually been broken for while and few people noticed), and after some internal discussion we decided that the intended behavior that hides the core editor's title element in favor of the "Subject" field in the sidebar was potentially more confusing than helpful. This PR retains both the core post title element (keeping the UX more consistent with all other WP post editor screens) and sidebar subject field, which may be a little redundant but the most cautious approach to avoid disrupting editorial workflows.

Tagging @thomasguillot on this because I see some activity on this in some [past](https://github.com/Automattic/newspack-newsletters/pull/773) [threads](https://github.com/Automattic/newspack-newsletters/pull/735), in case you want to revisit this UX.

1. Edit a newsletter post and confirm that the post title is always visible at the top of the post editor screen, as with all other WP post editor screens. There should be no other functional changes.

### Other information:

### DIFF
--- a/src/newsletter-editor/editor/index.js
+++ b/src/newsletter-editor/editor/index.js
@@ -49,7 +49,6 @@ const Editor = compose( [
 			),
 			meta,
 			sent,
-			isPublic: meta.is_public,
 			newsletterSendErrors: meta.newsletter_send_errors,
 			isCustomFieldsMetaBoxActive: getAllMetaBoxes().some( box => box.id === 'postcustom' ),
 		};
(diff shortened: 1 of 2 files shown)


leogermani (approved):
just works

PR #1729, opened 2024-12-18
fix: avoid creating merge fields if we can't get merge fields

dkoo:
### All Submissions:

When preparing a contact's data sync, we compare the contact's meta fields with the existing merge fields in the audience so that we can use the fields that exist and create the ones that don't yet exist. However, if we can't get info about the existing merge fields via the Mailchimp API for any reason, we should stop the sync to avoid potentially creating duplicate merge fields.

This PR also uses the `validate` method to catch any potential error-like responses from the Mailchimp API.

Coupled with https://github.com/Automattic/newspack-plugin/pull/3639, this PR will also schedule a retry sync after five minutes if we fail to fetch merge fields, up to a maximum of five times.

Addresses: `1200550061930446-as-1208959720314117`

1. Smoke test syncing new and existing contacts with Mailchimp and confirm that contact metadata gets synced correctly to existing merge fields, and that merge fields that don't yet exist get created.

NOTE: leaving the below in place for the record, but it's no longer part of this PR.

<details>
<summary>Click to see additional testing steps (no longer part of this PR)</summary>

2. Temporarily add a line `$response = [];` [here](https://github.com/Automattic/newspack-newsletters/pull/1729/files#diff-961258e170b9bfe6e81621bab727a9a65947cd67cc016bf5ce1926dd8c78d77cR1586) to simulate a non-error response from the Mailchimp API that nonetheless fails to return merge field data.
3. Trigger a contact sync to Mailchimp via any means.
4. Confirm that the sync now fails with an error, without creating any new merge fields.
5. Using WP CLI, run `wp cron event list` and confirm there's a non-recurring `newspack_scheduled_esp_sync` job scheduled for 5 minutes in the future.
6. Wait 5 minutes or run `wp cron event run newspack_scheduled_esp_sync` to trigger the job immediately. Repeat this 5 times.
7. After 5 failed syncs, confirm that a `Failed to sync contact <email address> after 5 retries` is logged, and that there's no more `newspack_scheduled_esp_sync` job scheduled.
8. Remove the temporary error line added in step 2 and resync the contact.
9. Using WP CLI, run `Newspack\Reader_Activation\Sync::schedule_sync( $user_id, 'Testing scheduled sync', 30 );` and confirm that the scheduled sync happens successfully after 30 seconds.

</details>

### Other information:

### DIFF
--- a/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php
+++ b/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php
@@ -1574,13 +1574,7 @@ final class Newspack_Newsletters_Mailchimp extends \Newspack_Newsletters_Service
 
 		// Get and match existing merge fields.
 		try {
-			$existing_fields = $mc->get(
-				"lists/$audience_id/merge-fields",
-				[
-					'count' => 1000,
-				],
-				60
-			)['merge_fields'];
+			$existing_fields = Newspack_Newsletters_Mailchimp_Cached_Data::get_merge_fields( $audience_id );
 		} catch ( \Exception $e ) {
 			do_action(
 				'newspack_log',
--- a/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php
+++ b/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php
@@ -1587,14 +1587,14 @@ final class Newspack_Newsletters_Mailchimp extends \Newspack_Newsletters_Service
 
 			// If we didn't get ANY merge fields in the response, something is wrong. Bail to avoid creating duplicate fields.
 			if ( empty( $response['merge_fields'] ) ) {
-				throw new Exception( esc_html__( 'Could not fetch merge fields', 'newspack-newsletters' ) );
+				throw new Exception( esc_html__( 'Response returned no merge_fields', 'newspack-newsletters' ) );
 			}
 			$existing_fields = $response['merge_fields'];
 		} catch ( \Exception $e ) {
 			do_action(
 				'newspack_log',
 				'newspack_mailchimp_prepare_merge_fields',
-				__( 'Error getting merge fields', 'newspack-newsletters' ),
+				'Error getting merge fields',
 				[
 					'type'       => 'error',
 					'data'       => [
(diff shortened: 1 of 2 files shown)


leogermani:
> This PR also uses the validate method to catch any potential error-like responses from the Mailchimp API.

Good catch! That's why this error was never being logged!

I couldn't get it all to work, here are a few things I saw:

* When the error is triggered, it doesn't stop the sync from happening, which results in the contact missing the merge fields
* When I run `wp cron event run newspack_scheduled_esp_sync`, it succeeds, but I don't see the actual sync going through
* Another sync is not scheduled (I'm running the two branches as expected)

dkoo:
@leogermani I think 345229b should stop the sync by throwing an exception if we encounter a merge fields error. Note that since we're using the `newspack_log` action, the errors won't be logged directly to debug.log, but to the Mailchimp log file in your server's `/tmp` directory.

dkoo:
> When I run `wp cron event run newspack_scheduled_esp_sync`, it succeeds, but I don't see the actual sync going through

This might be the case if you still have the temp error being triggered by the `$response = [];` line. If it's still there then the scheduled syncs will continue failing with an error (logged in the local `/tmp` logfile, not debug.log). On my site once I remove that line the next scheduled sync will succeed.

leogermani:
> > When I run `wp cron event run newspack_scheduled_esp_sync`, it succeeds, but I don't see the actual sync going through
> 
> This might be the case if you still have the temp error being triggered by the `$response = [];` line. If it's still there then the scheduled syncs will continue failing with an error (logged in the local `/tmp` logfile, not debug.log). On my site once I remove that line the next scheduled sync will succeed.

Right, it's expected to fail, but also to schedule a new retry until it reaches 5 attempts. It didn't schedule the second retry

includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php
```diff
@@ -1793,24 +1806,7 @@ public function add_contact( $contact, $list_id = false, $tags = [], $interests
...
-				return new WP_Error(
-					'newspack_newsletters_mailchimp_add_contact_failed',
-					sprintf(
-						/* translators: %s: Mailchimp error message */
-						__( 'Failed to add contact to list. %s', 'newspack-newsletters' ),
-						isset( $result['detail'] ) ? $result['detail'] : ''
-					)
-				);
-			}
+			$result      = $this->validate( $mc->put( "lists/$list_id/members/$member_hash", $update_payload ) );
```
dkoo, line 1813: @leogermani I think this change is still safe because it should result in the same behavior and error code—the removed code was mainly duplicating some of the checks in the `validate()` method. Happy to revert this too if you prefer.

dkoo:
@leogermani I've scaled back the changes in this PR to include only the additional validation and error logging, so this should be ready for another review.

leogermani (approved):

PR #1732, opened 2025-01-07
fix: throw error if we fail to fetch merge fields

dkoo:
### All Submissions:

Reinstates a reverted change from #1729. Throws an error if the request to fetch existing merge fields from Mailchimp fails for any reason. This should help prevent duplicate merge fields being created in this scenario.

1. Temporarily add a line `$response = [];` [here](https://github.com/Automattic/newspack-newsletters/pull/1732/files#diff-961258e170b9bfe6e81621bab727a9a65947cd67cc016bf5ce1926dd8c78d77cR1587) to simulate a non-error response from the Mailchimp API that nonetheless fails to return merge field data.
2. Trigger a contact sync to Mailchimp via any means.
3. Confirm that the sync now fails with an error, without creating any new merge fields.

### Other information:

### DIFF
--- a/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php
+++ b/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php
@@ -1605,6 +1605,10 @@ final class Newspack_Newsletters_Mailchimp extends \Newspack_Newsletters_Service
 					'file'       => 'newspack_mailchimp',
 				]
 			);
+			return new \WP_Error(
+				'newspack_mailchimp_prepare_merge_fields',
+				$e->getMessage()
+			);
 		}
 
 		if ( empty( $existing_fields ) ) {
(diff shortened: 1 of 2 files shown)


includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php
```diff
@@ -1793,8 +1797,8 @@ public function add_contact( $contact, $list_id = false, $tags = [], $interests
 				 * @param array $merge_fields The merge fields payload to pass to the Mailchimp API.
 				 */
 				$merge_fields = apply_filters( 'newspack_mailchimp_merge_fields', $this->prepare_merge_fields( $list_id, $contact ) );
-				if ( ! empty( $merge_fields ) ) {
-					$update_payload['merge_fields'] = $merge_fields;
+				if ( is_wp_error( $merge_fields ) ) {
+					throw new Exception( $merge_fields->get_error_message() );
 				}
```
leogermani, line 1802: dont we need an else clause to update the `$update_payload` var in case there are merge fields?

dkoo: Whoops, you're right—added that back in d4a2b86

leogermani (approved):
I did some testing to make sure merge fields wouldn't get cleared if `$merge_fields` was passed with an empty array. We're safe

PR #1745, opened 2025-01-23
feat: better reader-facing error messages

dkoo:
### All Submissions:

Implements a filterable generic reader-facing error message to show to readers in favor of just passing on the raw errors from the ESP's API. The raw errors and any details we can glean from them are logged for debugging purposes using `newspack_log`.

Only affects the `add_contact` methods for each ESP for now, which is the main context in which readers might encounter API errors when signing up.

1. Check out this PR.
2. Using the Newsletter Subscription Form block, the Reader Registration block, ~the auth modal, and a post-checkout newsletter signup modal~ (EDIT: apparently these don't currently handle errors, so let's skip for now), sign up an email address for a newsletter list and ensure it will result in an error. You can force an error by :
  - Using an email you know can't be subscribed to the list—for example, for Mailchimp, signing up using an email address that has previously unsubscribed from the same audience
  - Or, by throwing an exception or WP_Error in the active ESP's `add_contact` method (whether to throw an exception or an error depends on the specific ESP you're using).

3. Confirm that in the reader-facing front-end, you see the generic error message:

Screenshot 2025-01-22 at 4 57 06 PM

4. Also confirm that the "real" error message is logged via `newspack_log`.
5. Define `NEWSPACK_NEWSLETTERS_DEBUG_MODE` as true in wp-config.php and repeat the signup attempt, but this time confirm that the "real" error message is shown instead of the generic one.
6. Optionally, test the new `newspack_newsletters_add_contact_reader_error_message` filter by adding your own custom error message via `add_filter()`.

### Other information:

---
  - https://app.asana.com/0/0/1208879786064023

### DIFF
--- a/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php
+++ b/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php
@@ -1840,16 +1840,12 @@ final class Newspack_Newsletters_Mailchimp extends \Newspack_Newsletters_Service
 			 * @param string $reader_error The default error message.
 			 * @param string $email_address The email address that was attempted to be subscribed.
 			 * @param string $list_id The Mailchimp list ID that the email address was attempted to be subscribed to.
-			 * @param array $tags The tags that were attempted to be added to the email address.
-			 * @param array $interests The interests that were attempted to be added to the email address.
 			 */
 			$reader_error = apply_filters(
-				'newspack_newsletters_mailchimp_add_contact_reader_error_message',
+				'newspack_newsletters_add_contact_reader_error_message',
 				__( "Sorry, this email cannot be subscribed to this newsletter. Please contact support with the email list you were trying to subscribe to and we'll add you to the list.", 'newspack-newsletters' ),
 				$email_address,
-				$list_id,
-				$tags,
-				$interests
+				$list_id
 			);
 			$result      = $this->validate( $mc->put( "lists/$list_id/members/$member_hash", $update_payload ), $reader_error, $email_address );
 		} catch ( \Exception $e ) {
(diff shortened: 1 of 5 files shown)


includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php
```diff
@@ -1815,7 +1833,21 @@ public function add_contact( $contact, $list_id = false, $tags = [], $interests
...
+			 * @param string $reader_error The default error message.
+			 * @param string $email_address The email address that was attempted to be subscribed.
+			 * @param string $list_id The Mailchimp list ID that the email address was attempted to be subscribed to.
+			 */
+			$reader_error = apply_filters(
+				'newspack_newsletters_add_contact_reader_error_message',
+				__( "Sorry, this email cannot be subscribed to this newsletter. Please contact support with the email list you were trying to subscribe to and we'll add you to the list.", 'newspack-newsletters' ),
+				$email_address,
+				$list_id
+			);
```
chickenn00dle: Nonblocking: This bit is repeated a few times in the code. Maybe we can move this into a helper method in the `Newspack_Newsletters_Service_Provider` class? (something like `get_add_contact_reader_error_message`)

chickenn00dle (approved):
Works as described! Left a non-blocking comment, but otherwise this is good to go 👍

includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php
dkoo: Great idea! I'll add this before merging and ping you for a re-review.

dkoo: @chickenn00dle 16edd2f implements a `get_add_contact_reader_error_message` method in the parent `Newspack_Newsletters_Service_Provider` class. It also consolidates additional params for the filter into a single arg, for more flexibility.

chickenn00dle (approved):
Looks good! Thanks @dkoo

PR #1686, opened 2024-10-30
fix: dont html encode ampersands in subject

leogermani:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/trunk/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Fixes a small issue where ampersands are HTML encoded in the Subject.

Not sure if there's a better way to do it. But it's weird because the only character that is being HTML encoded is the ampersand, so loading a library to html_decode everything looked unnecessary.

### How to test the changes in this Pull Request:

1. In `trunk` edit a newsletters
2. add a `&` to the title and see that `$amp;` is added to the Subject input in the Newsletter sidebar
3. Checkout this branch and do the same
4. See that `&` is preserved as is

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

PR #1748, opened 2025-01-29
fix: avoid moving cursor when editing Subject input field

dkoo:
### All Submissions:

Fixes a small bug with the Subject input field from #1686. Because WP HTML-encodes the title string, we need to convert the encoded string back to plain text before setting the value of the subject field, to avoid double-encoding the title. But because the input value directly changes the title, which triggers the conversion back to plain text, this results in the entire string being replaced in the component state upon every keypress. This PR fixes this issue by only updating the actual post title when the input field is blurred (unfocused).

1. On `trunk`, edit a newsletter and edit the Subject field in the newsletter sidebar. Place your cursor in the middle of the string and edit, and observe that upon every keypress the cursor jumps to the end of the string.
2. Check out this branch and repeat. This time, confirm that the cursor does not reposition itself while typing in the input field, and that after you click or tab out of the input field, the post title is updated to match the input.
3. Smoke test editing the subject by changing the title in the block editor content area (the subject field should still update automatically to match).
4. Smoke test saving the post with HTML characters (such as ampersands) in the title and refresh the editor. Confirm that the title saved to the DB is HTML-encoded (use WP CLI: `wp post get <post ID> --fields=post_title`, but that it still renders as plain text in both the big content-area title and in the Subject input field.

### Other information:

---
  - https://app.asana.com/0/0/1209079466938940

### DIFF
--- a/src/newsletter-editor/sidebar/index.js
+++ b/src/newsletter-editor/sidebar/index.js
@@ -40,7 +40,7 @@ const Sidebar = ( {
 	stringifiedCampaignDefaults,
 	postId,
 } ) => {
-	const [ plainTextTitle, setPlainTextTitle ] = useState( title );
+	const [ plainTextTitle, setPlainTextTitle ] = useState( null );
 	const isRetrieving = useIsRetrieving();
 	const newsletterData = useNewsletterData();
 	const newsletterDataError = useNewsletterDataError();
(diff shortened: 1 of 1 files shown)


leogermani (approved):
Just works

laurelfulford:
Looks like I overlapped in testing with @leogermani 😆 

The only thing I spotted is that if you edit the post in the Title Block last, HTML characters like `&` are saved like `&amp;` when you run `wp post get <post ID> --fields=post_title`, but if you edit them in the Subject field last, they're saved like `&`. They still show as `&` on the front-end after exiting the editor, so I don't know if this is actually an issue.

dkoo:
@laurelfulford good catch! It's probably safest to preserve the entity encoding behavior of standard WP posts, so e10d09c fixes this—now, the title string saved to the DB is always HTML-encoded regardless of which field you use to edit it before saving.

dkoo:
@laurelfulford this could use another review when you have a moment!

laurelfulford (approved):
Sorry @dkoo, I missed the last update. This looks good now! Editing in either location saves the `&` as `&amp;` 🚢

dkoo:
Thanks, @laurelfulford!

PR #1754, opened 2025-02-11
fix: allow for context-specific reader-facing error messsages

dkoo:
### All Submissions:

Builds on filters added in #1745 by allowing reader-facing error messages to be altered depending on the context of the original error thrown by the ESP's API. 

This PR adds a special error message for the Mailchimp error that occurs when a contact attempts to resubscribe to a list they've previously unsubscribed from. In all other cases, the error shown will be a generic error: `Sorry, an error has occurred. Please try again later or contact us for support.`

Also standardizes the error code used across all ESP's when `add_contact` results in an error to: `newspack_{esp}_api_error`

#### set up a contact for the compliance state

1. Check out this branch.
2. With Mailchimp enabled, sign up as a reader with a real email address for a test newsletter list/group/tag via any means on the site.
3. Send a live newsletter campaign to the list/group/tag that the reader will receive.
4. Once you receive the email in the reader's inbox, click the "Unsubscribe" link at the bottom and confirm.
5. In a new session, attempt to resubscribe to the same newsletter lists using the same reader email.
6. Confirm the following error is shown:

Screenshot 2025-02-11 at 4 49 10 PM

7. Temporarily force an error to occur in the first line of the Mailchimp class `validate()` method:

```diff
	public function validate( $result, $preferred_error = null, $payload = [] ) {
+		$result['status'] = 400;
+		$result['title']  = 'Test error';
+		$result['detail'] = 'Test error message';
		$default_error = $this->get_reader_error_message( $payload, $result );
```

8. In a new session, subscribe to a newsletter list with a fresh email address and confirm you get the generic error message:

Screenshot 2025-02-11 at 4 55 17 PM

### Other information:

### DIFF
--- a/includes/service-providers/constant_contact/class-newspack-newsletters-constant-contact.php
+++ b/includes/service-providers/constant_contact/class-newspack-newsletters-constant-contact.php
@@ -1206,7 +1206,7 @@ final class Newspack_Newsletters_Constant_Contact extends \Newspack_Newsletters_
 			// Log the error with any details returned from the API.
 			do_action(
 				'newspack_log',
-				'newspack_constant_contact_add_contact_failed',
+				'newspack_' . $this->service . '_api_error',
 				'Error adding contact to Constant Contact.',
 				[
 					'type'       => 'error',
@@ -1219,11 +1219,12 @@ final class Newspack_Newsletters_Constant_Contact extends \Newspack_Newsletters_
 				]
 			);
 
-			$reader_error = $this->get_add_contact_reader_error_message(
+			$reader_error = $this->get_reader_error_message(
 				[
 					'email'   => $contact['email'],
 					'list_id' => $list_id,
-				]
+				],
+				$result
 			);
 
 			return Newspack_Newsletters::debug_mode() ? $result : new \WP_Error( 'newspack_constant_contact_add_contact_failed', $reader_error );
--- a/includes/service-providers/constant_contact/class-newspack-newsletters-constant-contact.php
+++ b/includes/service-providers/constant_contact/class-newspack-newsletters-constant-contact.php
@@ -1201,33 +1201,8 @@ final class Newspack_Newsletters_Constant_Contact extends \Newspack_Newsletters_
 		}
 
 		$result = $cc->upsert_contact( $contact['email'], $data );
-		if ( is_wp_error( $result ) || empty( $result ) ) {
-
-			// Log the error with any details returned from the API.
-			do_action(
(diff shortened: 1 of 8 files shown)


includes/service-providers/constant_contact/class-newspack-newsletters-constant-contact.php
```diff
@@ -1206,7 +1206,7 @@ public function add_contact( $contact, $list_id = false ) {
 			// Log the error with any details returned from the API.
 			do_action(
 				'newspack_log',
-				'newspack_constant_contact_add_contact_failed',
+				'newspack_' . $this->service . '_api_error',
```
leogermani: this logging repeats in every esp. Is it possible to move this up in the abstraction and automatically do it for all of them?

Also, wdyt of adding all errors into the same bucket and add the provider as a piece of data in the params. It's what we do with all the other logs

dkoo: 4acbef7 moves the reader-facing error handling and logging to the higher-level `Newspack_Newsletters_Contacts::upsert()`. It simplifies things a bit by repurposing the existing log with the `newspack_esp_sync_upsert_contact` error code to log the error message from the ESP's API instead of the reader-facing error message, so there's now no longer a separate log for each. There's probably no reason to log the reader-facing error message for debugging purposes.

The testing instructions remain the same, but you can also verify that the logged output shows the original error message from the API instead of the reader-facing error message:

Screenshot 2025-02-18 at 12 35 17 PM

leogermani (commented):
@dkoo Looking good. I think things are better organized now and all tests well.

My only doubt now is that looking at the message we chose for the Mailchimp Compliance error, I don't like it very much:

"Sorry, this email cannot be subscribed to this newsletter. Please contact support with the email list you were trying to subscribe to and we'll add you to the list."

I think we could improve this a little, to be more generic for any publisher. Here are some AI suggestions:

"We noticed this email was previously unsubscribed from our newsletter. For your privacy protection, please contact our support team to reactivate your subscription."
"This email address has previously opted out of our communications. To resubscribe, please contact our support team at [support email/number]."
"Your email address requires additional verification before resubscribing. Please reach out to our support team for assistance."

I like the first one. WDYT?

dkoo:
After some internal discussion, we arrived at this more concise wording for the error message in question:

> We'll need to subscribe this email address manually. Please contact our support team.

Updated in a904224.

leogermani (approved):

PR #1758, opened 2025-02-14
feat: starting reorganizing esp abstraction

leogermani:
### All Submissions:

Reorganizes the ESP abstractions to make it easier to understand and build a new ESP.

1. Smoke test as much as you can

### Other information:


leogermani:
This is a first pass on the idea of making our abstraction easier to parse.

In the end, we might not need this intermediary class that has only two methods. The only reason it exists is because they offer a default implementation that providers may or may not override... All the rest is in the interface. 

If someone wants to implement a new provider, just look at the interface.

Next:

* Consider if we need the base and intermediary class, or just move the default implementation to the base class and have just one class...
* Expand the doc blocks in the interface to give more context and information
* add a readme giving an overview... for example, explaining what are local lists
* Rename some methods for better readability. For example: `send` to `send_campaign` and `list` to `set_campaign_list`

Question:

* I don't see the `sender` method being called anywhere. Where is it invoked?

cc @miguelpeixe @dkoo

dkoo:
One argument I can see in favor the intermediate class is if we make this default implementation the class for the "manual" ESP setting. Then have the fully-supported ESP classes extend from that. This would require some refactoring since there's currently no class for the "manual" ESP.

leogermani:
> One argument I can see in favor the intermediate class is if we make this default implementation the class for the "manual" ESP setting. Then have the fully-supported ESP classes extend from that. This would require some refactoring since there's currently no class for the "manual" ESP.

Yes, that would be interesting, but I don't want to go that far...

I did a bit more work on this. Left just one parent class. In order to create a new provider, you just need to look at the interfaces and implement the methods.

Still need to fix tests and rename some methods

leogermani:
Ok, this is ready for a review. I didn't rename some methods I'd like, but I think there are enough changes to make things a bit more organized

dkoo:
> I don't see the `sender` method being called anywhere. Where is it invoked?

You're right, this can be removed as syncing sender name + email is now handled using the `senderName` and `senderEmail` meta values. [Here's an example of how it's done in Mailchimp](https://github.com/Automattic/newspack-newsletters/blob/feat/improve-esp-abstraction/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php#L1013-L1026).

dkoo (approved):
I tested subscription lists and creating/editing new newsletters with AC, CC, and MC and saw no issues. 

Preapproving because there are no blocking issues, but feel free to remove all `sender()` methods before merging as I tested without those methods and saw no issues.

Also, I think this is unrelated to this PR, but there's a PHP warning when saving CC newsletters: `PHP Deprecated:  Implicit conversion from float 593508.0051422119 to int loses precision in /newspack-repos/newspack-newsletters/includes/service-providers/constant_contact/class-newspack-newsletters-constant-contact-sdk.php on line 110`

leogermani:
Thanks!

Yes, I saw that warning too but that's unrelated.

Let's keep things safe and merge this as is... We can get rid of `sender` and maybe rename a few methods I wanted to rename later

PR #1763, opened 2025-02-24
fix(subscribe): optionally include metadata on subscribe

dkoo:
### All Submissions:

With https://github.com/Automattic/newspack-plugin/pull/3779, adds an optional param to allow contact metadata fields to be synced upon a `subscribe()` call. The cases we'd want this to happen are upon newsletter signup + account registration via Newsletter Subscription Form and Reader Registration blocks, where there may be account registration, referrer, and UTM meta fields captured during the signup, but not synced via the Newspack Plugin's `Sync` classes.

Follow testing instructions in https://github.com/Automattic/newspack-plugin/pull/3779.

### Other information:


includes/class-newspack-newsletters-contacts.php
```diff
@@ -38,10 +38,11 @@ class Newspack_Newsletters_Contacts {
 	 * @param string[]|false $lists   Array of list IDs to subscribe the contact to. If empty or false, contact will be created but not subscribed to any lists.
 	 * @param bool           $async   Whether to add the contact asynchronously. Default is false.
 	 * @param string         $context Context of the update for logging purposes.
+	 * @param bool           $include_metadata Whether to include metadata when adding a contact, such as when subscribing via the Newsletter Subscription Form block. Default is false.
 	 *
 	 * @return array|WP_Error|true Contact data if it was added, or error otherwise. True if async.
 	 */
-	public static function subscribe( $contact, $lists = false, $async = false, $context = 'Subscribe contact' ) {
+	public static function subscribe( $contact, $lists = false, $async = false, $context = 'Subscribe contact', $include_metadata = false ) {
```
leogermani: Do you remember which case we were handling when we decided to remove the metadata from here?

If we are setting this param to `true` everywhere we call this method there will likely be a regression... or is there anywhere this will still be called as `false`?

dkoo: I think we were accounting for when the reader updates their newsletter subscriptions via the My Account > Newsletters UI. In that scenario, we shouldn't be sending any additional metadata as part of the subscribe request.

This PR doesn't set the param to `true` everywhere it's called—it only sets it to `true` when calling `Newspack_Newsletters_Contacts::subscribe()` following a form submission via the Newsletter Subscription Form block or Registration block. All other instances of `Newspack_Newsletters_Contacts::subscribe()` should be unchanged.

leogermani: I don't feel confident... It seems we are circling back to something we intentionally tried to avoid.

WDYT of simply adding the other metadata that are part of newsletter subscription to the list of accepted metadata in the subscribe method. Currently it reads

```
$accepted_metadata = [ 'status', 'name' ];
```

Makes sense?

@miguelpeixe , if you can help us [remember the rationale](https://github.com/Automattic/newspack-plugin/pull/3415#pullrequestreview-2296678044) behind our choices here as well it would be awesome

dkoo: Hmm, but for these particular cases we'd want any valid metadata fields to be synced. Maybe we should be calling `upsert()` instead of `subscribe()` for signup requests from the blocks?

dkoo: Looking at this a bit more, it seems like the only real differences between `subscribe` and `upsert` methods are that the former allows for async execution, and it strips all metadata except for `status` and `name` keys. After that, `subscribe` passes its data through to the `upsert` method anyway.

dkoo: Wracking my memory for the reason behind these decisions...I think we wanted to retain the async option for certain newsletter signup flows so we could show an optimistic UI to make it feel faster, and show any errors to the user at a later point in their My Account dashboard. These are all of the flows where the async `subscribe` method is currently used:

- Registration block
- Newsletter Subscription Form block
- RAS auth modal registration
- Post-modal-checkout newsletter signup modal
- Newsletter signup changes via the My Account dashboard
- A CLI command to resync premium newsletter memberships to premium newsletter signups (`wp newspack-newsletters sync-membership-tied-subscribers`)

I think for all of these except for the CLI command and the My Account page, we actually want all of the contact metadata to be synced to the ESP.

leogermani: > it seems like the only real differences between subscribe and upsert methods are that the former allows for async execution, and it strips all metadata except for status and name keys. 

A key difference is that it triggers a different action, that will trigger a [different data event](https://github.com/Automattic/newspack-plugin/blob/46869f774f62034aff9943bb4843fc5ef73f2bb6/includes/data-events/listeners.php#L100)

idk, maybe we don't need to strip metadata out? I can't remember why this specific piece is there... would love to know if @miguelpeixe remembers

miguelpeixe: IIRC, the rationale was mostly regarding the separation of concerns and better control of the data flow. `subscribe()` is a user action to subscribe to list(s).

We're mixing names here. `upsert()` is the lower-level method, used by both `subscribe()` and `sync()`.

`sync()` is our async/background strategy for metadata updates.

I don't have a strong opinion on whether we should allow the metadata to be sent via `subscribe()` or trigger a `sync()` after the `subscribe()`. I lean more to let `sync()` be the sole owner of that task.

I'm not a fan of restoring the normalize contact data filter as proposed in https://github.com/Automattic/newspack-plugin/pull/3779. It creates room for metadata coming from everywhere and debugging more difficult, as it was before the refactors.

dkoo: Looks like we're already firing a `newspack_newsletters_contact_subscribed` action and `newsletter_subscribed` with the unfiltered metadata after subscribing and syncing on the data event, but not including the metadata from the event payload. So if we include the metadata from this data event in the sync, then the changes in this PR are no longer necessary. See changes in [6b571e9](https://github.com/Automattic/newspack-plugin/pull/3779/commits/6b571e9e2255e80e66d267ac8c96000b52af7a6c) for details.

After these changes, we can remove the normalize contact data filter and rely on the syncs triggered by data events, since contact metadata from these events will be included in the triggered syncs, and metadata from queued syncs will be carried through to the single sync executed on shutdown.

leogermani: Ah, good call! I think that makes much more sense!

PR #1764, opened 2025-02-25
feat: allow email address updates to ESPs

chickenn00dle:
### All Submissions:

See https://github.com/Automattic/newspack-plugin/pull/3780 for details

See https://github.com/Automattic/newspack-plugin/pull/3780 for testing instructions

### Other information:

### DIFF
--- a/includes/service-providers/constant_contact/class-newspack-newsletters-constant-contact.php
+++ b/includes/service-providers/constant_contact/class-newspack-newsletters-constant-contact.php
@@ -1199,8 +1199,15 @@ final class Newspack_Newsletters_Constant_Contact extends \Newspack_Newsletters_
 				$data['custom_fields'][ strval( $key ) ] = strval( $value );
 			}
 		}
-
-		$result = $cc->upsert_contact( $contact['email'], $data );
+		$email = $contact['email'];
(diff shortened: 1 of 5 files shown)


includes/service-providers/constant_contact/class-newspack-newsletters-constant-contact.php
```diff
@@ -1196,11 +1196,19 @@ public function add_contact( $contact, $list_id = false ) {
 		if ( isset( $contact['metadata'] ) ) {
 			$data['custom_fields'] = [];
 			foreach ( $contact['metadata'] as $key => $value ) {
-				$data['custom_fields'][ strval( $key ) ] = strval( $value );
+				// Update request will fail if metadata contains an empty value, so we set these to '-'.
```
chickenn00dle, line 1199: As the comment notes, update requests fail with the following error: `Value can't be blank`. So I'm replacing empty string here with `-`. Alternatively I can remove these values from the `custom_fields` array altogether, but that would need to involve checking the existing metadata to verify a value that currently exists isn't intentionally being removed.

Let me know which we prefer!

adekbadek (approved):
Haven't been able to test with CC ESP due to login issues on their platform. But I'll assume it's ok, and the code looks good 🤞

PR #1769, opened 2025-03-03
fix: handle email change esp sync errors

chickenn00dle:
### All Submissions:

See https://github.com/Automattic/newspack-plugin/pull/3792 for details

See https://github.com/Automattic/newspack-plugin/pull/3792 for testing instructions

### Other information:

### DIFF
--- a/includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign.php
+++ b/includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign.php
@@ -1301,30 +1301,15 @@ final class Newspack_Newsletters_Active_Campaign extends \Newspack_Newsletters_S
 			$contact['metadata'] = [];
 		}
 		$action  = 'contact_add';
+		$email   = $contact['email'];
 		$payload = [
-			'email' => $contact['email'],
+			'email' => $email,
 		];
-
 		$has_list_id = false !== $list_id;
 		if ( $has_list_id ) {
 			$payload[ 'p[' . $list_id . ']' ]      = $list_id;
 			$payload[ 'status[' . $list_id . ']' ] = 1;
 		}
-		$email            = $contact['email'];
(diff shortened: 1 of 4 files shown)


includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign.php
```diff
@@ -1381,6 +1365,22 @@ public function add_contact( $contact, $list_id = false ) {
...
+			$action               = 'contact_edit';
+			$payload['id']        = is_wp_error( $contact_data ) ? $existing_contact_data['id'] : $contact_data['id'];
+			$payload['overwrite'] = 0;
+			// If the email address exists, but is different from the one we're trying to upsert, delete the existing contact.
+			if ( ! is_wp_error( $contact_data ) && ! is_wp_error( $existing_contact_data ) && $existing_email !== $email ) {
+				$delete_res = $this->delete_contact( $existing_email );
+				if ( is_wp_error( $delete_res ) ) {
+					Newspack_Newsletters_Logger::log( 'Error deleting existing contact during upsert: ' . $delete_res->get_error_message() );
+				}
+			}
```
miguelpeixe, line 1382: ActiveCampaign's automations can rely a lot on contact activity and history. Won't this introduce a reset and impact how publishers handle their campaigns?

miguelpeixe (approved):
Working as described! I left a non-blocking question, which could be valid for Mailchimp as well but I'm more familiar with this practice on ActiveCampaign.

includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign.php
chickenn00dle: I think it would, although I don't think there is an alternative solution since there is no way to "replace" a contacts activity and history with another's that I know of. This is also something we don't handle currently and is probably an edge case, so I don't think its something we need to worry about unless we get a report.

PR #1776, opened 2025-03-07
fix: add UTM parameters to link URLs inside newsletter content

dkoo:
### All Submissions:

We currently add `utm_medium=email` to all link URLs inside newsletter content, but no other params. This PR adds the following UTM params as well:

- `utm_campaign` - Should match the value of the Campaign Name field (but URL-encoded)
- `utm_source` - The ID of the top-level send list the newsletter is being sent to (but URL-encoded)

Note that the connected ESP or other third-party integrations may overwrite some or all of these parameters with its own. For example, if using Mailchimp with the [Google Analytics extension](https://mailchimp.com/help/integrate-google-analytics-with-mailchimp/), the GA extension will set its own `utm_campaign` and `utm_source` values, overwriting our defaults.

1. Create a new newsletter and make sure its content contains some links.
2. Save the newsletter to trigger a sync to the connected ESP.
3. In the connected ESP, view the synced campaign's HTML and confirm that all links in the content have `utm_medium`, `utm_campaign`, and `utm_source` parameters as defined above
4. Send the campaign to a test list at which you can receive the live email. Click through the links in the email content and confirm that `utm_medium`, `utm_campaign`, and `utm_source` parameters are passed through to the ultimate destination. (This last step is necessary because most ESPs filter in-content URLs through their own tracking redirects.)

### Other information:

---
  - https://app.asana.com/0/0/1209564094585411

### DIFF
--- a/includes/class-newspack-newsletters-renderer.php
+++ b/includes/class-newspack-newsletters-renderer.php
@@ -407,10 +407,10 @@ final class Newspack_Newsletters_Renderer {
 			'utm_medium' => 'email',
 		];
 		if ( $campaign_name ) {
-			$utm_params['utm_campaign'] = $campaign_name;
+			$utm_params['utm_campaign'] = rawurlencode( $campaign_name );
 		}
 		if ( $send_list_id ) {
-			$utm_params['utm_source'] = $send_list_id;
+			$utm_params['utm_source'] = rawurlencode( $send_list_id );
 		}
 		foreach ( $urls as $index => $url ) {
 			/** Skip if link was already processed. */
--- a/includes/class-newspack-newsletters-renderer.php
+++ b/includes/class-newspack-newsletters-renderer.php
@@ -398,8 +398,20 @@ final class Newspack_Newsletters_Renderer {
 	 */
 	public static function process_links( $html, $post = null ) {
 		preg_match_all( '/href="([^"]*)"/', $html, $matches );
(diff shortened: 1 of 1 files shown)


includes/class-newspack-newsletters-renderer.php
```diff
@@ -398,8 +398,20 @@ function ( $key ) use ( &$attrs ) {
 	 */
 	public static function process_links( $html, $post = null ) {
 		preg_match_all( '/href="([^"]*)"/', $html, $matches );
-		$href_params = $matches[0];
-		$urls        = $matches[1];
+		$href_params   = $matches[0];
+		$urls          = $matches[1];
+		$provider      = Newspack_Newsletters::get_service_provider();
```
adekbadek, line 403: If the provider is manual, `$provider` will be null. In this case, I'd suggest the campaign to be the post title, and source empty.

adekbadek (changes requested):

includes/class-newspack-newsletters-renderer.php
dkoo: Whoops, good call! 74185cf defaults `utm_campaign` to the post title if `$provider` isn't available, and `utm_source` doesn't get set at all if there's no `send_list_id` meta on the post (which would be the case for a post created with the "manual" provider).

adekbadek (approved):

PR #1799, opened 2025-04-10
fix: subscribe to local list methods

leogermani:
### All Submissions:

Fixes an issue where a contact would not be subscribed if subscribing only to a local list.

In the `add_tag_to_contact` and `add_group_to_contact` methods, there was a check for an existing contact. If the contact didn't exist, we were not adding it.

This PR fixes it.

ATTENTION: THIS IS NOT READY. It's probably breaking Mailchimp, because we are passing the label instead of the ID. But for groups, we want the ID, while for tags we want the label. Need to figure this out yet.

Repeat this with all ESPs

1. Set up a new local list
2. Add a subscription block to a page
3. as a visitor, subscribe via block, entering a name and making sure to select ONLY the local list
4. Go to the ESP and confirm the contact was added in the correct list/audience with the correct group/tag

### Other information:


leogermani:
Closing in favor of #1801

PR #1814, opened 2025-04-15
fix: ignore sticky posts in the post inserter query when not otherwise manipulated

laurelfulford:
### All Submissions:

This is an odd edge case, but when you have some posts marked as 'Sticky' and 'Display sponsored posts' turned on, the sticky posts will show at the top of the Post Inserter block. When you disable 'Display sponsored posts', the plugin manipulates the post query so sticky posts aren't at the top by default.

1. Mark 2-3 older posts as Sticky. You can do this from the Quick Edit view under Posts, or by clicking "Published" in the sidebar on a single post (that opens a model with "Sticky" at the bottom).
2. Make a new newsletter, and add the Post Inserter block.
3. Note your sticky post(s) are at the top, ahead of newer posts. 
4. Turn off "Display sponsored posts" and note your newer posts are now at the top.
5. Apply this PR.
6. Refresh the editor, and toggle on and off "Display sponsored posts"; confirm that the sticky posts don't appear at the top in either case. 

See p1744750283207519-slack-newspack-support

### Other information:

### DIFF
--- a/includes/class-newspack-newsletters-editor.php
+++ b/includes/class-newspack-newsletters-editor.php
@@ -52,8 +52,7 @@ final class Newspack_Newsletters_Editor {
 		add_filter( 'block_categories_all', [ __CLASS__, 'add_custom_block_category' ] );
 		add_filter( 'allowed_block_types_all', [ __CLASS__, 'newsletters_allowed_block_types' ], 10, 2 );
 		add_action( 'rest_post_query', [ __CLASS__, 'maybe_filter_excerpt_length' ], 10, 2 );
-		add_action( 'rest_post_query', [ __CLASS__, 'maybe_exclude_sponsored_posts' ], 10, 2 );
-		add_action( 'rest_post_query', [ __CLASS__, 'rest_post_query_filter' ] );
+		add_action( 'rest_post_query', [ __CLASS__, 'rest_post_query_filter' ], 10, 2 );
 		add_action( 'rest_api_init', [ __CLASS__, 'add_newspack_author_info' ] );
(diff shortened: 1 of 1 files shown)


includes/class-newspack-newsletters-editor.php
```diff
@@ -500,6 +500,9 @@ public static function maybe_exclude_sponsored_posts( $args, $request ) {
 					];
 				}
 			}
+		} else {
+			// If we're not removing sponsors by modifying the query, make sure to ignore sticky posts.
+			$args['ignore_sticky_posts'] = true;
```
leogermani: nit, but it seems to me that this method should be renamed and this clause should be always added...

Looks like this is needed in all cases, and it's odd that this lives just in a small `else` clause of something that is not completely related.

It could be called simply `rest_post_query_filter`, add this clause at the top and handle sponsors if needed.

Makes sense?

laurelfulford: Thanks Leo -- that makes sense! I went with this route since the changes we make to the query to remove sponsored posts also causes sticky posts not to stick, but it doesn't make for as clear of code. 

I went with the suggested more generic function in [3d2f3c1](https://github.com/Automattic/newspack-newsletters/pull/1814/commits/3d2f3c191b6a1bfdcb9bee8fc44094ec66b1dcd6). `maybe_filter_excerpt_length` also uses the same filter, but I left it out of the combined function it doesn't change the args, just filters the excerpt length, but let me know if it'd make sense to include, too!

leogermani (approved):

laurelfulford:
Thanks @leogermani!

PR #1835, opened 2025-05-27
Feat/tracking improvements

leogermani:
### All Submissions:

This PR does two things.

1. Remove the opened and click count from the Newsletter UI. Our main purpose is to track Ads. Tracking the newsletter itself was a byproduct and generates confusion because the number can differ from what people see in their ESP. 

2. Improves how the log is processed to avoid memory limits
* Process 500 at each time, and not 100
* Stores the line where we stopped, and start again from there in the next request
* Don't truncate the file and create a new one
* Instead, keep reading from the same file. Whenever we get to the end, delete the file and start a new one

Confirm that the "opened" and "clicked" columns are no longer visible in the Newsletters admin list.

1. Look at the tests
4. Change the default `$max_lines` value in the `process_logs` method to 1, to make it easier to test
5. Write a Newsletter with a link in it and send a test
6. Grab the link from the email, and access it several times, changing the email parameter
7. Confirm you see several lines being created in the log files
8. Trigger the cron job `wp cron event run newspack_newsletters_tracking_pixel_process_log` a few times
9. Check the `tracking_pixel_seen` post meta to see that the counts are being bumped in each run
10. Confirm that after all the lines in the file are processed, the file gets deleted

### Other information:


dkoo (approved):
Works as described!

EDIT: now I'm not so sure...see comment below

dkoo:
@leogermani one thing I encountered but am not sure is an issue: when changing the `$max_lines` param in `process_logs()` to `1`, every time I run the `newspack_newsletters_tracking_pixel_process_log` cron event, it processes one line. If I immediately run the `newspack_newsletters_tracking_pixel_process_log` cron event again, I see `Error: Invalid cron event 'newspack_newsletters_tracking_pixel_process_log'`, but the log file still exists on the server. If I then "clicked" on the link again, it would schedule a new cron event that I could run to process the next line.

Does this mean that if there are more events in the logfile than `$max_lines`, the rest won't be processed until another event occurs?

leogermani:
> Does this mean that if there are more events in the logfile than $max_lines, the rest won't be processed until another event occurs?

Yes.. We don't want to process the whole file in one go. Is that what you're asking?

But another event should be scheduled again every minute right?

dkoo:
> But another event should be scheduled again every minute right?

That's what I thought, but I'm not seeing it getting rescheduled automatically unless there's another click or view event.

leogermani:
> That's what I thought, but I'm not seeing it getting rescheduled automatically unless there's another click or view event.

Gotcha! I'll check!

leogermani:
@dkoo I gave it some test and it seems to be working fine.

I spent some time (much more than I wanted) debugging WP_Cron on my local environment just to realize it's not working as it should... maybe it has something to do with our Docker environment... or it's just me... let's see

leogermani:
I'm still testing and validating this though, so I'll keep it open for a little longer

leogermani:
DId one final adjustment in https://github.com/Automattic/newspack-newsletters/pull/1835/commits/9202782ae3be1ceb0f86dcbf5ee0855249b29997 and this is now ready to be merged!

miguelpeixe (approved):
💯

PR #1837, opened 2025-05-29
fix: delete contact by email address, not user ID

dkoo:
### All Submissions:

Related to https://github.com/Automattic/newspack-plugin/pull/4008. By the time the `Newspack_Newsletters_Contacts::delete()` method is called, the `$user` account will no longer exist, so this method will always return a `newspack_newsletters_invalid_user` error. This PR changes the method to use the deleted user's email address instead.

See testing instructions in https://github.com/Automattic/newspack-plugin/pull/4008.

### Other information:

### DIFF
--- a/includes/class-newspack-newsletters-contacts.php
+++ b/includes/class-newspack-newsletters-contacts.php
@@ -261,19 +261,14 @@ class Newspack_Newsletters_Contacts {
 	/**
 	 * Permanently delete a user subscription.
 	 *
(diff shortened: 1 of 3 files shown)


adekbadek (approved):
`delete_user_subscription` would need an update, too, since it uses the method with changed signature internally. It's deprecated since Aug '24, so I've removed it instead in efe08a1c9b26fcc8f354111f7414593670a3c197.

PR #1807, opened 2025-04-14
feat(admin-menu): remove non-applicable admin menu items

adekbadek:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Adjusts admin menu items to align with capabilities.

Part of NPPM-336

### How to test the changes in this Pull Request:

1. On `trunk`,
2. Activate `capability-manager-enhanced` plugin and edit the Editor role 
3. In the "Unique Post Type Capabilities" choose "Newsletters" and "Newsletter Ads" and click "Update"
4. Deselect all capabilities in "Newsletters" and "Newsletter Ads" rows:

image

5. Log in as the editor, observe Newsletters menu item with Categories, Tags, Advertising submenu items
6. Switch to this branch, observe no Newsletter menu item
7. Adjust editor permissions to all all of Newsletters, but not Newsletters Ads → observe the menu item displayed accordingly
8. Adjust permissions to all of Newsletters and Newsletter Ads → observe the menu item displayed accordingly
9 Adjust permissions to all of Newsletter Ads, but not Newsletters → observe the menu item displayed accordingly

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

PR #1854, opened 2025-06-27
fix: respect user permissions

leogermani:
### All Submissions:

Closes # .

Depends on https://github.com/Automattic/newspack-plugin/pull/4053

1. There are four scenarios to be tested, logged in as an editor

A - editor with permission to edit newsletters and Newspack plugin active
B - editor with permission to edit newsleters and Newspack plugin inactive
C - editor without permissino to edit newsletters and Newspack plugin active
D - editor without permission to edit newsleters and Newspack plugin inactive

In order to grant remove permissions to edit newsletters:

* Activate `capability-manager-enhanced` plugin and edit the Editor role 
* In the "Unique Post Type Capabilities" choose "Newsletters" and "Newsletter Ads" and click "Update"
* Deselect all capabilities in "Newsletters" and "Newsletter Ads" rows:

image

Now make sure you see or don't see the Newsletters menu in the expected way in each situation, and things work as expected.

Known issue:

Even without any permissions to see Newsletters, you will still see the menu with the Tags and Categories items under it... 

### Other information:

### DIFF
--- a/includes/class-newspack-newsletters-ads.php
+++ b/includes/class-newspack-newsletters-ads.php
@@ -177,6 +177,11 @@ final class Newspack_Newsletters_Ads {
 	 * Add ads page link.
 	 */
 	public static function add_ads_page() {
+
+		if ( ! \Newspack_Newsletters::user_can_edit_newsletters() ) {
+			return;
+		}
+
 		add_submenu_page(
 			'edit.php?post_type=' . Newspack_Newsletters::NEWSPACK_NEWSLETTERS_CPT,
 			__( 'Newsletters Ads', 'newspack-newsletters' ),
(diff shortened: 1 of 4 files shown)


adekbadek (changes requested):
With "Newsletter Ads" not permitted for an editor, the submenu item is still displayed (both with and without Newspack Plugin active). 

> Even without any permissions to see Newsletters, you will still see the menu with the Tags and Categories items under it...

This would be confusing for a user. 

---

Why a need for an alternative solution to https://github.com/Automattic/newspack-newsletters/pull/1807, when the original solution does not have these issues?

leogermani:
> Why a need for an alternative solution to #1807, when the original solution does not have these issues?

You're right, [we don't](https://github.com/Automattic/newspack-newsletters/pull/1807#pullrequestreview-2987422636).

After all the conversations we had I thought the root cause here was something else. I was trying to keep the consistency with the AI approach that hosts the overrides in the main plugin. But now I see that we are manually moving the Ads CPT menu around and that't the issue. Your fix is fine.

PR #1872, opened 2025-07-26
feat: improve CLI

leogermani:
### All Submissions:

This adds some memory management for this CLI command, to make it run in big sites. Also some parameters to allow us to resume a job from where it left, in case something goes wrong.

This has been tested already, so only needs some sanity check.

### Other information:

### DIFF
--- a/includes/plugins/woocommerce-memberships/class-sync-membership-tied-subscribers-cli.php
+++ b/includes/plugins/woocommerce-memberships/class-sync-membership-tied-subscribers-cli.php
@@ -54,6 +54,52 @@ Note that if a member has unsubscribed from a list, but has an active membership
 		$active_statuses = wc_memberships()->get_user_memberships_instance()->get_active_access_membership_statuses();
 		return in_array( $user_membership->get_status(), $active_statuses );
 	}
+	/**
+	 * Reset the local WordPress object cache
+	 *
+	 * This only cleans the local cache in WP_Object_Cache, without
+	 * affecting memcache
+	 */
+	private static function reset_local_object_cache() {
+		global $wp_object_cache;
+
+		if ( ! is_object( $wp_object_cache ) ) {
+			return;
+		}
+
+		$properties = [
+			'group_ops',
+			'memcache_debug',
+			'cache',
+		];
+
+		foreach ( $properties as $property ) {
+			if ( property_exists( $wp_object_cache, $property ) ) {
+					$wp_object_cache->$property = [];
+			}
+		}
+
+		if ( method_exists( $wp_object_cache, '__remoteset' ) ) {
+			$wp_object_cache->__remoteset(); // important.
+		}
+	}
+
+	/**
+	 * Reset the WordPress DB query log
+	 */
+	private static function reset_db_query_log() {
+		global $wpdb;
+
+		$wpdb->queries = [];
+	}
+
+	/**
+	 * Cleanup memory.
+	 */
+	private static function memory_cleanup() {
+		self::reset_local_object_cache();
+		self::reset_db_query_log();
+	}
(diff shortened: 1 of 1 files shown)


includes/plugins/woocommerce-memberships/class-sync-membership-tied-subscribers-cli.php
```diff
@@ -123,18 +188,43 @@ public static function cli_sync_membership_tied_subscribers( $args, $assoc_args
...
+						}
 						$list_public_id = $list->get_public_id();
 						\WP_CLI::log( sprintf( '  - Synchronizing list "%s" (#%d, public ID: %s)', $list->get_title(), $list->get_id(), $list_public_id ) );
+
+						$user_count = 0;
+
 						foreach ( $plan_memberships as $user_membership ) {
+
+							$user_count++;
+							if ( $user_count > 50 && function_exists( 'np_stop_the_insanity' ) ) {
```
rbcorrales: The call is gated on `function_exists( 'np_stop_the_insanity' )`. If that function doesn't exist, memory cleanup won't run. Is this intended? I don't see it added. Maybe a leftover from a refactor?

Also, probably not necessary to check its existence if the cleanup functions are part of the current class context.

includes/plugins/woocommerce-memberships/class-sync-membership-tied-subscribers-cli.php
```diff
@@ -123,18 +188,43 @@ public static function cli_sync_membership_tied_subscribers( $args, $assoc_args
...
 							$email = $user->user_email;
 							if ( ! $email ) {
 								\WP_CLI::warning( sprintf( 'No email for user #%d, skipping.', $user->ID ) );
 								continue;
 							}
+
+							if ( ! $start_processing_users && $resume_from_email === $email ) {
+								$start_processing_users = true;
+								\WP_CLI::log( sprintf( '  - Resuming from email %s.', $email ) );
+							}
```
rbcorrales, line 222: The logic here might work if the list is static, but if the list of users coming from `$plan->get_memberships()` changes between runs, and the command is restarted with the `resume-from-email` parameter, there is a chance to leave some users unprocessed. Or if the user is no longer present, then nothing will be processed. I'm not familiar enough with how these memberships work, but these are possibilities. Disregard if the user list is immutable.

rbcorrales (changes requested):

includes/plugins/woocommerce-memberships/class-sync-membership-tied-subscribers-cli.php
leogermani: > Maybe a leftover from a refactor?

busted! Thanks for catching that!

includes/plugins/woocommerce-memberships/class-sync-membership-tied-subscribers-cli.php
leogermani: it's not immutable as a new reader can register at any point, but they come in the same order every time...

This is good enough for the purposes of this command, which is used very rarely in very specific situations

includes/plugins/woocommerce-memberships/class-sync-membership-tied-subscribers-cli.php
leogermani: fixed in https://github.com/Automattic/newspack-newsletters/pull/1872/commits/9d4843aeb4191dc060f56589a429e76c29e4f140

rbcorrales (approved):
Thanks for addressing the comments. It will be nice if these functions gets eventually added to a shared library or a CLI class that other commands could extend with these methods.

PR #1938, opened 2025-10-06
fix(activecampaign): only trigger contact deletion when in change email context

chickenn00dle:
### All Submissions:

Closes https://linear.app/a8c/issue/NPPM-2296/urgent-issue-with-newsletter-sign-ups-deleting-contacts-i

This fixes an issue where re-submitting a newsletter subscription form with the same email but with different casing would trigger the email change logic during upserts.

We fix this by:

1. Normalizing the casing for email before upserting
2. Adding a flag for email change upserts to indicate when the email change logic should be triggered.

1. Set up ActiveCampaign as the ESP
2. Add a newsletter subscription block to any page or post
3. Subscribe as a reader using lower case for the email
4. Confirm the contact has been added in Active Campaign
5. Re-submit newsletter subscription form as reader, but make capitalize any letter in the email address
6. On `release` the initial email address will be deleted and a new contact with the capitalized email will be created. On this branch the contact should remain unchanged.\
7. Repeat the above tests with Mailchimp

Test email change flow
8. Now go to my account as the reader, and complete the email change flow
9. Confirm the initial ActiveCampaign contact is correctly deleted and a new email is created using the new email address.

### Other information:

### DIFF
--- a/includes/service-providers/constant_contact/class-newspack-newsletters-constant-contact.php
+++ b/includes/service-providers/constant_contact/class-newspack-newsletters-constant-contact.php
@@ -1201,9 +1201,9 @@ final class Newspack_Newsletters_Constant_Contact extends \Newspack_Newsletters_
 			}
 		}
 
-		$email = $contact['email'];
+		$email = trim( strtolower( $contact['email'] ) );
 		if ( isset( $contact['existing_contact_data']['email_address'] ) ) {
-			$existing_email = $contact['existing_contact_data']['email_address'];
+			$existing_email = trim( strtolower( $contact['existing_contact_data']['email_address'] ) );
 			if ( $existing_email->address !== $email ) {
 				$data['email_address'] = $email;
 				$email                 = $existing_email->address;
--- a/includes/service-providers/constant_contact/class-newspack-newsletters-constant-contact.php
+++ b/includes/service-providers/constant_contact/class-newspack-newsletters-constant-contact.php
@@ -1202,7 +1202,7 @@ final class Newspack_Newsletters_Constant_Contact extends \Newspack_Newsletters_
 		}
 
 		$email = trim( strtolower( $contact['email'] ) );
-		if ( isset( $contact['existing_contact_data']['email_address'] ) ) {
+		if ( isset( $contact['existing_contact_data']['email_address']->address ) ) {
 			$existing_email = trim( strtolower( $contact['existing_contact_data']['email_address'] ) );
 			if ( $existing_email->address !== $email ) {
 				$data['email_address'] = $email;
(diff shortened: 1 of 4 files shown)


dkoo (approved):
Working well, thanks for the fix! Will this cover scenarios where a contact might already have signed up with a non-lowercase email address, though?

chickenn00dle:
Thanks for the review @dkoo 

I tested this for Mailchimp and ActiveCampaign and was able to confirm this covers the scenario where an email is initially signed up with caps. In ActiveCampaign, the second attempt to subscribe causes the contact email to update to lowercase, and in Mailchimp it remains the same. In both cases the original contacts persisted.

I was not able to test this with Constant Contact though since my key seems to not be valid anymore... Any chance you can test this on your end. Here is what I did to test:

1. With `release` checked out, subscribe to a newsletter as a new reader with an email address that contains one or more capital letters.
2. Checkout this branch and build
3. Attempt to subscribe with the same email with no caps/different letters capped
4. Verify the contact is not deleted or archived in the ESP

includes/service-providers/constant_contact/class-newspack-newsletters-constant-contact.php
```diff
@@ -1201,9 +1201,9 @@ public function add_contact( $contact, $list_id = false ) {
 			}
 		}
 
-		$email = $contact['email'];
+		$email = trim( strtolower( $contact['email'] ) );
 		if ( isset( $contact['existing_contact_data']['email_address'] ) ) {
```
dkoo:
```suggestion
		if ( isset( $contact['existing_contact_data']['email_address']->address ) ) {
```

includes/service-providers/constant_contact/class-newspack-newsletters-constant-contact.php
```diff
@@ -1201,9 +1201,9 @@ public function add_contact( $contact, $list_id = false ) {
...
-			$existing_email = $contact['existing_contact_data']['email_address'];
+			$existing_email = trim( strtolower( $contact['existing_contact_data']['email_address'] ) );
 			if ( $existing_email->address !== $email ) {
 				$data['email_address'] = $email;
 				$email                 = $existing_email->address;
```
dkoo:
```suggestion
			$existing_email = trim( strtolower( $contact['existing_contact_data']['email_address']->address ) );
			if ( $existing_email->address !== $email ) {
				$data['email_address'] = $email;
				$email                 = $existing_email;
```

dkoo (changes requested):
@chickenn00dle I was able to test with Constant Contact. Looks like even on `release` the email address is normalized to lowercase upon getting created as a contact (perhaps this is CC API behavior). But I was running into a fatal error when trying to confirm an email change: `[06-Oct-2025 22:26:13 UTC] PHP Fatal error:  Uncaught TypeError: strtolower(): Argument #1 ($string) must be of type string, stdClass given in /newspack-repos/newspack-newsletters/includes/service-providers/constant_contact/class-newspack-newsletters-constant-contact.php:1207
Stack trace:`

Changing the below lines fixes the fatal.

chickenn00dle:
Thanks for testing this @dkoo! I applied your suggestions: https://github.com/Automattic/newspack-newsletters/pull/1938/commits/e0b405edefb7f7f53204ebd9e54048463a9f6a9d and https://github.com/Automattic/newspack-newsletters/pull/1938/commits/588769700bf976ae3f1d7e2f553847fa800a3b87

dkoo (approved):
Working well now! Thanks for the revisions!

PR #1961, opened 2025-11-15
Use continue instead of return when bulk tracking pixel processing

claudiulodro:
### All Submissions:

Currently the function that records log views uses a `return` if it encounters an invalid newsletter tracking ID. When that happens, all of the other events after it in the list get thrown away and are not tracked. Using `break` is better IMO, so any valid events after the invalid one get processed.

Not sure, but I encountered it in the wild. Something like this:

1. Have a log file with multiple newsletter's tracking pixels. 
2. Manually delete or change the tracking ID meta on one of the newsletters, so the check fails.
3. `wp shell` and `Newspack_Newsletters\Tracking\Pixel::process_logs()`
4. Observe the data for events after the invalid newsletter isn't recorded.

### Other information:

### DIFF
--- a/includes/tracking/class-pixel.php
+++ b/includes/tracking/class-pixel.php
@@ -196,7 +196,7 @@ final class Pixel {
 
 			// Bail if tracking ID mismatch.
 			if ( $newsletter_tracking_id !== $tracking_id ) {
-				return;
+				break;
 			}
 
 			$pixel_seen = \get_post_meta( $newsletter_id, 'tracking_pixel_seen', true );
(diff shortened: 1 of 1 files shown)


includes/tracking/class-pixel.php
```diff
@@ -196,7 +196,7 @@ public static function bulk_track_seen( $events ) {
 
 			// Bail if tracking ID mismatch.
 			if ( $newsletter_tracking_id !== $tracking_id ) {
-				return;
+				break;
```
leogermani: shouldn't it be `continue`?

With `break` it seems that the result will be the same, since it will go to the end of the `foreach` statement which is just before the method end...

claudiulodro: Oh yeah you're totally right 🤦

dkoo (approved):

PR #2005, opened 2026-01-05
fix(ads): start/end date value format

miguelpeixe:
### All Submissions:

NPPM-2482

Start/end dates for newsletter ads do not consider time. It's uncommon, but `get_option( 'date_format' )` may include the time, which can display a value that is misleading to the user.

1. While on trunk, navigate to Settings -> General
2. Modify date format to `F j, Y, g:i a`
3. Navigate to `Newsletters -> Ads` and create an ad with start and end dates
4. Back to the `Newsletters -> Ads` table, confirm the value renders the time
5. Checkout this branch, refresh the page, and confirm the time is omitted

### Other information:

### DIFF
--- a/includes/ads/class-ads.php
+++ b/includes/ads/class-ads.php
@@ -615,14 +615,14 @@ final class Ads {
 		if ( 'start_date' === $column_name ) {
 			$start_date = get_post_meta( $post_id, 'start_date', true );
 			if ( ! empty( $start_date ) ) {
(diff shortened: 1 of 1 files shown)


dkoo (approved):
This works...it's a shame we can't just strip the "time" variables from the `get_option( 'date_format' )` string, but that seems like it'd add a lot of complexity for little gain.

miguelpeixe:
I don't think there's much gain in a dynamic date format here. Enforcing a standard seems better.

PR #2019, opened 2026-01-27
fix: restore newsletter rendering of RDB blocks using get_value()

jason10lee:
### All Submissions:

Recent changes to Remote Data Blocks meant that data from remote bindings were no longer persisted in post content.  Newsletter rendering (but not in-editor preview!) of RDB-derived blocks depended on that behavior, and remote data that depended on those bindings was no longer included on send.

This patch (heavily inspired by the new preview rendering behavior in RDB) uses the new get_value() implementation in RDB to resolve those bindings and populate remote data on send.  This essentially restores the old behavior as far as our newsletters are concerned.

Addresses [NPPM-2483: Continuing problem with trying to add Foundation -- Event block to newsletters](https://linear.app/a8c/issue/NPPM-2483/continuing-problem-with-trying-to-add-foundation-event-block-to).

1.  Ensure that Newspack Newsletters and Foundation Integration are set up on your test environment.
2.  Pull down the latest version of the [`fix/rdb-rendering` topic branch](https://github.com/Automattic/newspack-newsletters/tree/fix/rdb-rendering).
3.  Create a new newsletter.
4. Add a RDB block (newsletters support `Foundation - Event` or `Foundation - Location` blocks) to the newsletter; select data to display in it; choose a Newsletter pattern.  Verify the remote data shows up in the editor.
5. Hit **Preview Email** button in the upper right.  Verify the remote data shows on the preview popup.
6. **If you're set up to test newsletter sending** hit **Send** in the upper right and kick off a test send.  Verify the recipient sees the remote data.

### Other information:

### DIFF
--- a/includes/class-newspack-newsletters-renderer.php
+++ b/includes/class-newspack-newsletters-renderer.php
@@ -515,7 +515,9 @@ final class Newspack_Newsletters_Renderer {
 	 * Expand template blocks for multiple query results.
 	 * Creates a copy of inner blocks for each result with the appropriate index.
 	 *
-	 * Heavily inspired by new RDB preview rendering.
+	 * Heavily inspired by new RDB preview rendering.  See changes to
+	 * src/block-editor/binding-sources/remote-data-binding.ts at
+	 * https://github.com/Automattic/remote-data-blocks/commit/a4b9249e248140c741e987f949b215be5edabf1d
 	 *
 	 * @param array $inner_blocks The inner blocks to clone.
 	 * @param array $results      The query results.
--- a/includes/class-newspack-newsletters-renderer.php
+++ b/includes/class-newspack-newsletters-renderer.php
@@ -812,9 +812,9 @@ final class Newspack_Newsletters_Renderer {
 	 */
 	private static function update_link_text_in_html( $html, $text ) {
 		return preg_replace_callback(
-			'/>([^<]*)<\/a>/',
+			'/(>)(.*?)(<\/a>)/s',
 			function ( $matches ) use ( $text ) {
-				return '>' . wp_kses_post( $text ) . '</a>';
+				return $matches[1] . wp_kses_post( $text ) . $matches[3];
 			},
 			$html
 		);
--- a/includes/class-newspack-newsletters-renderer.php
+++ b/includes/class-newspack-newsletters-renderer.php
@@ -1638,8 +1638,7 @@ final class Newspack_Newsletters_Renderer {
 
 		$is_posts_inserter_block = 'newspack-newsletters/posts-inserter' == $block_name;
 		$is_grouped_block        = in_array( $block_name, [ 'core/group', 'core/list', 'core/list-item', 'core/quote' ], true );
-		// Remote Data Blocks are rendered to HTML then converted to MJML with proper structure,
-		// so they should not be wrapped in additional section/column elements.
+		// Remote Data Blocks require special handling to resolve bindings.
 		$is_rdb_block = strpos( $block_name, 'remote-data-blocks/' ) === 0;
 
 		if (
--- a/includes/class-newspack-newsletters-renderer.php
+++ b/includes/class-newspack-newsletters-renderer.php
@@ -498,6 +498,305 @@ final class Newspack_Newsletters_Renderer {
 	}
 
 	/**
+	 * Get Remote Data Blocks context from a container block.
+	 *
+	 * @param array $block The block array.
+	 * @return array The remote data context or empty array if not found.
+	 */
+	private static function get_rdb_context( $block ) {
+		// Remote data is stored in the block's remoteData attribute.
+		if ( isset( $block['attrs']['remoteData'] ) ) {
+			return $block['attrs']['remoteData'];
+		}
+		return [];
+	}
+
+	/**
+	 * Expand template blocks for multiple query results.
+	 * Creates a copy of inner blocks for each result with the appropriate index.
+	 *
+	 * Heavily inspired by new RDB preview rendering.
+	 *
+	 * @param array $inner_blocks The inner blocks to clone.
+	 * @param array $results      The query results.
+	 * @param array $remote_data  The full remote data context.
+	 * @return array Expanded array of blocks with indices set.
+	 */
+	private static function expand_rdb_template_blocks( $inner_blocks, $results, $remote_data ) {
+		$expanded_blocks = [];
+
+		foreach ( $results as $index => $result ) {
+			$source_args = [
+				'block'            => $remote_data['blockName'] ?? '',
+				'enabledOverrides' => $remote_data['enabledOverrides'] ?? [],
+				'index'            => $index,
+				'queryKey'         => $remote_data['queryKey'] ?? null,
+				'queryInputs'      => $remote_data['queryInputs'] ?? null,
+			];
+
+			$cloned_blocks = self::clone_blocks_with_index( $inner_blocks, $source_args );
+			$expanded_blocks = array_merge( $expanded_blocks, $cloned_blocks );
+		}
+
+		return $expanded_blocks;
+	}
+
+	/**
+	 * Deep clone blocks and merge source args into all binding configurations.
+	 *
+	 * @param array $blocks      The blocks to clone.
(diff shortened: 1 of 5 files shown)


includes/class-newspack-newsletters-renderer.php
```diff
@@ -497,6 +497,305 @@ function( $matches ) use ( $properties ) {
...
+
+	/**
+	 * Update link text in HTML.
+	 *
+	 * @param string $html The original HTML.
+	 * @param string $text The new link text.
+	 * @return string The updated HTML.
+	 */
+	private static function update_link_text_in_html( $html, $text ) {
+		return preg_replace( '/>([^<]*)<\/a>/', '>' . $text . '</a>', $html );
```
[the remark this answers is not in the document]
jason10lee: Incoming data may have HTML; using `wp_kses_post()` to clean.

includes/class-newspack-newsletters-renderer.php
```diff
@@ -497,6 +497,305 @@ function( $matches ) use ( $properties ) {
...
+	/**
+	 * Update href attribute in HTML (for links/buttons).
+	 *
+	 * @param string $html The original HTML.
+	 * @param string $href The new href value.
+	 * @return string The updated HTML.
+	 */
+	private static function update_href_in_html( $html, $href ) {
+		return preg_replace( '/href="[^"]*"/', 'href="' . esc_url( $href ) . '"', $html );
+	}
```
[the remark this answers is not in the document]
jason10lee: Addressed in upcoming patch.

includes/class-newspack-newsletters-renderer.php
```diff
@@ -497,6 +497,305 @@ function( $matches ) use ( $properties ) {
...
+	 * @param string $alt  The new alt text.
+	 * @return string The updated HTML.
+	 */
+	private static function update_alt_in_html( $html, $alt ) {
+		// Replace alt attribute value.
+		if ( preg_match( '/alt="[^"]*"/', $html ) ) {
+			return preg_replace( '/alt="[^"]*"/', 'alt="' . esc_attr( $alt ) . '"', $html );
+		}
+		// Add alt if not present.
+		return preg_replace( '/<img/', '<img alt="' . esc_attr( $alt ) . '"', $html );
```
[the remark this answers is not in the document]
jason10lee: Addressed in upcoming patch.

includes/class-newspack-newsletters-renderer.php
```diff
@@ -497,6 +497,305 @@ function( $matches ) use ( $properties ) {
...
+	 * Update src attribute in HTML (for images).
+	 *
+	 * @param string $html The original HTML.
+	 * @param string $url  The new URL.
+	 * @return string The updated HTML.
+	 */
+	private static function update_src_in_html( $html, $url ) {
+		// Replace src attribute value.
+		return preg_replace( '/src="[^"]*"/', 'src="' . esc_url( $url ) . '"', $html );
+	}
```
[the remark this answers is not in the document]
jason10lee: Addressed in upcoming patch.

includes/class-newspack-newsletters-renderer.php
```diff
@@ -1247,8 +1552,47 @@ function ( $block ) {
...
+					];
+					$blocks_to_render = self::clone_blocks_with_index( $inner_blocks, $source_args );
+				}
+
+				// Resolve bindings and update innerHTML.
+				$resolved_blocks = array_map(
+					function ( $b ) use ( $remote_data ) {
+						return self::resolve_rdb_block_bindings( $b, $remote_data );
+					},
+					$blocks_to_render
```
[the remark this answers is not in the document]
jason10lee: That's an excellent observation!  Yes, it does; I'll explore these tests in a future commit.

includes/class-newspack-newsletters-renderer.php
```diff
@@ -497,6 +497,342 @@ function( $matches ) use ( $properties ) {
...
+			return $block['attrs']['remoteData'];
+		}
+		return [];
+	}
+
+	/**
+	 * Expand template blocks for multiple query results.
+	 * Creates a copy of inner blocks for each result with the appropriate index.
+	 *
+	 * Heavily inspired by new RDB preview rendering.
```
chickenn00dle: Nonblocking nit: Might be worth adding a link to that implementation for future reference.

chickenn00dle (changes requested):
This is looking great @jason10lee!

I did run into a few issues though. It seems some of the RDB blocks are not being resolved properly. For example, here is an image from a movie block which doesn't render:

Editor (after ungrouping to resolve nested group error in the editor):
Image

Preview:
Screenshot 2026-01-30 at 15 22 31

Here is another example from an event block of a button which seems to be resolved as a paragraph:

Image

In my testing it seems only images and buttons run into issues, BUT not in every case.

One other nitpick, which is nonblocking is when I select a foundation block and select data, the block renders with this `No results found` text underneath despite results being found:

Image

This happens on trunk as well though, and is not something we need to address in this PR, but wanted to bring it up in case it was an easy fix.

includes/class-newspack-newsletters-renderer.php
jason10lee: Thanks!  I've got an inkling of what's happening on the images/buttons; I'll follow up on those suspicions with a patch if so...

jason10lee: @chickenn00dle, were you using the 'Newsletter' patterns for `Foundation - Event` and `Foundation - Location` blocks?  If not--that's on me for not specifying!  

I believe we've got a few spinoff issues here:
- Remove or hide 'No results found'.
- Filter out non-newsletter RDB patterns in the newsletter editor.
- Support film/movie RDB via newsletter-specific patterns.

I recall some conversation on each; will track down where we ended up on them and create issues if they don't already exist.

chickenn00dle: > were you using the 'Newsletter' patterns for Foundation - Event and Foundation - Location blocks? If not--that's on me for not specifying!

Ah, one of the examples above was using the movie foundation block. So I guess the image issue is accounted for.

However I am still seeing events that aren't correctly rendering the button:
Screenshot 2026-02-02 at 12 51 00

Note: buttons seem to be correctly rendering for locations on my end.

chickenn00dle: > 'Newsletter' patterns for Foundation - Event and Foundation

I should also verify, is there a special block specific to newsletters? This is what I see in the newsletters editor when I search for foundation blocks/patterns:

Screenshot 2026-02-02 at 12 55 35

Are there any specific to newsletters and if so how do I identify them?

jason10lee: Thanks for reviewing, @chickenn00dle!  I think at this point, there are two things to watch out for:

- Use singular `Foundation` blocks (`Foundation - Event`, `Foundation - Location`).
- When selecting the pattern to display, choose the one with `Newsletter` in the name.

If that doesn't render correctly, let me know the endpoint you're connected to--there may be some data coming through that needs to be accounted for!

chickenn00dle: Thanks @jason10lee!

> When selecting the pattern to display, choose the one with Newsletter in the name.

Aha! Found it. That did the trick. Sorry I was conflating the foundation block pattern with the general WP editor concept of pattern and didn't scroll to the bottom to find the former. 

This is working as expected now! 

> Filter out non-newsletter RDB patterns in the newsletter editor.

I think publishers will very easily miss the newsletter specific pattern unless we do this. To clarify, is the plan to tackle this in a follow up PR? It's also worth removing the Foundation - Movie(s) block from the newsletter editor until thats supported.

jason10lee: Excellent; and yes--I'm reviewing our RDB discussions so far with an eye to setting up some followups to tackle as separate issues:

- Remove or hide 'No results found'.
- Filter out non-newsletter RDB patterns in the newsletter editor.
- Support film/movie RDB via newsletter-specific patterns.
- Filter out RDB blocks that aren't supported in the newsletter editor.

chickenn00dle (approved):
Looks good @jason10lee! Thanks for answering all of my questions!

PR #2021, opened 2026-01-30
chore: use reusable WP.org release workflow

rbcorrales:
The inline `release-wporg` job does a fresh checkout + `npm ci` but never builds the `release/` directory that `release-wporg.sh` expects. In CircleCI this worked via `persist_to_workspace` / `attach_workspace`.

This replaces the inline job with the new `reusable-release-wporg.yml` from Automattic/newspack-scripts#226, which downloads the release artifacts uploaded by `reusable-release.yml` and runs `release-wporg.sh`. This also avoids repeating the same inline job across every repo that publishes to WordPress.org.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -30,7 +30,7 @@ jobs:
   release-wporg:
     if: github.ref == 'refs/heads/release'
     needs: release
-    uses: Automattic/newspack-scripts/.github/workflows/reusable-release-wporg.yml@refactor/deprecate-circle-ci
+    uses: Automattic/newspack-scripts/.github/workflows/reusable-release-wporg.yml@trunk
     secrets:
       WP_ORG_USERNAME: ${{ secrets.WP_ORG_USERNAME }}
       WP_ORG_PASSWORD: ${{ secrets.WP_ORG_PASSWORD }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -30,22 +30,7 @@ jobs:
   release-wporg:
     if: github.ref == 'refs/heads/release'
     needs: release
-    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -30,22 +30,7 @@ jobs:
...
-      - name: Install dependencies
-        run: npm ci --legacy-peer-deps
-
-      - name: Release to WordPress.org
-        env:
-          WP_ORG_USERNAME: ${{ secrets.WP_ORG_USERNAME }}
-          WP_ORG_PASSWORD: ${{ secrets.WP_ORG_PASSWORD }}
-          WP_ORG_PLUGIN_NAME: ${{ github.event.repository.name }}
-        run: ./node_modules/newspack-scripts/release-wporg.sh
+    uses: Automattic/newspack-scripts/.github/workflows/reusable-release-wporg.yml@refactor/deprecate-circle-ci
```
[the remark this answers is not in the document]
rbcorrales: This is handled inside the reusable workflow itself. It defaults `WP_ORG_PLUGIN_NAME` to `github.event.repository.name` and also accepts an optional plugin-name input if the slug ever needs to differ. No changes needed on the caller side.

miguelpeixe (approved):

.github/workflows/release.yml
miguelpeixe:
```suggestion
    uses: Automattic/newspack-scripts/.github/workflows/reusable-release-wporg.yml@trunk
```

This is now in trunk

rbcorrales: Updated to trunk: [965b463](https://github.com/Automattic/newspack-newsletters/pull/2021/commits/965b4634dbc1f9e39df4e934e761543c5f830fd3).

PR #2031, opened 2026-02-13
fix: improve error handling in upsert contact method

miguelpeixe:
### All Submissions:

Refactor the reader-friendly error logic so that the response from the `upsert()` method still contains meaningful information about what happened.

Further details and test instructions at https://github.com/Automattic/newspack-plugin/pull/4488

Follow the test instructions from #1754 and confirm there are no regressions.

### Other information:

### DIFF
--- a/includes/class-newspack-newsletters-contacts.php
+++ b/includes/class-newspack-newsletters-contacts.php
@@ -248,15 +248,19 @@ class Newspack_Newsletters_Contacts {
 		);
 
 		if ( $errors->has_errors() ) {
-			// Get a reader-friendly error message to show to the user.
-			$reader_error = $provider->get_reader_error_message(
-				[
-					'email' => $contact['email'],
-					'lists' => $lists,
-				],
-				is_wp_error( $result ) ? $result : $errors
+			$error = new WP_Error(
+				'newspack_newsletters_upsert_contact_error',
+				// Get a reader-friendly error message to show to the user.
+				$provider->get_reader_error_message(
+					[
+						'email' => $contact['email'],
+						'lists' => $lists,
+					],
+					is_wp_error( $result ) ? $result : $errors
+				)
 			);
-			return Newspack_Newsletters::debug_mode() ? $errors : new \WP_Error( 'newspack_newsletters_upsert_contact_error', $reader_error );
+			$error->merge_from( $errors );
+			return $error;
 		}
 
 		return $result;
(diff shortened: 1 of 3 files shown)


includes/class-newspack-newsletters-contacts.php
```diff
@@ -248,15 +248,19 @@ public static function upsert( $contact, $lists = false, $context = 'Unknown', $
...
-			$reader_error = $provider->get_reader_error_message(
-				[
-					'email' => $contact['email'],
-					'lists' => $lists,
-				],
-				is_wp_error( $result ) ? $result : $errors
+			$error = new WP_Error(
+				'newspack_newsletters_upsert_contact_error',
+				// Get a reader-friendly error message to show to the user.
+				$provider->get_reader_error_message(
```
leogermani: Something tricky, but we don't want this friendly error message in the sync logs. We want the raw error message... The same one that goes into `newspack_log` a few lines above.

This get_reader_error_message exists for a reason.

For example, when an email is in compliance mode in Mailchimp, it returns this error message. But this was confusing and not helpful for readers, so we changed that in Mailchimp (see this method in the mailchimp provider). But in our logs, we get the raw error...

So that's the idea: a simpler error message for the reader, but the raw error in the logs for the publisher...

miguelpeixe: Yeah... This is a somewhat hacky solution. The better alternative is a bigger refactor, which is to not return reader-facing message in the `upsert()` method. It's too low-level to worry about that.

I can repurpose this PR. Nothing changes in https://github.com/Automattic/newspack-plugin/pull/4488 though

leogermani: actually maybe this is not hacky. And what we should do is to filter the error message for a better one [here](https://github.com/Automattic/newspack-newsletters/blob/copilot/sub-pr-2028/src/blocks/subscribe/index.php#L531). That is too low level to replace error messages with more human readable ones... we should do this just before surfacing the messages to readers...

miguelpeixe: I meant that my solution is somewhat hacky, because even though we should be sending "all errors", to make the reader-facing message one of those errors feels like an anti-pattern.

Yes, I agree that this is not the right place for the reader-facing message. We should move this closer to the UI methods.

miguelpeixe: Fixed in 6fabbee4607a53abb2cd6343b23f9c5c435c25ac. The upsert error response should no longer include the reader-facing message. This has moved to the relevant UI methods.

I looked through other upsert calls in `newspack-plugin` and they are all backend and sync related.

leogermani (approved):

PR #2042, opened 2026-02-27
feat: add block theme support; update post-inserter

thomasguillot:
### All Submissions:

Improves compatibility with block themes and cleans up the post-inserter block.

**Block theme support**
* Restrict the editor content width to 600px when using a block theme, to match the email max-width
* Support block theme font sizes in the newsletter editor

**Post-inserter block**
* Replace the footer button with an `HStack`-based header that includes the "Insert posts" button — reduces visual noise and simplifies the layout
* Increase the preview `min-height` to `200px` for all alignment variants, ensuring the preview has enough room regardless of image alignment
* Use `@wordpress/base-styles` variables for colors, border-radius, spacing, and font sizes
* Fix the side-image preview transform to correctly center the wider preview canvas

**Typography settings**
* Replace the custom `SelectControlWithOptgroup` component with the core `SelectControl` (which now supports `optgroup`) and remove the custom component

**Editor styles**
* Scope editor styles to the newsletter post type to avoid leaking into other block editors

1. Activate a block theme
2. Create a new newsletter
3. Notice some inconsistencies (content width, post-inserter layout, etc.)
4. Switch to this branch
5. Create a new newsletter
6. Verify the content width is constrained to 600px
7. Open Styles and verify typography settings work correctly
8. Add a Post Inserter block and check the preview with image aligned top, left, and right

### Other information:


laurelfulford (changes requested):
Sorry this took so long, @thomasguillot! This is looking good overall - when I compare the trunk and this PR with the classic theme, there are very slight font size differences -- like things that were 18px are now 16px -- but I don't think it's anything noticeable unless you're comparing them directly.

The only thing that's different that I think could be an issue is the height of the post inserter in some cases. The min-height hasn't changed, but the display height for the Post Inserter is a lot shorter when the image is aligned left or right and there's only one post showing -- this is it using the classic theme:

Image

Compared to the starter template on trunk:

Image

Is it possible to bump that up a bit again?

Otherwise, I'm not sure if we'll get pushback about the subject getting removed, but I'll defer to you there 😅

thomasguillot:
Thanks for the review, @laurelfulford!

**Font size (18px → 16px):** This is intentional — the MJML renderer uses 16px as the default font size when sending emails, so 16px is actually more accurate to what recipients see.

**Post Inserter height:** Addressed in 27033f0e. When the image is aligned left or right, the preview renders at a wider viewport width (1148px), which causes the content to lay out horizontally and thus take up less vertical space. I've added a `min-height: 200px` to the preview for all alignment variants.

thomasguillot:
Updated in 27033f0e to apply the min-height to all preview states, not just side-aligned images.

laurelfulford (changes requested):
Thanks @thomasguillot! 

The height of the post inserter looks better but I just noticed some weirdness with posts without titles in the same layout that doesn't happen on trunk:

Image

I know this block had some style changes recently merged, so it could be that this is only fixed because of that, and those changes just need to be moved to this branch. I poked at merging in trunk, but there are a few conflicts that need review. Can you try updating this branch with trunk and see if that fixes this?

thomasguillot:
Closing this — trunk's #1969 refactor landed after this branch was opened and substantially overlaps with the post-inserter changes here. Rather than untangle the rebase, I'll open a fresh PR scoped to the block theme support pieces (renderer font sizes, editor layout filter, editor styles). cc @laurelfulford for when that lands — will re-request review then.

laurelfulford:
Thanks @thomasguillot!

PR #2078, opened 2026-04-14
fix(block-theme): override theme.json styles in the newsletter editor

chickenn00dle:
### All Submissions:

Block themes inject styles through `theme.json` global styles, which are not removed by `remove_editor_styles()` in `strip_editor_modifications()`. This causes three issues when using the block theme with the newsletter editor:

1. **Buttons** lose their background color/border-radius because `theme.json` uses CSS custom properties (`var(--wp--preset--color--accent)`) that resolve to nothing after theme stripping
2. **Font sizes** use `clamp()`/`rem` values from `theme.json` instead of email-safe `px` values
3. **Spacing** is inconsistent because block theme spacing presets with `clamp()`/`rem` bleed through

This PR:
- Adds a `wp_theme_json_data_theme` filter that replaces block theme values with email-safe `px` equivalents for font sizes, spacing, layout, and button styles when in the newsletter editor context
- Adds default button CSS (`background-color`, `border-radius`, `color`) in the editor stylesheet
- Adds spacing CSS variable overrides (`--wp--preset--spacing--*`) in the editor stylesheet
- Moves the MJML image-caption default attributes from both themes into the plugin renderer (`array_to_attributes`), eliminating duplicate code

Related PRs:
- Automattic/newspack-block-theme — remove migrated MJML filter
- Automattic/newspack-theme — remove migrated MJML filter

Addresses https://linear.app/a8c/issue/NPPD-1316, https://linear.app/a8c/issue/NPPD-1452, https://linear.app/a8c/issue/NPPD-1453, https://linear.app/a8c/issue/NPPD-1454

1. Activate the Newspack Block Theme
2. Create a Newsletter post with all of the non-RDB blocks (I decided not to tackle these here as this one is already quite hairy) listed in [this comment](https://linear.app/a8c/issue/NPPD-1316/newsletter-newsletter-editor-testing#comment-f5f59aa6)
3. Verify buttons have a visible blue background, border-radius, and white text
4. Verify font size presets in the editor show correct px values (Small=12px, Medium=16px, Large=24px, Extra Large=36px)
5. Verify block spacing is consistent (no extra left padding on blocks, quote blocks centered)
6. Send a test email and verify styling in the preview and received email matches the editor
7. Switch to the classic Newspack Theme and repeat — behavior should be identical

### Other information:

🤖 Generated with [Claude Code](https://claude.com/claude-code)


includes/class-newspack-newsletters-renderer.php
```diff
@@ -1434,6 +1521,21 @@ function ( $block ) {
...
+
+					// List and quote blocks skip generic column wrapping (they are
+					// "grouped blocks"). Add an explicit column wrapper so borders
+					// and padding from the block are applied — mj-text does not
+					// support border, but mj-column does.
+					$list_quote_col_attrs = [
+						'padding' => isset( $attrs['padding'] ) ? $attrs['padding'] : '12px',
+					];
+					if ( isset( $attrs['border'] ) ) {
+						$list_quote_col_attrs['border'] = $attrs['border'];
```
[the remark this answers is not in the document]
chickenn00dle: Added in https://github.com/Automattic/newspack-newsletters/pull/2078/commits/8da8da2d53e5c2415feed51c80f7e7e257b38bf1

includes/class-newspack-newsletters-renderer.php
```diff
@@ -376,6 +381,22 @@ private static function process_attributes( $attrs ) {
...
+					'bottom' => '0',
+					'left'   => '0',
+				],
+				$attrs['style']['spacing']['margin']
+			);
+			foreach ( $margin as $key => $value ) {
+				$margin[ $key ] = self::get_spacing_value( $value );
+			}
+			$attrs['margin'] = sprintf( '%s %s %s %s', $margin['top'], $margin['right'], $margin['bottom'], $margin['left'] );
+		}
```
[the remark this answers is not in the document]
chickenn00dle: Added in https://github.com/Automattic/newspack-newsletters/pull/2078/commits/8da8da2d53e5c2415feed51c80f7e7e257b38bf1

laurelfulford (commented):
Thanks for tackling this one, @chickenn00dle!! It's a beast 🙂 

Things are looking good in the block theme with these changes, but I think some of the changes to the classic theme could be an issue on live sites:

**Button Default Styles**
Image

Both the border radius and colour change are pretty noticeable -- would it be possible to keep these the same in the classic theme? I think it's worth trying to change for the block theme, but maybe with a `wp_is_block_theme()` check if it wouldn't get too messy?

Failing that, I think the border radius is fine but we probably need to keep the more neutral dark grey since the blue is pretty opinionated. 

As an aside: thanks for fixing the button colour preview in the editor in this PR! 

**Spacing around buttons**
This looks to be coming from the `$column_attrs['padding'] = ‘0’;` change -- button blocks no longer have padding around them, which reduces spacing around them vertically and -- much more critically -- horizontally. Now they are misaligned on the left compared to other blocks. This is one of our baked-in templates: 

Release:

Image

This PR:

Image

Just let me know if you have questions about either of these, and sorry for being so nitpicky!

laurelfulford (changes requested):
Sorry, the previous comment should've been a 'Request changes' one, switching the review status with this one 🤦‍♀️

chickenn00dle:
Thanks for the review @laurelfulford!

I think I've addressed both points in https://github.com/Automattic/newspack-newsletters/pull/2078/commits/b1ad0af54289dd1c5ebd73d56a910fc330875a3d

Let me know if that did the trick!

chickenn00dle:
Just an FYI @laurelfulford, I noticed an issue with some alignment issues with some of our prebuilt newsletter layouts and snuck in a fix for this here: https://github.com/Automattic/newspack-newsletters/pull/2078/commits/8fd7d1f4b5bd0f3756bf3a7f2e1fa175cf388fb2

includes/class-newspack-newsletters-renderer.php
```diff
@@ -1169,7 +1239,10 @@ function ( $block ) {
 
 					$attrs         = self::process_attributes( $button_block['attrs'] );
 					$text          = $anchor->textContent; // phpcs:ignore WordPress.NamingConventions.ValidVariableName.UsedPropertyNotSnakeCase
-					$border_radius = isset( $attrs['borderRadius'] ) ? $attrs['borderRadius'] : '999px';
+					$is_block_theme       = wp_is_block_theme();
+					$default_bg           = $is_block_theme ? '#36f' : '#32373c';
+					$default_border_radius = $is_block_theme ? '5px' : '999px';
```
laurelfulford: Nit: Given that these values won't change per button, what do you think about moving them outside of `foreach ( $inner_blocks as $button_block ) {` they're all just set once?

includes/class-newspack-newsletters-editor.php
```diff
@@ -273,6 +306,129 @@ public static function newspack_font_sizes() {
...
+		];
+
+		// Only override button element styles for block themes — classic themes
+		// use their own neutral defaults and don't need the opinionated blue.
+		if ( wp_is_block_theme() ) {
+			$email_overrides['styles'] = [
+				'elements' => [
+					'button' => [
+						'color'   => [
+							'background' => '#36f',
```
laurelfulford: Nit: waaaay less of a concern for now since it doesn't affect live sites, but this may still be a bit opinionated for a default. For Lite Site support [I added a bit of code to Newspack Plugin that can get the first colour from the palette](https://github.com/Automattic/newspack-plugin/blob/trunk/includes/lite-site/class-lite-site.php#L377-L387) in a very low stakes feature parity tweak 😆 What do you think about doing something like that?

laurelfulford (approved):
Thanks @chickenn00dle! I think this is looking good overall -- thank you for button tweaks! I added a couple more nits, but they're not blocking.

includes/class-newspack-newsletters-renderer.php
chickenn00dle: Good catch! Updated in https://github.com/Automattic/newspack-newsletters/pull/2078/commits/cd06b54c093039cec5129c0178188ca6d67eeeb7

includes/class-newspack-newsletters-editor.php
chickenn00dle: Oh nice! Thanks for calling this out. I updated to use this method in https://github.com/Automattic/newspack-newsletters/pull/2078/commits/904ad46d9fde772481f0c557d162dc84b7ecdfd0

chickenn00dle:
Thanks for the second round of review @laurelfulford!

I ended up going with your suggestions. Mind having one more look before I merge.

chickenn00dle:
I'll go ahead and merge this one since the last review was an approval and the final two changes were pretty low stakes.

PR #2079, opened 2026-04-14
fix: use active campaign saved segments summary endpoint

leogermani:
### All Submissions:

Fixes the Active Campaign provider by using the [saved segments summary endpoint](https://developers.activecampaign.com/reference/get_saved_segment_summaries) instead of the generic segments endpoint.

This is the endpoint that reflects what people see in the Segments dashboard in AC.

1. Make sure setting a segment to send the newsletter to shows the right segments with their data (subscriber count, name, id)
2. Make sure when you send a newsletter it goes to the right segment and the campaign created in AC is tied to the correct segment
3. Test the `get_segments` method on WP Shell and see if it works as expected. Manually edit the code and change the default `page_size` to 1 to check if pagination is working.
4. During the same session on CLI, call get_segments passing a search or ids to check that the method behaves correctly with the meomized result.
5. Double check things :) 

### Other information:

### DIFF
--- a/includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign.php
+++ b/includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign.php
@@ -667,8 +667,8 @@ final class Newspack_Newsletters_Active_Campaign extends \Newspack_Newsletters_S
 				return $segments;
 			}
 			foreach ( $segments as $segment ) {
-				$segment_name = ! empty( $segment['name'] ) ?
-					$segment['name'] . ' (ID ' . $segment['id'] . ')' :
+				$segment_name = ! empty( $segment['attributes']['name'] ) ?
+					$segment['attributes']['name'] :
 					sprintf(
 						// Translators: %s is the segment ID.
 						__( 'Untitled %s', 'newspack-newsletters' ),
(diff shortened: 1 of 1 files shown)


includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign.php
```diff
@@ -667,8 +667,8 @@ public function get_send_lists( $args = [], $to_array = false ) {
 				return $segments;
 			}
 			foreach ( $segments as $segment ) {
-				$segment_name = ! empty( $segment['name'] ) ?
-					$segment['name'] . ' (ID ' . $segment['id'] . ')' :
+				$segment_name = ! empty( $segment['attributes']['name'] ) ?
+					$segment['attributes']['name'] :
```
adekbadek, line 671: Non-blocking: the previous code appended ` (ID <id>)` to named segments, which the PR description's test step 1 still references. Without it, two segments sharing a name will look identical in the editor dropdown. Worth restoring.

adekbadek (approved):
Tested on a local site with AC configured. Pagination works correctly (verified by temporarily setting `page_size` default to 1 – all segments still returned), and the memoization path behaves correctly for the real usage pattern: first call unfiltered, subsequent calls with `search`/`ids` filter against the cache without re-hitting the API. Verified with single-id, multi-id, matching search, and non-matching search.

One non-blocking piece of feedback inline on the segment name change.

includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign.php
leogermani: The thing is that the IDs before were simple and beautiful numbers like `51`. The IDs now are very long hash strings like `3acb7bc7-2c2a-4553-90fd-68ae37904e11`, so it breaks the UI

Also, we had the problem of segments with the same name before because we were listing a bunch of things that were not intentional saved segments... these were things like "Advanced search"... We don't have that problem now

PR #2082, opened 2026-04-17
feat(esp): enrich contact field schema for integrations

leogermani:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-newsletters/blob/trunk/.github/CONTRIBUTING.md)?

Rework the ESP service provider's contact-field method so the Newspack Reader Activation ESP integration can auto-configure incoming fields (access rules / segmentation criteria) based on the ESP-reported field type, and so enumerated fields (dropdown / radio / listbox / checkbox) expose their option lists in the admin UI.

- Rename `get_contact_fields()` → `get_contact_fields_for_integrations()` on the ESP service-provider base class to make the method's purpose explicit.
- Expand the returned schema from `{ key }` to a structured payload whose keys mirror Newspack `Incoming_Field` setters: `key`, `name`, `value_type`, `matching_function`, `options`, `description`, `is_access_rule`, `is_segment_criteria`.
- Mailchimp override: inspect merge-field `type` and auto-promote `text / number / date / radio / dropdown`; build `options` from the inline `options.choices` payload for `radio / dropdown`.
- ActiveCampaign override: promote `text / textarea / date / datetime / dropdown / radio / listbox / checkbox`; fetch option lists for enumerated fields via `GET /fields/{id}/options`; keep the existing 5-minute cache; all AC fields use `default` matching.
- Remove stale `var_dump()` calls from both provider overrides.

### Related PR

Paired with [Automattic/newspack-plugin#4676](https://github.com/Automattic/newspack-plugin/pull/4676), which consumes this schema in the ESP integration's `configure_incoming_field()` hook. That PR depends on this one — trunk of newspack-plugin will break against trunk of newspack-newsletters only during the window where one is merged without the other, so please coordinate merges.

1. Start the workspace, `n setup --yes`, and check out this branch.
2. Configure **Mailchimp** as the ESP with a list containing a mix of merge-field types (including a dropdown with choices).
3. Run `n wp eval 'print_r( Newspack_Newsletters_Contacts::get_fields( "<list_id>" ) );'` and verify:
   - Every entry has the new schema keys (`name`, `value_type`, `matching_function`, `options`, `is_access_rule`, `is_segment_criteria`).
   - `dropdown` / `radio` fields return a populated `options` array; other types return `options = []`.
   - `text`, `number`, `date`, `radio`, `dropdown` return both promotion flags `true`; `phone`, `url`, `imageurl`, `birthday`, `zip`, `address` return both `false`.
4. Switch the ESP to **ActiveCampaign** and rerun `n wp eval 'print_r( Newspack_Newsletters_Contacts::get_fields() );'`:
   - `text / textarea / date / datetime / dropdown / radio / listbox / checkbox` fields return both promotion flags `true`; `hidden` and NULL-typed fields return both `false`.
   - Enumerated fields have their `options` array populated from the `/v3/fields/{id}/options` endpoint.
5. With the paired newspack-plugin PR applied, verify that **Audience → Content Gates** surfaces promoted ESP fields as access rules with the right option lists, and that popups segmentation picks up the same options.
6. Confirm `get_contact_fields()` no longer exists on the provider classes (calling it directly now throws a fatal). The only callers (this repo's `Newspack_Newsletters_Contacts::get_fields()` and `newspack-plugin`'s ESP integration) have been updated.

### Other information:

### DIFF
--- a/includes/class-newspack-newsletters-contacts.php
+++ b/includes/class-newspack-newsletters-contacts.php
@@ -408,6 +408,6 @@ class Newspack_Newsletters_Contacts {
 			return new WP_Error( 'newspack_newsletters_invalid_provider', __( 'Provider is not set.' ) );
 		}
 
-		return $provider->get_contact_fields( $list_id );
+		return $provider->get_contact_fields_for_integrations( $list_id );
 	}
 }
--- a/includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign.php
+++ b/includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign.php
@@ -1780,14 +1780,19 @@ final class Newspack_Newsletters_Active_Campaign extends \Newspack_Newsletters_S
 	}
 
 	/**
+	 * Object cache group for the integrations field schema and per-field option lists.
+	 */
+	const INTEGRATIONS_CACHE_GROUP = 'newspack_newsletters_active_campaign';
+
+	/**
 	 * Get contact fields for Newspack integrations.
 	 *
 	 * @param string|null $list_id The List ID (unused — ActiveCampaign contact fields are global).
 	 * @return array|WP_Error
 	 */
 	public function get_contact_fields_for_integrations( $list_id = null ) {
-		$cache_key     = 'active_campaign_contact_fields_for_integrations';
-		$cached_fields = wp_cache_get( $cache_key );
+		$cache_key     = $this->integrations_cache_key( 'fields' );
+		$cached_fields = wp_cache_get( $cache_key, self::INTEGRATIONS_CACHE_GROUP );
 		if ( false !== $cached_fields ) {
 			return $cached_fields;
 		}
@@ -1802,17 +1807,44 @@ final class Newspack_Newsletters_Active_Campaign extends \Newspack_Newsletters_S
 				$fields[] = $mapped;
 			}
 		}
-		wp_cache_set( $cache_key, $fields, '', 5 * MINUTE_IN_SECONDS );
+		wp_cache_set( $cache_key, $fields, self::INTEGRATIONS_CACHE_GROUP, 5 * MINUTE_IN_SECONDS );
 		return $fields;
+	}
+
+	/**
+	 * Build a cache key for integrations data, namespaced by the configured AC account URL.
+	 *
+	 * Two sites sharing an object cache backend can be configured against different AC accounts;
+	 * a flat global key would let one site's cached schema leak into the other.
+	 *
+	 * @param string $suffix Per-call key suffix (e.g. 'fields', 'options:34').
+	 * @return string
+	 */
+	private function integrations_cache_key( $suffix ) {
+		$credentials = $this->api_credentials();
+		if ( ! empty( $credentials['url'] ) ) {
+			$account_hash = substr( md5( (string) $credentials['url'] ), 0, 12 );
+		} else {
+			// Defensive fallback for a code path that bypasses credential checks; salt with
+			// the blog id so two unconfigured sites sharing an object cache don't collide.
+			$account_hash = 'noaccount-' . get_current_blog_id();
+		}
+		return $account_hash . ':' . $suffix;
 	}
 
 	/**
 	 * Map an ActiveCampaign contact field to the Newspack integrations schema.
 	 *
 	 * AC types eligible for access-rule / segmentation defaults: text, textarea, date, datetime,
-	 * dropdown, radio, listbox, checkbox. Hidden and NULL-typed fields are exposed but not
-	 * promoted by default. All fields use the 'default' matching function — AC stores dropdown
-	 * selections as a single string value, so exact-equality matching against a chosen option works.
+	 * dropdown, radio, listbox, checkbox, multiselect. Hidden and NULL-typed fields are exposed
+	 * but not promoted by default.
+	 *
+	 * Matching function depends on selection cardinality. Per AC's Contact Custom Fields API
+	 * Guide, dropdown / radio / listbox are single-selection types (their stored value is the
+	 * raw chosen option), so 'default' (strict equality) matching is correct. Checkbox and
+	 * multiselect are multi-selection types: AC stores the chosen options with a `||` delimiter
+	 * (e.g. `||Option A||Option C||`), which `default` matching cannot resolve — those types
+	 * use 'list__in', and the consumer's parse_list_value() recognizes the delimiter.
 	 *
 	 * @param array $field Raw field from the ActiveCampaign v3 /fields endpoint.
 	 * @return array|null Mapped field, or null if no usable identifier is available.
@@ -1823,10 +1855,13 @@ final class Newspack_Newsletters_Active_Campaign extends \Newspack_Newsletters_S
 			return null;
 		}
 
-		$type                   = isset( $field['type'] ) ? $field['type'] : 'text';
-		$enumerated_types       = [ 'dropdown', 'radio', 'listbox', 'checkbox' ];
-		$eligible_types         = array_merge( [ 'text', 'textarea', 'date', 'datetime' ], $enumerated_types );
-		$is_promoted_by_default = in_array( $type, $eligible_types, true );
+		$type                       = isset( $field['type'] ) ? $field['type'] : 'text';
+		$single_select_enum_types   = [ 'dropdown', 'radio', 'listbox' ];
+		$multi_select_enum_types    = [ 'checkbox', 'multiselect' ];
+		$enumerated_types           = array_merge( $single_select_enum_types, $multi_select_enum_types );
+		$eligible_types             = array_merge( [ 'text', 'textarea', 'date', 'datetime' ], $enumerated_types );
+		$is_promoted_by_default     = in_array( $type, $eligible_types, true );
+		$is_multi_select            = in_array( $type, $multi_select_enum_types, true );
 
 		$options = [];
 		if ( in_array( $type, $enumerated_types, true ) && ! empty( $field['id'] ) ) {
@@ -1837,7 +1872,7 @@ final class Newspack_Newsletters_Active_Campaign extends \Newspack_Newsletters_S
 			'key'                 => $perstag,
 			'name'                => ! empty( $field['title'] ) ? $field['title'] : $perstag,
 			'value_type'          => 'string',
-			'matching_function'   => 'default',
+			'matching_function'   => $is_multi_select ? 'list__in' : 'default',
 			'options'             => $options,
 			'description'         => ! empty( $field['descript'] ) ? $field['descript'] : '',
 			'is_access_rule'      => $is_promoted_by_default,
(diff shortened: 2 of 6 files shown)


includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php
```diff
@@ -2292,17 +2287,50 @@ public function get_contact_fields( $list_id = null ) {
...
+					'value' => $choice,
+					'label' => $choice,
+				];
+			}
+		}
+
+		return [
+			'key'                 => $field['name'],
+			'name'                => ! empty( $field['name'] ) ? $field['name'] : $field['tag'],
+			'value_type'          => 'string',
```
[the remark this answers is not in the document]
leogermani: Addressed in e7e78c622b8ecbd57ad2ff86b7ba39c6e9631f25 — `key` now uses Mailchimp's merge field `tag` (the stable machine identifier), with `name` as the human label.

includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php
```diff
@@ -2292,17 +2287,50 @@ public function get_contact_fields( $list_id = null ) {
...
+				$options[] = [
+					'value' => $choice,
+					'label' => $choice,
+				];
+			}
+		}
+
+		return [
+			'key'                 => $field['name'],
+			'name'                => ! empty( $field['name'] ) ? $field['name'] : $field['tag'],
```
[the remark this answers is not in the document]
leogermani: Addressed in e7e78c622b8ecbd57ad2ff86b7ba39c6e9631f25 — fields without a usable identifier (`tag`) are now skipped, and `name` falls back to `tag` if the human label is missing.

includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign.php
```diff
@@ -1801,11 +1797,69 @@ public function get_contact_fields( $list_id = null ) {
...
+
+		$options = [];
+		if ( in_array( $type, $enumerated_types, true ) && ! empty( $field['id'] ) ) {
+			$options = $this->fetch_field_options( $field['id'] );
+		}
+
+		$name = ! empty( $field['title'] ) ? $field['title'] : ( $field['perstag'] ?? '' );
+
+		return [
+			'key'                 => $field['title'],
```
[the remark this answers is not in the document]
leogermani: Addressed in e7e78c622b8ecbd57ad2ff86b7ba39c6e9631f25 — `key` now uses ActiveCampaign's `perstag` (the stable machine identifier), with `title` as the human label. Fields without a `perstag` are skipped.

includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign.php
```diff
@@ -1801,11 +1797,69 @@ public function get_contact_fields( $list_id = null ) {
...
+	private function map_contact_field_to_integration_schema( $field ) {
+		$type                   = isset( $field['type'] ) ? $field['type'] : 'text';
+		$enumerated_types       = [ 'dropdown', 'radio', 'listbox', 'checkbox' ];
+		$eligible_types         = array_merge( [ 'text', 'textarea', 'date', 'datetime' ], $enumerated_types );
+		$is_promoted_by_default = in_array( $type, $eligible_types, true );
+
+		$options = [];
+		if ( in_array( $type, $enumerated_types, true ) && ! empty( $field['id'] ) ) {
+			$options = $this->fetch_field_options( $field['id'] );
+		}
```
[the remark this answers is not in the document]
leogermani: Acknowledged. AC accounts in our typical deployments have a small number of custom fields (handful, not hundreds), and the result is cached for 5 minutes, so the per-refresh cost is bounded. The AC v3 `/fields` endpoint doesn't expose options inline, and there's no documented bulk endpoint for field options. Will revisit if we see real-world refreshes hitting rate limits.

includes/class-newspack-newsletters-contacts.php
```diff
@@ -408,6 +408,6 @@ public static function get_fields( $list_id = null ) {
 			return new WP_Error( 'newspack_newsletters_invalid_provider', __( 'Provider is not set.' ) );
 		}
 
-		return $provider->get_contact_fields( $list_id );
+		return $provider->get_contact_fields_for_integrations( $list_id );
```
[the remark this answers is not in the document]
leogermani: Acknowledged. The schema contract is owned by newspack-plugin (the consumer), so the integration tests will live there alongside the consumer. Adding provider-level tests for the mapping is reasonable as a follow-up but out of scope for this PR.

includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign.php
```diff
@@ -1801,11 +1797,75 @@ public function get_contact_fields( $list_id = null ) {
...
+
+		$type                   = isset( $field['type'] ) ? $field['type'] : 'text';
+		$enumerated_types       = [ 'dropdown', 'radio', 'listbox', 'checkbox' ];
+		$eligible_types         = array_merge( [ 'text', 'textarea', 'date', 'datetime' ], $enumerated_types );
+		$is_promoted_by_default = in_array( $type, $eligible_types, true );
+
+		$options = [];
+		if ( in_array( $type, $enumerated_types, true ) && ! empty( $field['id'] ) ) {
+			$options = $this->fetch_field_options( $field['id'] );
```
adekbadek, line 1868: Each enumerated field triggers its own `/fields/{id}/options` call here. With many enumerated fields this becomes 1 + N synchronous HTTP calls on every cold cache (each call up to the 45s `api_v3_request` timeout). Consider caching each options call independently with a longer TTL than the parent so a 5-minute outer-cache miss doesn't re-fetch every option list, or check whether AC's `/fields` endpoint can return options inline.

includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign.php
```diff
@@ -1801,11 +1797,75 @@ public function get_contact_fields( $list_id = null ) {
...
+		}
+
+		return [
+			'key'                 => $perstag,
+			'name'                => ! empty( $field['title'] ) ? $field['title'] : $perstag,
+			'value_type'          => 'string',
+			'matching_function'   => 'default',
```
adekbadek: **Updated based on AC's official docs.** Per the [Contact Custom Fields API Guide](https://developers.activecampaign.com/reference/contact-custom-fields-api-guide), AC's field type taxonomy is:

- `dropdown` — single selection
- `radio` — single selection
- `listbox` — single selection (the multi-select example further down on the same page contradicts this — looks like a doc error)
- `checkbox` — **multiple selections**, stored with `||val1||val2||` delimiter (e.g. `value: "||Option A||Option C||"`)
- AC also has a `multiselect` type the PR doesn't expose at all.

So the correctness concern is narrower than I first claimed: `default` matching is fine for `dropdown`, `radio`, and `listbox`, but breaks for `checkbox` (and would also break for `multiselect` if added). The plugin-side `default` matcher is strict `===` (`class-promoted-fields.php:246`), and `||Option A||Option C|| === "Option A"` is always false.

Suggested fix: split `$enumerated_types` into single-select (current `default` matching is correct) and multi-select. Multi-select types should emit `matching_function = 'list__in'`, and the plugin-side `Promoted_Fields::parse_list_value()` needs to learn the `||` delimiter — today it returns `[ $value ]` for any string. Adding `multiselect` to the multi-select set is also worth doing for coverage.

includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign.php
```diff
@@ -1780,17 +1780,13 @@ public function get_usage_report() {
...
-	 *
-	 * @param string|null $list_id The List ID. Optional, as some providers might not have different fields per list.
-	 * @return array|WP_Error The contact fields for the list. Each field should be an array with 'key' key at least. WP_Error if the request to fetch the fields failed.
+	 * @param string|null $list_id The List ID (unused — ActiveCampaign contact fields are global).
+	 * @return array|WP_Error
 	 */
-	public function get_contact_fields( $list_id = null ) {
-		$cache_key = 'active_campaign_contact_fields';
+	public function get_contact_fields_for_integrations( $list_id = null ) {
+		$cache_key     = 'active_campaign_contact_fields_for_integrations';
```
adekbadek: The cache key sits in the default object-cache group with no AC-account namespacing. Worth namespacing it (or moving to a dedicated group), especially in network/multi-site contexts where two sites might point at different AC accounts and share an object cache.

includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign.php
```diff
@@ -1801,11 +1797,75 @@ public function get_contact_fields( $list_id = null ) {
...
+
+	/**
+	 * Fetch the option list for an enumerated ActiveCampaign field.
+	 *
+	 * @param int|string $field_id The AC field ID.
+	 * @return array Array of [ 'value' => ..., 'label' => ... ] pairs (empty on failure).
+	 */
+	private function fetch_field_options( $field_id ) {
+		$response = $this->api_v3_request( 'fields/' . rawurlencode( (string) $field_id ) . '/options', 'GET' );
+		if ( is_wp_error( $response ) || empty( $response['fieldOptions'] ) ) {
```
adekbadek: An API failure here returns `[]`, indistinguishable from a real empty list, and that empty result then sits in the parent 5-minute cache. At minimum, log the WP_Error so a developer investigating a missing-options report has something to trace.

includes/service-providers/class-newspack-newsletters-service-provider.php
```diff
@@ -941,16 +941,28 @@ public function get_transient_name( $post_id ) {
...
+	 * Reader Activation incoming fields (content gate access rules and popups segmentation criteria).
 	 *
-	 * This is used by Newspack integrations to sync contact data.
+	 * Each returned entry is an associative array with the following keys. All keys except 'key' are
+	 * optional — defaults kick in from the Incoming_Field class when a key is missing.
+	 *
+	 *   - key (string, required): Machine key for the field. Used as the Incoming_Field key.
+	 *   - name (string): Human-readable label. Defaults to the key.
+	 *   - value_type (string): 'string' or 'boolean'. Defaults to 'string'.
+	 *   - matching_function (string): 'default' | 'list__in' | 'list__not_in' | 'range'. Defaults to 'default'.
```
adekbadek: The docblock advertises `value_type ∈ {string, boolean}` and `matching_function ∈ {default, list__in, list__not_in, range}`, but the providers only emit `string` and `default` today. Worth either trimming this to what's actually emitted, or adding a note that the wider enums are forward-looking.

adekbadek (changes requested):
Reviewed alongside the paired plugin PR. The rename is contained, the schema expansion is well thought out, and consumer alignment with the plugin side is clean — every key emitted here maps 1:1 to an `Incoming_Field` setter on the consumer.

Two things I'd want resolved before merge, plus a few smaller suggestions inline:

1. **AC option fetch is N+1 on every cold cache.** `fetch_field_options()` is called per enumerated field with no per-call cache, so a list with 30 dropdowns means 30 sequential AC HTTP calls every 5 minutes (each up to a 45s timeout). Either cache each `/fields/{id}/options` independently with a longer TTL, batch the fetch, or document that this method must not be called in a synchronous request path.
2. **`checkbox` and `listbox` are promoted but mapped to `default` (exact-match) matching.** AC stores multi-select values with `||val1||val2||` wrapping (and even a single selection looks like `||val||`), so strict equality against one option string can never match. Detail in the inline.

The other items (no tests in this repo for the new mappers, schema docblock vs. emitted values, raw ESP labels passing through unsanitized) are below as inline comments.

Also — the PR description mentions removing stale `var_dump()` calls, but I don't see any `var_dump` removal in the diff. Worth trimming the description.

includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign.php
adekbadek: Addressed in 8208b06c. `fetch_field_options()` now caches each field's option list independently (1h TTL inside the `newspack_newsletters_active_campaign` group), so the parent's 5-min cache miss only re-fetches roster-side changes. Picking up this work while @leogermani is unavailable.

includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign.php
adekbadek: Addressed in 8208b06c. Split `$enumerated_types` into single-select (`dropdown`/`radio`/`listbox` → `default` matching) and multi-select (`checkbox`/`multiselect` → `list__in`). The companion plugin PR #4676 extends `Promoted_Fields::parse_list_value()` to recognize the `||` delimiter.

includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign.php
adekbadek: Addressed in 8208b06c. Introduced an `INTEGRATIONS_CACHE_GROUP` constant and an `integrations_cache_key()` helper that hashes the AC account URL into the key, with a blog-id-salted fallback for the no-credentials path.

includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign.php
adekbadek: Addressed in 8208b06c. WP_Error path now logs via `Newspack_Newsletters_Logger::log()` with the failing field id and the error message. The empty-but-successful response still caches `[]` for the full TTL.

includes/service-providers/class-newspack-newsletters-service-provider.php
adekbadek: Addressed in 8208b06c. Trimmed the docblock to what providers emit today, with notes that `list__not_in`/`range` are forward-looking on the consumer side.

adekbadek (approved):
All inline items addressed (resolved threads above). I picked up the remaining work since @leogermani is out — split single/multi-select matching per AC's Contact Custom Fields API Guide, per-field options cache with account-namespaced keys, error logging on the failure path, and fixture-driven mapper tests for both Mailchimp and ActiveCampaign.

Re-tested the suite locally: 194 / 194 pass, 686 assertions. CI on the fix commit is also green.

Approving so the coordinated merge with #4676 can proceed.

PR #2116, opened 2026-05-08
fix: remove block-level CSS field from Newsletter editor

laurelfulford:
### All Submissions:

WordPress 7.0 introduces a block-level CSS field under the 'Advanced' panel in the block editor. This CSS seems to be stripped out on sent emails, so this hotfix removes it from the newsletter editor. 

It also updates the code removing the 'Hide' option from newsletters so it's blocked from the Newsletter Ads CPT, too.

Closes NEWS-2203

1. Apply this PR to a test site running the latest version of the WP 7.0 RC.
2. Create a newsletter, and add some blocks.
3. Click on a block, and then open the 'Advanced' panel in the right sidebar. Confirm that there isn't a "Additional CSS" text field.
4. Using the block toolbar, click the `...` menu item, and confirm that 'Hide' is not available as an option.
5. Repeat steps 2-4 with a Newsletter Ads post type.
6. Repeat step 2-4 on a regular post or page; confirm that there is still an "Additional CSS" text field in the Advanced panel in other CPTs, and that blocks still have the 'Hide' option.

### Other information:

### DIFF
--- a/src/editor/index.js
+++ b/src/editor/index.js
@@ -64,7 +64,9 @@ addFilter( 'blocks.registerBlockType', 'newspack-newsletters/core-blocks', ( set
 		settings.supports = { ...settings.supports, align: [ 'full' ] };
 	}
 
-	/* Remove 'Hide' and 'Custom CSS' options for all blocks for Newsletters CPTs, Newsletter Ads */
+	/* This bundle is only enqueued in the email editor (see
+	 * Newspack_Newsletters_Editor::enqueue_block_editor_assets), so disabling
+	 * these block supports applies to the newsletter and newsletter-ad CPTs only. */
 	settings.supports = { ...settings.supports, customCSS: false, visibility: false };
 
 	return settings;
--- a/src/editor/index.js
+++ b/src/editor/index.js
@@ -66,6 +66,9 @@ addFilter( 'blocks.registerBlockType', 'newspack-newsletters/core-blocks', ( set
 		settings.supports = { ...settings.supports, align: [ 'full' ] };
 	}
 
+	/* Remove block-level Custom CSS field; CSS added there does not work in sent newsletters. */
+	settings.supports = { ...settings.supports, customCSS: false };
+
 	/* Remove 'Hide' option only for the newsletter CPT. */
 	if ( isEditingNewsletterCpt ) {
 		settings.supports = { ...settings.supports, visibility: false };
--- a/src/editor/index.js
+++ b/src/editor/index.js
@@ -23,8 +23,6 @@ import { addBlocksValidationFilter } from './blocks-validation/blocks-filters';
 import { NestedColumnsDetection } from './blocks-validation/nesting-detection';
 import MJML from './mjml';
 
-const isEditingNewsletterCpt = newspack_email_editor_data.newsletter_post_type === newspack_email_editor_data.current_post_type;
-
 addBlocksValidationFilter();
 registerAdBlock();
 registerPostsInserterBlock();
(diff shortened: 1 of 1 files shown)


src/editor/index.js
```diff
@@ -66,6 +66,9 @@ addFilter( 'blocks.registerBlockType', 'newspack-newsletters/core-blocks', ( set
 		settings.supports = { ...settings.supports, align: [ 'full' ] };
 	}
 
+	/* Remove block-level Custom CSS field; CSS added there does not work in sent newsletters. */
+	settings.supports = { ...settings.supports, customCSS: false };
+
 	/* Remove 'Hide' option only for the newsletter CPT. */
 	if ( isEditingNewsletterCpt ) {
 		settings.supports = { ...settings.supports, visibility: false };
```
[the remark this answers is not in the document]
laurelfulford: Good catch! Fixed in [ef611b3](https://github.com/Automattic/newspack-newsletters/pull/2116/commits/ef611b324d3a356dd926f371e4e5e4dc9c5c3df6) -- neither of these options will work for Newspack Ads CPTs either, so they're hidden there now, too.

src/editor/index.js
```diff
@@ -66,10 +64,8 @@ addFilter( 'blocks.registerBlockType', 'newspack-newsletters/core-blocks', ( set
 		settings.supports = { ...settings.supports, align: [ 'full' ] };
 	}
 
-	/* Remove 'Hide' option only for the newsletter CPT. */
-	if ( isEditingNewsletterCpt ) {
-		settings.supports = { ...settings.supports, visibility: false };
-	}
+	/* Remove 'Hide' and 'Custom CSS' options for all blocks for Newsletters CPTs, Newsletter Ads */
```
adekbadek: **Suggestion:** Make the implicit CPT gating explicit in the comment — The previous code contained a JS-side `if ( isEditingNewsletterCpt )` check; that's now gone, but the comment still talks about *which* CPTs the filter applies to. A future maintainer reading just this file may wonder where that gating happens — the answer is the server-side enqueue gate in `Newspack_Newsletters_Editor::enqueue_block_editor_assets`, which only loads this bundle on the newsletter and newsletter-ad CPTs. Worth a one-liner pointing at that.

```js
/* This bundle is only enqueued in the email editor (see
 * Newspack_Newsletters_Editor::enqueue_block_editor_assets), so disabling
 * these block supports applies to the newsletter and newsletter-ad CPTs only. */
settings.supports = { ...settings.supports, customCSS: false, visibility: false };
```

adekbadek (approved):
Approve. Tiny, well-scoped hotfix; the support keys (`customCSS`, `visibility`) match WP 7.0-RC1, and the bundle's server-side enqueue gate (`is_editing_email()`) keeps the filter from leaking into regular post, page, or site-editor contexts — verified against all three. Only minor nits on the in-code comment.

**Follow-up (out of scope):** with `isEditingNewsletterCpt` gone, the only consumers of `newspack_email_editor_data.newsletter_post_type` and `current_post_type` (quick grep across `src/` finds none) look unused — worth a tidy-up pass after the hotfix lands to drop them from the localized payload in `Newspack_Newsletters_Editor::enqueue_block_editor_assets` (`includes/class-newspack-newsletters-editor.php`).

src/editor/index.js
laurelfulford: Thanks @adekbadek! I updated the comment in [7193ccd](https://github.com/Automattic/newspack-newsletters/pull/2116/commits/7193ccd16cebc3e3317eb754de9267877414bd8e).

laurelfulford: Also created an issue for following up on the out-of-scope bit here: NPPM-2839

PR #2131, opened 2026-05-13
fix(esp): abort send when configured segment cannot be resolved

adekbadek:
### All Submissions:

Closes the catastrophic-by-default failure mode that turned the recent #2079 hotfix regression into a downstream incident.

When a newsletter post had `send_sublist_id` set but the saved segment/sublist could no longer be resolved against the ESP (e.g. after an endpoint change, a deleted segment, or a transient API error), both providers silently fell through to a list-only payload:

- **ActiveCampaign**: `'segmentid' => $send_sublist_id ?? 0` passed an unresolvable ID straight through, and AC interprets `segmentid=0` as "send to the entire parent audience."
- **Mailchimp**: when the sublist lookup returned empty / no entity_type, the switch in `get_sync_payload()` was silently skipped, leaving `recipients = { list_id }` with no `segment_opts` (also "send to entire list").

Combined with the irreversibility of email sends, a routine upstream regression was enough to mail an entire audience by accident.

This PR validates the configured segment/sublist before the campaign is submitted and aborts with `WP_Error` if the lookup errors or returns empty. An *explicitly* empty `send_sublist_id` (no segment was ever picked) is still treated as an intentional whole-list send, so existing whole-list workflows are unaffected.

Error codes added:

- `newspack_newsletters_active_campaign_segment_lookup_failed`
- `newspack_newsletters_active_campaign_segment_not_found`
- `newspack_newsletters_mailchimp_sublist_lookup_failed`
- `newspack_newsletters_mailchimp_sublist_not_found`
- `newspack_newsletters_mailchimp_sublist_unknown_type` (defensive: covers unexpected entity_type values returned by the lookup)

The `WP_Error` propagates through the existing `send()` → `send_newsletter()` flow, so failures land in the `newsletter_send_errors` meta, trigger the admin failure email, and `wp_die` the editor request (or trigger the scheduled-send retry with exponential backoff). No new wiring.

Prereqs: ActiveCampaign and/or Mailchimp credentials configured for the active provider, with at least one list/audience and at least one segment/sublist available.

**ActiveCampaign — abort path:**

1. Switch the active provider to ActiveCampaign.
2. Create a draft newsletter post; in the editor pick a list and a segment, save.
3. Via WP-CLI: `wp post meta update <ID> send_sublist_id 99999999` (overwrite with a bogus ID).
4. Attempt to publish (status → `private` or `publish`).
5. Expected: send is aborted, no campaign is created in AC, the failure surfaces in the editor with the message "The selected segment could not be found in ActiveCampaign. Sending was aborted to avoid sending to the entire audience. Please re-select a segment and try again." For scheduled sends the post drops back to `future` with exponential backoff (existing behavior).

**ActiveCampaign — happy paths:**

6. With a real segment ID set as `send_sublist_id`, publish. Expected: campaign is created in AC with the segment applied, no abort.
7. With `send_sublist_id` empty / absent, publish. Expected: campaign is created with `segmentid=0` (intentional whole-list send), no abort.

**Mailchimp — abort path:**

8. Switch the active provider to Mailchimp.
9. Create a draft newsletter post; pick a list and a group/tag/segment, save.
10. Via WP-CLI: `wp post meta update <ID> send_sublist_id BOGUS_ID`.
11. Trigger sync (e.g. open the editor or call `get_sync_payload` via `wp eval`). Expected: sync errors with "The selected sublist could not be found in Mailchimp. Sending was aborted to avoid sending to the entire audience..." surfaced via the existing sync-error transient.

**Mailchimp — happy paths:**

12. With a real group ID as `send_sublist_id`, sync. Expected: `recipients.segment_opts.conditions` contains the `Interests` condition.
13. With a real tag ID as `send_sublist_id`, sync. Expected: `recipients.segment_opts.conditions` contains the `StaticSegment` condition.
14. With a real saved-segment ID as `send_sublist_id`, sync. Expected: `recipients.segment_opts` is populated from `fetch_segment( ... )['options']`.
15. With no `send_sublist_id`, sync. Expected: `recipients = { list_id, segment_opts: {} }` (intentional whole-list send), no abort.

**Regression:**

16. `npm run lint:php` and `n test-php` from the plugin directory should pass (no test changes; verifying the existing 198-test suite is unaffected).

### Other information:

* [ ] Have you written new tests for your changes, as applicable? — Existing PHPUnit suite continues to pass. The new code paths were validated end-to-end against real AC and MC sandboxes (both providers, abort + happy paths) on a staging site; a dedicated PHPUnit test for the abort path would require mocking `get_send_lists` and is left as a follow-up if desired.

### DIFF
--- a/includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign.php
+++ b/includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign.php
@@ -1181,6 +1181,37 @@ final class Newspack_Newsletters_Active_Campaign extends \Newspack_Newsletters_S
 		$send_list_id    = get_post_meta( $post->ID, 'send_list_id', true );
 		$send_sublist_id = get_post_meta( $post->ID, 'send_sublist_id', true );
 
+		// A configured-but-unresolvable segment must NOT silently fall through
+		// to "no segment" — AC interprets segmentid=0 as "send to the entire
+		// parent audience", and sent email cannot be unsent. Verify the segment
+		// resolves before submitting; only an explicitly empty send_sublist_id
+		// (no segment ever picked) is treated as an intentional whole-list send.
+		if ( ! empty( $send_sublist_id ) ) {
+			$segment_check = $this->get_send_lists(
+				[
+					'type'      => 'sublist',
+					'ids'       => [ $send_sublist_id ],
+					'parent_id' => $send_list_id,
+				]
+			);
+			if ( is_wp_error( $segment_check ) ) {
+				return new \WP_Error(
+					'newspack_newsletters_active_campaign_segment_lookup_failed',
+					sprintf(
+						// Translators: %s is the upstream error message from ActiveCampaign.
+						__( 'Could not verify the selected segment with ActiveCampaign (%s). Sending was aborted to avoid sending to the entire audience.', 'newspack-newsletters' ),
+						$segment_check->get_error_message()
+					)
+				);
+			}
+			if ( empty( $segment_check ) ) {
+				return new \WP_Error(
+					'newspack_newsletters_active_campaign_segment_not_found',
+					__( 'The selected segment could not be found in ActiveCampaign. Sending was aborted to avoid sending to the entire audience. Please re-select a segment and try again.', 'newspack-newsletters' )
+				);
+			}
+		}
+
 		$is_public = get_post_meta( $post->ID, 'is_public', true );
 		if ( empty( $campaign_name ) ) {
 			$campaign_name = $this->get_campaign_name( $post );
@@ -1192,7 +1223,7 @@ final class Newspack_Newsletters_Active_Campaign extends \Newspack_Newsletters_S
 			'name'                                  => $campaign_name,
 			'fromname'                              => $from_name,
 			'fromemail'                             => $from_email,
-			'segmentid'                             => $send_sublist_id ?? 0, // 0 = No segment.
+			'segmentid'                             => ! empty( $send_sublist_id ) ? $send_sublist_id : 0, // 0 = No segment.
 			'p[' . $send_list_id . ']'              => $send_list_id,
 			'm[' . $sync_result['message_id'] . ']' => 100, // 100 = 100% of contacts will receive this.
 			'addressid'                             => $this->get_address_id(),
--- a/includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign.php
+++ b/includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign.php
@@ -998,7 +998,7 @@ final class Newspack_Newsletters_Active_Campaign extends \Newspack_Newsletters_S
 		}
 		/** Create disposable campaign for sending a test. */
 		$campaign_name = sprintf( 'Test for %s', $this->get_campaign_name( $post ) );
-		$campaign      = $this->create_campaign( get_post( $post_id ), $campaign_name );
+		$campaign      = $this->create_campaign( get_post( $post_id ), $campaign_name, true );
 		if ( is_wp_error( $campaign ) ) {
 			return $campaign;
 		}
@@ -1156,10 +1156,11 @@ final class Newspack_Newsletters_Active_Campaign extends \Newspack_Newsletters_S
 	 *
 	 * @param WP_Post $post          Post to create campaign for.
 	 * @param string  $campaign_name Optional custom title for this campaign.
+	 * @param bool    $is_test       Whether this campaign is a disposable test send. Test sends deliver via `action=test` with explicit recipient emails, so the campaign's segmentid is never used — segment validation is skipped to let publishers test newsletters that reference a deleted-but-still-saved segment.
 	 *
 	 * @return array|WP_Error Campaign data or error.
 	 */
-	private function create_campaign( $post, $campaign_name = '' ) {
+	private function create_campaign( $post, $campaign_name = '', $is_test = false ) {
 		$sync_result = $this->sync( $post );
 		if ( is_wp_error( $sync_result ) ) {
 			return $sync_result;
@@ -1184,9 +1185,20 @@ final class Newspack_Newsletters_Active_Campaign extends \Newspack_Newsletters_S
 		// A configured-but-unresolvable segment must NOT silently fall through
 		// to "no segment" — AC interprets segmentid=0 as "send to the entire
 		// parent audience", and sent email cannot be unsent. Verify the segment
-		// resolves before submitting; only an explicitly empty send_sublist_id
-		// (no segment ever picked) is treated as an intentional whole-list send.
-		if ( ! empty( $send_sublist_id ) ) {
+		// resolves before submitting; only an explicitly unset send_sublist_id
+		// (null or '' — no segment ever picked) is treated as an intentional
+		// whole-list send. A literal "0" is treated as configured-but-invalid
+		// rather than as "no segment", since AC segment IDs are positive
+		// integers and zero is the sentinel for the whole-audience case.
+		$has_configured_segment = ! $is_test && null !== $send_sublist_id && '' !== $send_sublist_id;
+		if ( $has_configured_segment ) {
+			// Note: AC segments are global per account, not scoped to a parent
+			// list. `get_send_lists()` accepts `parent_id` but only echoes it
+			// back on the returned `Send_List` — it does NOT filter the AC
+			// `audiences` lookup. A segment that belongs to a different list
+			// will therefore still resolve here. Cross-list mismatches are an
+			// inherent AC limitation rather than something this guard can
+			// catch.
 			$segment_check = $this->get_send_lists(
 				[
 					'type'      => 'sublist',
@@ -1223,7 +1235,7 @@ final class Newspack_Newsletters_Active_Campaign extends \Newspack_Newsletters_S
 			'name'                                  => $campaign_name,
 			'fromname'                              => $from_name,
 			'fromemail'                             => $from_email,
-			'segmentid'                             => ! empty( $send_sublist_id ) ? $send_sublist_id : 0, // 0 = No segment.
+			'segmentid'                             => $has_configured_segment ? $send_sublist_id : 0, // 0 = No segment (intentional whole-list send, or a test send where the segment is irrelevant because delivery is via `action=test` with explicit recipients).
 			'p[' . $send_list_id . ']'              => $send_list_id,
 			'm[' . $sync_result['message_id'] . ']' => 100, // 100 = 100% of contacts will receive this.
 			'addressid'                             => $this->get_address_id(),
--- a/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php
+++ b/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php
@@ -1054,7 +1054,20 @@ final class Newspack_Newsletters_Mailchimp extends \Newspack_Newsletters_Service
 				'list_id' => $send_list_id,
 			];
 			$send_sublist_id = get_post_meta( $post->ID, 'send_sublist_id', true );
-			if ( ! empty( $send_sublist_id ) ) {
+			// Only an explicitly unset send_sublist_id (null or '' — no sublist
+			// ever picked) is treated as an intentional whole-list send. Any
+			// other value, including a literal "0", is treated as
+			// configured-and-must-resolve so a garbage value can't quietly fall
+			// through to the whole audience.
+			if ( null !== $send_sublist_id && '' !== $send_sublist_id ) {
+				// Note: Mailchimp groups and tags are looked up via
+				// Newspack_Newsletters_Mailchimp_Cached_Data, which refreshes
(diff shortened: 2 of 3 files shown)


includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php
```diff
@@ -1063,47 +1063,74 @@ public function get_sync_payload( $post ) {
...
+						break;
+					case 'segment':
+						$segment_data = Newspack_Newsletters_Mailchimp_Cached_Data::fetch_segment( $send_sublist_id, $send_list_id );
+						if ( is_wp_error( $segment_data ) ) {
+							return $segment_data;
+						}
+						if ( ! empty( $segment_data['options'] ) ) {
+							$payload['recipients']['segment_opts'] = $segment_data['options'];
+						} else {
+							return new WP_Error( 'newspack_newsletters_mailchimp_error', __( 'Could not fetch segment criteria for segment ', 'newspack-newsletters' ) . $sublist['name'] );
```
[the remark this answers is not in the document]
adekbadek: Good catch — that was a pre-existing bug at the segment-criteria-missing fallthrough. Fixed in ddeb879e by switching to `$sublist[0]->get_name()` and using `sprintf` with a placeholder (plus terminating the sentence with a period instead of a trailing space).

includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php
```diff
@@ -1063,47 +1063,74 @@ public function get_sync_payload( $post ) {
...
+							__( 'Could not verify the selected sublist with Mailchimp (%s). Sending was aborted to avoid sending to the entire audience.', 'newspack-newsletters' ),
+							$sublist->get_error_message()
+						)
+					);
+				}
+				if ( empty( $sublist ) || empty( $sublist[0]->get_entity_type() ) ) {
+					return new WP_Error(
+						'newspack_newsletters_mailchimp_sublist_not_found',
+						__( 'The selected sublist could not be found in Mailchimp. Sending was aborted to avoid sending to the entire audience. Please re-select a sublist and try again.', 'newspack-newsletters' )
+					);
```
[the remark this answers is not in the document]
adekbadek: Agreed. Fixed in ddeb879e by moving the `is_wp_error($payload)` short-circuit ahead of `apply_filters('newspack_newsletters_mc_payload_sync', ...)`, so third-party callbacks always receive the array shape their hook contract documents. Also updated `get_sync_payload()`'s return docblock to `array|WP_Error`.

includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php
dkoo: ### 🟡 Suggestion: `"0"` / numeric-string coercion bypasses the new guard

Both providers use `! empty( $send_sublist_id )` to decide whether to validate. The values `"0"`, `0`, `false`, `null`, `''` all skip validation and fall through to the whole-list payload.

- AC: `class-newspack-newsletters-active-campaign.php:1189` → whole-list payload at `class-newspack-newsletters-active-campaign.php:1226`
- MC: `class-newspack-newsletters-mailchimp.php:1057` → empty `segment_opts` at `class-newspack-newsletters-mailchimp.php:1142`

The use of `empty()` conflates two different states:

- **"Never picked a segment"** → `null` from `get_post_meta` when the key doesn't exist, or `''` if explicitly cleared by the editor → intentional whole-list (correct)
- **"Configured but garbage value"** → `"0"`, `"false"`, etc. → should be treated like an unresolvable ID and abort (the new code does NOT do this)

#### How likely is this in practice?

There's no evidence that today's code paths produce `"0"` in `send_sublist_id`, so this is **adversarial / defense-in-depth** rather than a known live trigger—hence why this is marked as a non-blocking suggestion.

But it can be easily reproduced by doing the following:

```bash
wp post meta update 123 send_sublist_id 0
wp eval 'var_export( ( Newspack_Newsletters_Mailchimp::instance() )->get_sync_payload( get_post(123) ) );'
```

If the printed payload has `recipients => [ list_id => ..., segment_opts => stdClass {} ]` with **no WP_Error**, the bypass is confirmed live.

**Suggested fix:** Replace `! empty( $send_sublist_id )` with a stricter check that only treats truly-unset values as intentional:
- `'' !== $send_sublist_id && null !== $send_sublist_id` (treats `"0"` as configured)
- Same fix applies symmetrically at `class-newspack-newsletters-active-campaign.php:1189`.

Or, document that the `empty()` check is enough because any falsy value (in addition to `null` or `''`) should be considered equivalent to an intentionally unset value.

dkoo, line 1434: ### 🟡 Suggestion: `audience_segments()` (Mailchimp) has the same silent whole-list bug as `empty( $send_sublist_id )` with a different entry point 

`class-newspack-newsletters-mailchimp.php:1421` initializes `$segment_opts = (object) []` and at L1436 `$mc->get("lists/$list_id/segments/$segment_id")` is used without `validate()`. If neither `$interest_id` nor `$segment_id` resolves to a real Mailchimp object, an empty `segment_opts` is PATCHed to the campaign — same "send to entire audience" condition. This is the REST endpoint for changing audience on an existing campaign (reachable from the editor).

**Suggested fix:** Consider validating that `$interest_id` and `$segment_id` values resolve to real Mailchimp objects and throw an error before sending to the whole audience if they don't.

includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign.php
```diff
@@ -1181,6 +1181,37 @@ private function create_campaign( $post, $campaign_name = '' ) {
...
+		// resolves before submitting; only an explicitly empty send_sublist_id
+		// (no segment ever picked) is treated as an intentional whole-list send.
+		if ( ! empty( $send_sublist_id ) ) {
+			$segment_check = $this->get_send_lists(
+				[
+					'type'      => 'sublist',
+					'ids'       => [ $send_sublist_id ],
+					'parent_id' => $send_list_id,
+				]
+			);
```
dkoo, line 1208: ### 🟡 Suggestion: AC `parent_id` is metadata-only, not a query filter — segment cross-list validation isn't actually enforced

`get_send_lists()` receives `parent_id`, but for `type=sublist` the AC implementation (`class-newspack-newsletters-active-campaign.php:657`) only forwards `ids` and `search` to `get_segments()`. A segment ID belonging to a *different* AC parent list will still pass validation. AC segments are global per-account at the API level, so this may be the best we can do — but either compare `$segment_check[0]->get_parent_id() === $send_list_id` after lookup, or add a comment clarifying the limitation.

includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign.php
dkoo: ### 🔴 Blocker: AC `test()` sends are now blocked by segment validation, but test sends don't broadcast

`create_campaign()` now runs the segment check. AC test sends use explicit recipient emails (L1024-1036), not the segment — so aborting them is overly cautious. Users won't be able to test newsletters with a deleted-but-still-referenced segment. Consider a `$is_test` flag to skip validation in the test path. An edge case, but definitely possible when using a custom layout with references to saved segments.

includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php
```diff
@@ -1063,47 +1063,81 @@ public function get_sync_payload( $post ) {
...
+					return new WP_Error(
+						'newspack_newsletters_mailchimp_sublist_lookup_failed',
+						sprintf(
+							// Translators: %s is the upstream error message from Mailchimp.
+							__( 'Could not verify the selected sublist with Mailchimp (%s). Sending was aborted to avoid sending to the entire audience.', 'newspack-newsletters' ),
+							$sublist->get_error_message()
+						)
+					);
+				}
+				if ( empty( $sublist ) || empty( $sublist[0]->get_entity_type() ) ) {
```
dkoo, line 1093: ### 🟡 Suggestion: Mailchimp validation is cache-backed for group/tag (but not for saved segments)

`get_send_lists()` uses `Newspack_Newsletters_Mailchimp_Cached_Data`, which refreshes cached audience data via cron once every 10 minutes. Only saved-segment lookups get a live `fetch_segment()` call (`class-newspack-newsletters-mailchimp.php:1115`). Maybe unlikely in practice, but a deleted/renamed tag or group could still pass validation from stale cache data if a subsequent send using this deleted object happens before cache refresh. Consider an optional live lookup or cache refresh at send-time, or shorter cache TTL.

dkoo (changes requested):
@adekbadek This is a good change: when attempting to send a campaign with a stale/nonexistent sublist ID, the send aborts with an error for both AC and MC.

One thing I noticed is that this also aborts AC _test_ sends, which could be a minor annoyance for publishers.

Another question: should we consider making similar changes for Campaign Monitor and Constant Contact provider classes too? The usage level is far below AC and MC, but not making the change for the other providers bifurcates the provider behavior.

A few other non-blocking suggestions inline, too—I'm not sure whether `"0"` bypassing the new guard is an actual issue, but wanted to make sure before we proceed.

includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php
adekbadek: Good defense-in-depth catch. Addressed in 88d1a06f on both providers: switched from `! empty( $send_sublist_id )` to `null !== $send_sublist_id && '' !== $send_sublist_id`, so only a truly-unset value counts as an intentional whole-list send. `"0"`, `"false"`, and any other configured-but-garbage value now hit the validation and abort with `*_segment_not_found` / `*_sublist_not_found` instead of falling through.

Verified locally against the AC sandbox and MC sandbox: with `send_sublist_id = "0"` both providers now return `*_segment_not_found` / `*_sublist_not_found`; the truly-empty-meta case still produces an intentional whole-list payload.

adekbadek: Good catch — same failure shape, different entry point. Addressed in 88d1a06f: `audience_segments()` now validates the `$segment_data` shape from `lists/{list_id}/segments/{segment_id}` before constructing `segment_opts`. Two new abort paths:

- `newspack_newsletters_mailchimp_segment_not_found` when the response is empty or has no `type`.
- `newspack_newsletters_mailchimp_segment_unknown_type` when the response has a `type` that isn't `static` or `saved` (defensive `default` arm).

The interest_id branch wasn't changed — it doesn't make a network call, so there's no equivalent failure mode there. End-to-end exercise of this REST entry point wasn't easy from CLI, but the existing 198-test suite still passes and lint is clean.

includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign.php
adekbadek: You're right that comparing `->get_parent_id()` to `$send_list_id` would be meaningless here — `Send_List::parent_id` is set from the `parent_id` arg passed in to `get_send_lists()`, not from any AC-side relationship, so the comparison would always be tautologically true. I went with option (b): added an inline comment in 88d1a06f above the `get_send_lists` call documenting that AC segments are global per account and that the `parent_id` arg is echoed-back metadata, so this guard cannot catch cross-list mismatches by design.

includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign.php
adekbadek: Excellent catch — fixed in 88d1a06f. Added an `$is_test = false` parameter to `create_campaign()`; `test()` now passes `true` so the segment validation is skipped on the test-send path. Also coerced `segmentid` to `0` in that branch so AC doesn't reject the upstream `campaign_create` call when the saved segment is missing — test sends deliver via `action=test` with explicit emails, so the campaign's segmentid is unused.

Verified end-to-end against the AC sandbox: a test send with a bogus `send_sublist_id` (99999999) now succeeds (campaign created, then cleaned up), while a real send with the same bogus value still aborts with `*_segment_not_found`.

includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php
adekbadek: Documented the limitation in 88d1a06f — added an inline comment above the `get_send_lists` call in `get_sync_payload()` noting that groups/tags are looked up via `Newspack_Newsletters_Mailchimp_Cached_Data` (~10 min async TTL) so a stale-cache window exists, while saved segments go through the live `fetch_segment()` call further down and don't have the gap.

Closing the gap properly means an extra live API call per send. I'd rather land that as a small follow-up so this PR stays scoped to the catastrophic-by-default failure mode — happy to open the follow-up issue / PR if you'd like.

dkoo (approved):
@adekbadek thanks for addressing those threads—confirmed that the latest changes plug the remaining "send-to-whole-audience" holes in the Mailchimp class, test AC campaigns no longer trigger segment validation, and acknowledge the updates to CC as well as the decision to ignore CM as a deprecated provider class.

PR #2163, opened 2026-05-21
feat: ship admin UX modernisation milestone to trunk

thomasguillot:
Ships the **Admin UX modernisation** milestone to trunk. First milestone of the newsletter modernisation effort to land under the per-milestone shipping model.

> [!IMPORTANT]
> **Must merge in lockstep** with the companion `newspack-plugin` PR: https://github.com/Automattic/newspack-plugin/pull/4735.
> The two PRs are coupled: the newsletters side ships the new React Settings surface, DataView list screens, and admin shell scaffolding; the newspack-plugin side ships the wizard-bridge wiring and the rebuilt Settings page that hosts it. Either landing first will leave the other repo temporarily broken in production.

## Milestone scope

See the milestone roll-up at https://github.com/Automattic/newspack-newsletters/pull/2141 for the full breakdown of what landed across 18 per-ticket PRs (#2143–#2160). High-level surfaces:

- **Admin shell scaffold** — page transitions, header save bar, snackbar notices, shared layout primitives (NEWS-1927).
- **Newsletters list** — DataView with author avatars, status badges, auto-draft visibility, Author / Categories / Tags / Send-List filters (NEWS-1928, NEWS-2175, NEWS-2197).
- **Quick Edit side panel** — replaces the inline `WP_List_Table` Quick Edit with a Site-Editor-style side panel (NEWS-1932, NEWS-1933).
- **Newsletter Ads list** — DataView for the ads CPT with meta-backed sortable columns (NEWS-1930, NEWS-2196).
- **Advertisers list** — DataView replacing the classic `edit-tags.php` taxonomy admin (NEWS-1951).
- **Layouts** — new Layouts list (DataView grid with previews), authoring editor with send-suppression, `DataViewsPicker` entry point (NEWS-1929, NEWS-1967, NEWS-2198, NEWS-2212, NEWS-2220).
- **Settings + Subscription Lists** — standalone-mode React shell with single header Save, bundled-mode parity, modal-based local-list edit, ESP card grid (NEWS-1931, NEWS-2088, NEWS-2093, NEWS-2152, NEWS-2168).
- **Merge-tag picker** — toolbar button on text blocks, generic `{}` autocomplete trigger, live ESP fetching, ActiveCampaign supported alongside Mailchimp (NEWS-2242, NEWS-2244).
- **Empty states** across newsletters / ads / advertisers (NEWS-1952).

Plus 16 follow-up code-review PRs (#2143 through #2160) that tightened the implementation — utility extractions, defensive guards, REST status codes, layout-id stability, and the milestone-final cleanup pass.

## Scaffolding cleanup

This PR also drops `docs/newsletter-modernisation/CONTEXT.md`. That file is the modernisation scaffolding and lives on `epic/newsletters-modernisation` (the scaffolding parent); future milestone branches inherit it on cut and strip it on their own trunk-promotion PR. Recipe documented in CONTEXT.md § *Branch structure* on the scaffolding parent, and mirrored in each Linear milestone's description.

## Test plan

A milestone-wide regression sweep against a `--block-theme=false` site with `--woocommerce` and `--campaigns`:

1. **Newsletters list** — DataView renders with author / categories / tags / send-list filters; status column shows the right kind icon for sent/scheduled/draft/trash; Quick Edit side panel opens and saves; bulk actions work.
2. **Add new newsletter** — DataViewsPicker shows layouts with previews; selecting a layout creates a draft with the right `template_id`.
3. **Newsletter Ads list** — DataView renders with sortable status / impressions / clicks / date columns.
4. **Advertisers list** — DataView with parent/child support and ad counts; self-parent prevention works.
5. **Layouts list** — grid view with previews; opening a layout enters the editor with the Send action suppressed.
6. **Layout test-send** — single recipient + multi recipient; partial-failure response carries `failed_recipients`.
7. **Settings (bundled mode, with newspack-plugin)** — React shell loads inside the wizard, single header Save, ESP card grid, OAuth flow for Constant Contact, Subscription Lists collapsed by default with modal-based Edit.
8. **Settings (standalone mode)** — `wp-admin/admin.php?page=newspack-newsletters-settings` renders the React Settings shell directly.
9. **Merge-tag picker** — toolbar button on text blocks, generic `{}` autocomplete trigger.
10. **Standalone vs bundled parity** — toggle the `newspack-plugin` active state and confirm both modes render the same surfaces.

## Companion-PR merge order

Merge **this PR first**, then `newspack-plugin#4735`.


includes/admin/class-admin-shell-legacy-redirect.php
```diff
@@ -0,0 +1,142 @@
...
+			$forwarded = '' === $forwarded ? [] : [ 'post_status' => $forwarded ];
+		}
+
+		foreach ( self::FORWARDED_LEGACY_ARGS as $key ) {
+			if ( ! empty( $forwarded[ $key ] ) ) {
+				$args[ $key ] = $forwarded[ $key ];
+			}
+		}
+
+		return add_query_arg( $args, admin_url( 'edit.php' ) );
```
chickenn00dle, line 142: **Suggestion (hardening).** Forwarded `$_GET` values (`s`, `orderby`, `author`, `categories`, `tags`, `newspack_newsletters_send_list_id`, …) are `sanitize_text_field`'d but passed into `add_query_arg()` here without `rawurlencode`. `wp_safe_redirect` constrains the host but not the query string, so an unencoded value containing `&` / `=` could inject extra query params into the redirect target (parameter hijacking). Suggest `rawurlencode`-ing the forwarded values before building `$args`.

includes/admin/class-settings-rest.php
```diff
@@ -0,0 +1,527 @@
...
+	];
+
+	const OAUTH_STATE_CACHE_TTL = 60;
+
+	/**
+	 * Boot hooks.
+	 */
+	public static function init() {
+		add_action( 'rest_api_init', [ __CLASS__, 'register_routes' ] );
+		add_action( 'newspack_newsletters_provider_credentials_changed', [ __CLASS__, 'bust_oauth_cache' ] );
```
chickenn00dle, line 55: **Suggestion (optional).** This handler busts the OAuth/validity cache on `newspack_newsletters_provider_credentials_changed`, but only Constant Contact fires that action — Mailchimp / ActiveCampaign credential changes rely on the explicit `bust_oauth_cache` calls in this controller (lines 136–137 / 169–170). That covers the REST path, but a credential change made via any other path (legacy settings form, WP-CLI, direct `update_option`) won't bust the transient. Low impact (cheap re-verify, short TTL); consider firing `provider_credentials_changed` from the MC/AC `set_api_credentials` too, so all providers are handled uniformly.

src/wizard-bridge/local-list-modal-host.js
```diff
@@ -0,0 +1,103 @@
...
+		const handleConfirmDelete = event => {
+			const list = event.detail?.list;
+			if ( list ) {
+				setDeletePending( list );
+			}
+		};
+		document.addEventListener( EVENTS.OPEN_MODAL, handleOpen );
+		document.addEventListener( EVENTS.OPEN_CONFIRM_DELETE, handleConfirmDelete );
+		// Listeners installed; signal readiness. A sync consumer dispatch on `bridge-mounted` must land here, not before.
+		window.newspackNewslettersBridgeReady = true;
```
chickenn00dle, line 44: **Nit.** The readiness flag is set here but never reset in the effect cleanup (the cleanup at the bottom of this effect removes the listeners but leaves `window.newspackNewslettersBridgeReady === true`). Under a remount (React 18 StrictMode dev double-invoke, or any future remount) there's a narrow window where the flag reads `true` while listeners are detached, so a consumer dispatch in that microtask could be lost. Production single-mount + the `boot()` double-root guard make this low-probability; setting the flag back to `false` in cleanup keeps the flag and listener state consistent.

chickenn00dle (commented):
Focused cross-repo review (companion to Automattic/newspack-plugin#4735), concentrating on the wizard-bridge contract, the REST surface the Settings UI calls into, and the new admin-shell PHP — not a line-by-line pass of the full milestone roll-up of already-reviewed sub-PRs. CI is green, and PHP/security quality is high: REST routes carry capability-scoped permission callbacks, `$wpdb` queries are prepared (including the user-scope fragments, which are themselves prepared before concatenation), and the irreversible ESP-send hazard is well-contained — no `post_status` transitions in the new list/settings handlers. **No blockers** — inline notes are two suggestions + two nits.

**Additional non-inline items:**

- **Suggestion — `GET /lists` round-trips the ESP on every Settings mount** (`includes/class-newspack-newsletters-subscription.php`, `get_lists()`). The fetch always hits the remote ESP, and the companion plugin UI calls it again on every `LOCAL_LIST_SAVED` / `LOCAL_LIST_DELETED` event, amplifying the cost. Largely pre-existing, but worth a short-lived transient cache (busted on save/delete), or patching local list state from the event `detail` (which already carries the saved `list`) instead of a full refetch.
- **Nit — PHP type hints.** The new admin classes (`class-settings-rest.php`, `class-*-list-rest.php`, `class-asset-loader.php`, `class-admin-page.php`, the traits, etc.) document params/returns in PHPDoc but mostly omit signature and property types. The codebase targets PHP 8.3; property types and pure-helper return types are the safe wins (REST / filter callback signatures sometimes need to stay loose to match core's invocation). House-style call.

Cross-repo JS/wizard notes are on Automattic/newspack-plugin#4735.

includes/class-wizard-bridge.php
```diff
@@ -0,0 +1,76 @@
...
+		add_action( 'admin_enqueue_scripts', [ __CLASS__, 'maybe_enqueue' ] );
+	}
+
+	/**
+	 * Decide whether to enqueue on the current admin screen.
+	 *
+	 * @return bool
+	 */
+	public static function should_enqueue() {
+		if ( ! is_admin() ) {
```
chickenn00dle, line 39: **Nit.** `should_enqueue` gates on `is_admin()` + `class_exists` + `?page=`, with no `current_user_can` check. Low risk here — it only enqueues a JS bundle and a `debug` boolean (no sensitive data, no state mutation), and the wizard page itself is capability-gated — but `is_admin()` alone isn't a permission check, so a cap check is worth adding for defense-in-depth if this ever localizes more than a debug flag.

chickenn00dle (changes requested):
This is amazing @thomasguillot! Thanks for all of your hard work here. I left some nonblocking comments that came out of Claudes review. All of those are optional, however I did run into some blockers in my manual testing:

1. Creating a new Advertiser or Layout and then going back without making any changes results in this strange Auto Draft title. This may be intentional, but it seems like we should have a more descriptive generic title in these cases:

Image

2. On FireFox, existing layouts are not clickable. I have to click the action menu to edit:

https://github.com/user-attachments/assets/cba7aac9-4aea-4ae5-a619-2c49510bd957

3. The title and body are misaligned in the newsletter editor now:

Image

4. This one is more of a nit, but there is some strangeness when a newsletter is selected where the row becomes selected and for a moment it seems like you are able to do more in this view until you are taken to the editor:

https://github.com/user-attachments/assets/4091c6dc-b508-4b6a-9ccc-c7ff4776ca54

includes/admin/class-admin-shell-legacy-redirect.php
thomasguillot: Addressed in 72e51130 — forwarded `$_GET` values are now `rawurlencode`'d before `add_query_arg()` (which does not URL-encode values), so a value containing `&`/`=` can no longer inject extra params into the redirect target.

includes/admin/class-settings-rest.php
thomasguillot: Addressed in 16ede1cb — Mailchimp and ActiveCampaign now fire `newspack_newsletters_provider_credentials_changed` from `set_api_credentials()` on success, so the OAuth/validity cache busts uniformly no matter which path changes the credentials.

src/wizard-bridge/local-list-modal-host.js
thomasguillot: Addressed in 72e51130 — the effect cleanup now resets `window.newspackNewslettersBridgeReady = false`, keeping the readiness flag consistent with listener state across remounts.

includes/class-wizard-bridge.php
thomasguillot: Addressed in 72e51130 — `should_enqueue()` now requires `current_user_can( 'manage_options' )` (the wizard page's own capability) before enqueuing, for defense-in-depth.

chickenn00dle (approved):
Thanks for addressing all of my feedback! 

This is looking good for the most part. Just one final minor issue:

The strange Auto Draft title issue happens for new newletters as well:

https://github.com/user-attachments/assets/fe5e5911-b985-480c-9b7e-cd977b03187e

I'll go ahead and approve this one as well since this change shouldn't be too big.

